PR #38, opened 2021-04-08
Registercomponent

BobWez98:
Still to do: login after register.

if one of the 2 provided ways is accepted, i'll add a nice way to login te user after register.


royduin:
Thanks 💯 made it possible with the changes from the 0.1.3 release: https://github.com/rapidez/core/releases/tag/0.1.3

PR #57, opened 2021-08-04
Configurable Widgets

BobWez98:


config/rapidez.php
```diff
@@ -63,6 +63,10 @@
         'block'        => Rapidez\Core\Models\Block::class,
     ],
 
+    'widgets' => [
+
```
BobWez98, line 67: This will be for example:
`'JustBetter\MultiPurposeWidget\Block\Widget\Notification' => App\Widgets\NotificationWidget::class`

royduin:
https://github.com/rapidez/core/blob/master/src/Models/Widget.php#L27 and the Block directive should also work with this. Maybe add some basic Magento widgets like `Magento\Cms\Block\Widget\Block`?

src/Rapidez.php
```diff
@@ -6,8 +6,30 @@ class Rapidez
...
+
+    public function parseWidgetContent($content)
+    {
+        $vars = $this->getWidgetVars($content);
+        if (!isset($vars['type'])) {
+            return '';
+        }
+
+        $parseClass = config('rapidez.widgets.'.$vars['type']);
+        $content = (new $parseClass($vars))->render();
```
royduin, line 20: Maybe check if the class exists? Otherwise a message which is only visible on non-production environments about a non-implemented widget?

royduin:
We also have this trait: https://github.com/rapidez/core/blob/master/src/Models/Traits/HasContentAttributeWithVariables.php which should also be applied everywhere. It's not only widgets but also blocks and other variables. Maybe we should get rid of that and have a universal `@content()` directive or a `Rapidez::content()` facade we can use everywhere which parses widgets, blocks and variables.

royduin:
Just a note; the `HasContentAttributeWithVariables` trait is also used here: https://github.com/rapidez/snowdog-menu/blob/master/src/Models/MenuItem.php#L14:L16

PR #58, opened 2021-08-04
Set payment information

jbclaudio:


royduin:
Thanks! 💯 added it already with: https://github.com/rapidez/core/commit/d4d8aaf22efea0c3f03564f1957929326c131f37

PR #60, opened 2021-08-25
Added eventy filter to add scopes to attribute

BobWez98:
I need to extend the query with if a filter is a multi or single select.


royduin:
Maybe make this generic and move it to https://github.com/rapidez/core/blob/master/src/Models/Model.php so we do not need this in every model? As far as I know all models are extending that class.

PR #63, opened 2021-09-13
Replace REST API with Graphql for Cart

indykoning:
This PR aims to get rid of the REST API for updating the cart, updating product qtys and removing products, and instead opt to use graphql for this.

In this case the minicart is the brains of the operation since it is loaded on any page that requires the cart information to be updated. 
It listens to an event fired to refresh the cart and then calls the graphql endpoint again to get updated information.

### DIFF
--- a/resources/js/app.js
+++ b/resources/js/app.js
@@ -24,6 +24,7 @@ require('./axios')
 require('./filters')
 require('./mixins')
 require('./turbolinks')
+require('./callbacks')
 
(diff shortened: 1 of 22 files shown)


royduin:
Note to self: Check if we can get the `sales_rule_label` through GraphQL.

royduin:
Closed in favor of: https://github.com/rapidez/core/pull/372

PR #64, opened 2021-09-14
Resolve swatches & dropdown values anywhere

indykoning:
This PR aims to fix the dropdown values on products not showing the attribute values if the user has not navigated to a category page.

it also fixes non-swatch values missing their label in category pages


resources/views/listing/partials/filter/swatch.blade.php
```diff
@@ -1,38 +1,37 @@
-<swatch-filter v-else-if="filter.text_swatch || filter.visual_swatch" v-slot="slotProps">
```
indykoning, line 1: Alleen swatch filter is weg gehaald en de `v-else-if="filter.text_swatch || filter.visual_swatch"` is verplaatst naar multi-list

royduin:
Thanks but there can be a lot of option values so I don't think it's a good idea to get and save them all.

PR #67, opened 2021-09-21
Add additional info for payment providers to use in the checkout

indykoning:
This will allow payment providers to check if their payment provider has been used.
And it will then allow them to use the order id which is returned after converting the quote


royduin:
Thanks 💯

PR #68, opened 2021-09-21
Resolve swatches anywhere

indykoning:
This PR aims to fix the dropdown values on products not showing the attribute values if the user has not navigated to a category page


royduin:
Thanks 💯

PR #71, opened 2021-09-22
Move responsibility for clearing cart to module canceling next step

indykoning:
Currently Rapidez removes the cart & order from the session the moment "Place order" is pressed.
This presents a problem in case the order is canceled or failed on the end of an external payment provider.

This PR moves the responsibility of clearing the cart from the session to the payment providers package.
This way a canceled payment does not result in a lost cart.

Currently existing payment provider modules will have to be updated, such as https://github.com/rapidez/mollie


royduin:
Looking forward to the PR which implements it in rapidez/mollie 😇

royduin:
Thanks 💯

PR #72, opened 2021-09-23
Add eventy filter to expose variables in config from modules

BobWez98:


royduin:
You can just add it:
```
config(['frontend.something' => 'value']);
```
We're already doing that on different places: https://github.com/search?l=PHP&q=org%3Arapidez+config%28%5B%27frontend%27&type=Code

PR #99, opened 2022-02-01
Added events for refreshCart and clearCart & Added runQuery to slot scope

jbclaudio:


resources/js/components/Cart/Cart.vue
```diff
@@ -37,6 +37,15 @@
                     Notify(error.response.data.message, 'error')
                 }
             }
+        },
+        created() {
+            this.$root.$once('refresh-cart', () => {
+                this.refreshCart()
+            });
+
+            this.$root.$once('clear-cart', () => {
```
indykoning, line 46: `clearCart(keys = [])` also supports an array of keys, maybe it's useful to add it as a parameter to the event listener as well

resources/js/components/Cart/Cart.vue
```diff
@@ -37,6 +37,15 @@
                     Notify(error.response.data.message, 'error')
                 }
             }
+        },
+        created() {
+            this.$root.$once('refresh-cart', () => {
```
indykoning, line 42: You could replace the callback with `this.refreshCart`
```
this.$root.$once('refresh-cart', this.refreshCart)
```

PR #100, opened 2022-02-04
Get Shipping and Totals information from shipping address country id

jbclaudio:


indykoning:
The browser test will need to be updated to be updated to select a country that has a shipping method attached https://github.com/rapidez/core/blob/master/tests/Browser/CheckoutTest.php#L30

PR #109, opened 2022-02-14
Prevent multiple code execution of script with the array cache driver + Only cast the Super Attributes that are really needed by the product

rbnmulder:


royduin:
Thanks! But I think this is a better: https://github.com/rapidez/core/commit/7db518afda9d7d5cc18509609bcaabb01a0dd8df

PR #119, opened 2022-03-28
Add ViewOnly widget class to pass options

quintenbuis:
This PR adds a ViewOnly class that takes the view name as the render method argument and passes options to that view.

Before the widget config option must follow the following structure:
```php
'widgets' => [
    'Magento\Cms\Block\Widget\Block'                   => Rapidez\Core\Widgets\Block::class,
],
```

Now instead mapping the Magento block to a full class string you can pass a view name and it will grab the `view_only_widget` class in the config and render the view:
```php
'widgets' => [
    'Magento\Cms\Block\Widget\Block'                   => 'widget-with-options-view',
],
```

The ViewOnly class can also be customized in the config:
```php
'view_only_widget' => \Rapidez\Core\Widgets\ViewOnly::class,
```


src/ViewDirectives/WidgetDirective.php
```diff
@@ -30,8 +30,16 @@ function () use ($location, $type, $handle, $entities, $replace) {
                     $widgetClass = config('rapidez.widgets.'.$widget->instance_type);
 
                     if (!class_exists($widgetClass)) {
```
indykoning, line 32: Might it be an idea to extract all this logging to a different function?
It'll allow us to return early and hopefully reduce the amount of nesting

royduin:
Thanks! 💯  Docs: https://docs.rapidez.io/0.x/extending.html#widgets

PR #134, opened 2022-05-25
Results per page filter

BobWez98:


resources/views/listing/partials/stats.blade.php
```diff
@@ -1,6 +1,12 @@
-<div class="flex-grow text-sm" slot="renderResultStats" slot-scope="{ numberOfResults, numberOfPages, currentPage, time }">
+<div class="flex-grow text-sm flex justify-between" slot="renderResultStats" slot-scope="{ numberOfResults, numberOfPages, currentPage, time }">
     @{{ numberOfResults }} @lang('products')
     <template v-if="numberOfPages > 1">
         (@lang('page'): @{{ currentPage + 1 }}/@{{ numberOfPages }})
     </template>
+
+    <div class="justify-self-end mr-1">
+        <select class="css-oswg5g !outline-none !rounded shadow focus:ring focus:ring-green-500" v-model="$root.listingResultsPerPage">
```
BobWez98, line 8: The css class is the same as the sort options to allow same default frontend.

PR #138, opened 2022-06-20
Added params to notify to use Magento feedback

BobWez98:


resources/js/components/Notifications/Notification.vue
```diff
@@ -31,12 +31,18 @@
...
             message: null,
             type: null,
-            show: null,
+            show: null
         }),
         methods: {
             close() {
                 this.show = false
+            },
+            format(str, arr) {
```
BobWez98, line 40: I don't know if this function should be elsewhere because i don't know if any other api feedback gives these types of messages.

PR #154, opened 2022-08-16
Remove error parameters from successful response

indykoning:
Currently the loader on the category pages for add to cart does not disappear since it throws an error that error is not defined


royduin:
Are there any other locations where this issue occurs? It was recently added: https://github.com/rapidez/core/pull/151

indykoning:
This was the only instance where i could find it incorrectly placed there in that PR.
In every other place the error variable is defined

royduin:
Awesome, thanks 💯

PR #163, opened 2022-10-17
Fallback to using option code if its not an id

indykoning:


royduin:
It should always be an ID right? If it's not there is something really wrong with your Magento installation and/or product import. This way it's ignored, is that the right thing to do? Or is it better to throw an exception?

indykoning:
If the options always have proper labels attached this shouldn't occur.
However if the labels are not properly given to the option, or the multiselect values were managed by code e.g. by an extension adding a custom attribute it is still possible to get a string in there instead of an id.

PR #164, opened 2022-10-17
Improve code readability

indykoning:


royduin:
StyleCI will revert the styling...

indykoning:
That's a shame, it would've made it a lot more readable...
Then the applicable changes in here are the docblocks added to the facade, which is very valuable when trying to check what's available.

And removing the use of the same config functions in different ways
https://github.com/rapidez/core/pull/164/files#diff-fe38fec1955243f9e20c1cdc08b64167096d11ee224c3c86fc8e4f4671750a51L40

PR #166, opened 2022-10-31
Implement better fallback routing

indykoning:
Fallback routes can now be added via 
```php
FallbackRoutes::add(UrlRewriteController::class, 5);
```
anywhere within your app before routing takes place. e.g. your web.php will be fine, but your serviceProvider is a better place.
You can optionally set a sorting value. Lower is earlier.
The old eventy filter logic is still available but sorted way at the end to ensure compatibility but lowest priority.

Actions can be passed by any method permitted by the normal router
e.g.
```php
FallbackRoutes::add(UrlRewriteController::class, 5);
FallbackRoutes::add('UrlRewriteController@index', 5);
FallbackRoutes::add([UrlRewriteController::class, 'index'], 5);

FallbackRoutes::add(function (Request $request) {return redirect('/');}, 5);
```


src/RapidezServiceProvider.php
```diff
@@ -80,6 +84,10 @@ protected function bootRoutes(): self
             $this->loadRoutesFrom(__DIR__.'/../routes/api.php');
         }
 
+        FallbackRoutes::add(UrlRewriteController::class, 5);
+        FallbackRoutes::add(CmsPageController::class, 10);
+        FallbackRoutes::add(LegacyFallbackController::class, 10000000);
```
royduin, line 89: A bit too obvious? 99, 999 of 9999 should also do the job.

royduin:
Maybe just use the existing `Rapidez` facade instead of a new one?

src/RapidezServiceProvider.php
indykoning: We want this ones priority to be lower than the default which is currently `99999`
Of course both of these numbers are completely arbitrary and could be whatever number as long as they are high end the legacy fallback is lower than the default.

I could put the default as `9999` and the fallback as `99999` so theres enough space before and after in case a fallback needs to have lower priority than is the default but higher than the legacy fallback

indykoning:
> Maybe just use the existing `Rapidez` facade instead of a new one?

That would be possible, but that means that the implementation of the Rapidez class would need to become a singleton. instead of a bind which reinitialises each time it is requested.
And of course the FallbackRoutesRepository is clear to only take care of fallback routes right now

PR #174, opened 2022-12-12
edit frontend PDP

JimmyHoenderdaal:


resources/views/product/partials/images.blade.php
```diff
@@ -1,73 +1,33 @@
...
-                    />
+            <div v-if="images.length > 1" class="flex" :class="zoomed ? 'fixed bottom-0 left-3 {{ config('rapidez.z-indexes.lightbox') }}' : ' flex-wrap mt-3'">
+                <a href="#" v-for="(image, imageId) in images" class="mr-3 mb-3 flex h-[100px] w-[100px] items-center justify-center rounded border bg-white p-2" :class="active == imageId ? 'border-primary' : ''" @click.prevent="change(imageId)">
+                    <img :src="'/storage/resizes/80x80/catalog/product' + image + '.webp'" alt="{{ $product->name }}" class="m-auto max-h-[80px] w-full object-contain" loading="lazy" width="80" height="80" />
                 </a>
             </div>
-
-            <div v-if="zoomed" class="fixed top-3 right-3 pointer-events-none {{ config('rapidez.z-indexes.lightbox')}}">
-                <x-heroicon-o-x class="h-6 w-6"/>
+            <div v-if="zoomed" class="{{ config('rapidez.z-indexes.lightbox') }} pointer-events-none fixed top-3 right-3">
```
indykoning, line 29: My preference with merging php variables into classes would be.
Keep in mind this does not get recognised by tailwind and might not get compiled if the class is not used somewhere else
```suggestion
            <div v-if="zoomed" @class([
            	config('rapidez.z-indexes.lightbox'), 
            	'pointer-events-none fixed top-3 right-3'
            ])>
```

indykoning (changes requested):

PR #179, opened 2022-12-16
Frontend fixes POP

Roene-JustBetter:


resources/views/category/overview.blade.php
```diff
@@ -6,7 +6,7 @@
 @section('content')
     @include('rapidez::category.partials.breadcrumbs')
 
-    <h1 class="font-bold text-3xl mb-5">{{ $category->name }}</h1>
+    <h1 class="container mx-auto font-bold text-3xl mb-5">{{ $category->name }}</h1>
```
royduin, line 9: Are we going to do this in every component? Or just once in the layout file and if we need something full width we break out of it?

PR #191, opened 2023-01-25
Fix slider pagination + Wishlist buttons

JimmyHoenderdaal:


resources/views/layouts/partials/header/account.blade.php
```diff
@@ -9,6 +9,7 @@
                 @if(App::providerIsLoaded('Rapidez\Account\AccountServiceProvider'))
                     <a class="block hover:bg-secondary px-3 py-2" href="/account">@lang('Account')</a>
                     <a class="block hover:bg-secondary px-3 py-2" href="/account/orders">@lang('Orders')</a>
+                    <a class="block hover:bg-secondary px-3 py-2" href="/account/wishlist">@lang('Wishlist')</a>
```
royduin, line 12: This page isn't included within the account page right? So if we remove the package the link is still present.

resources/views/listing/partials/item.blade.php
```diff
@@ -1,6 +1,9 @@
 <template {!! isset($slider) ? '' : 'slot="renderItem" slot-scope="{ item, count }"' !!}>
     <div class="flex-none w-1/2 lg:w-1/3 my-0.5 px-0.5 sm:px-2 sm:my-2 snap-start">
         <div class="w-full bg-white border rounded hover:opacity-80 group relative h-full flex flex-col" :key="item.id">
+            <div class="absolute top-0 right-0 z-10 group p-2">
+                @include('rapidez::wishlist.button')
```
royduin, line 5: The same problem, this should be conditional I think?

resources/views/product/partials/images.blade.php
```diff
@@ -1,4 +1,7 @@
 <div class="relative">
+    <div class="absolute top-0 right-0 z-10 group p-2">
+        @include('rapidez::wishlist.button')
```
royduin, line 3: The same problem, this should be conditional I think?

royduin (changes requested):
So where is the `rapidez/wishlist` package?

JimmyHoenderdaal:
> So where is the `rapidez/wishlist` package?

this is installed in rapidez/rapidez on the front-end branche

PR #194, opened 2023-01-26
Improve frontend

JimmyHoenderdaal:


JimmyHoenderdaal:
This includes the merge of master which added turbo

resources/views/product/overview.blade.php
```diff
@@ -45,6 +50,22 @@
                 </div>
             </div>
         </div>
+    </div>
+    <div class="bg-highlight my-5">
```
royduin: No if statement around this?

royduin: Nevermind

PR #196, opened 2023-01-26
Use minified Vue so development mode is disabled for production

danielsmink:


vite.config.js
```diff
@@ -1,27 +1,31 @@
-import path from 'path'
+import path from 'path';
```
royduin, line 1:
```suggestion
import path from 'path'
```

vite.config.js
```diff
@@ -4,24 +4,28 @@ import laravel from 'laravel-vite-plugin'
...
-        visualizer(),
-    ],
-    resolve: {
-        preserveSymlinks: true,
-        alias: {
-            '@': path.resolve(__dirname, './resources/js'),
-            'Vendor': path.resolve(__dirname, './vendor'),
-            'vue': path.resolve(__dirname, './node_modules/vue/dist/vue.js')
+export default ({ mode }) => {
+    let vueProduction = process.env?.NODE_ENV === 'production' ? '.min' : '';
```
royduin, line 8: Maybe use the debug boolean from the env for this? As on test environments everything is also compiled as production but the env is or should be staging.

royduin (approved):

indykoning:
See: https://github.com/rapidez/core/pull/197 this should make vue compilation work as it should including all variables it might need.

royduin:
Closed in favor of https://github.com/rapidez/core/pull/197

PR #201, opened 2023-01-27
Reset idle and loading check if either turn false within a duration

indykoning:
This should catch network requests fired in response to a network request finishing and wait for them too.
Default functionality if no network requests are done or the browser is idle is a pause of 500ms.

If either turns false during this 500ms it shall wait until both stay true during these 500ms


royduin:
Tests are failing?

PR #208, opened 2023-02-10
Added option to translate labels

Roene-JustBetter:


royduin:
This is already handled within the component: https://github.com/rapidez/core/blob/master/resources/views/components/input.blade.php#L6

PR #210, opened 2023-02-15
Mobile site update

Jade-GG:
- Changed mobile logo to be the smaller `R` logo
- Removed the content block widget things on mobile because they got way too messy
- Added the mobile version of the category menu (see also [this PR in the menu repo](https://github.com/rapidez/menu/pull/3))


royduin:
We don't want the menu 2x in the dom.

PR #211, opened 2023-02-16
Add fallback for having a trailing slash as suffix

rbnmulder:
Currently the Product Overview / Detail page doesn't work when Magento is configured to use a "/" as suffix for category / product pages. This is saved to the `url_rewrite` entries (request_path). When fetching a `url_rewrite` we use the `path` function which trims the slash from the beginning and end. This results in a 404.

This change will fix the 404 by using `/` as fallback. But it also creates 2 entry points in the proces for both category and product pages (if configured with using `/` as suffix. This should not be a problem because in Rapidez we have a canonical tag that stays the same.


royduin:
Just remove the `catalog/seo/category_url_suffix` setting and reindex, having a `/` as suffix isn't recommended anyway.

PR #215, opened 2023-02-21
Execute fallback on not found exception

indykoning:
As an improvement this handles routes that have been matched but threw a 404 for one reason or another and still allows us to handle that route with a fallback.
So it's a fallback for ANY route instead of just a fallback for routes that aren't defined.

Marking this as draft as a discussion point, and since it's impact hasn't been tested with error-handling libraries like Sentry


royduin:
Closed due inactivity.

PR #220, opened 2023-03-13
Define carrier code and methode code and use them when setting the sh…

rbnmulder:
…ipment_method

This prevents problems with resolving  `shipping_carrier_code` and `shipping_method_code` when a carrier / methode code contains extra `_` characters. Example shipment method with carrier_code`tig_postnl` and method_code `regular`.

### DIFF
--- a/resources/js/components/Checkout/Checkout.vue
+++ b/resources/js/components/Checkout/Checkout.vue
@@ -49,8 +49,6 @@
                     this.checkout.shipping_methods = response.data
 
                     if (response.data.length === 1) {
-                        this.checkout.shipping_carrier_code = response.data[0].carrier_code
-                        this.checkout.shipping_method_code = response.data[0].method_code
                         this.checkout.shipping_method = response.data[0].carrier_code+'_'+response.data[0].method_code
                     }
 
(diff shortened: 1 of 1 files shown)


indykoning (approved):

resources/js/components/Checkout/Checkout.vue
```diff
@@ -311,6 +311,9 @@
                 }
 
                 return this.removeUnusedAddressInfo(this.$root.checkout.billing_address)
+            },
+            currentShipmentMethod: function () {
+                return this.checkout.shipping_methods.find(obj => obj.carrier_code+'_'+ obj.method_code === this.checkout.shipping_method)
```
royduin, line 316: `obj`? And maybe a bit to long to do on one line?
```suggestion
                return this.checkout.shipping_methods.find((method) => {
                    return method.carrier_code + '_' + method.method_code === this.checkout.shipping_method
                })
```

resources/js/components/Checkout/Checkout.vue
```diff
@@ -311,6 +311,11 @@
                 }
 
                 return this.removeUnusedAddressInfo(this.$root.checkout.billing_address)
+            },
+            currentShipmentMethod: function () {
```
royduin, line 315:
```suggestion
            currentShippingMethod: function () {
```
Magento is using "shipping" everywhere (not "shipment"), using that is more inline I think?

royduin:
Please also create an PR to the master branch.

PR #221, opened 2023-03-14
Improve frontend

JimmyHoenderdaal:


royduin:
Did you sync master in your branch? The improve frontend is based on 0.x, maybe better to keep it that way and when everything is finished we'll merge improve frontend into master?

PR #223, opened 2023-03-20
Add Eventy hooks for before/after rapidez indexer

Jade-GG:
Simple solution to being able to quickly hook in multiple actions around the indexer

Can be used like this (e.g. in `App\Providers\AppServiceProvider:boot`):

```php
Eventy::addAction('index.before', function($context) {
    $context->call('rapidez:index:sortings');
});
Eventy::addAction('index.after', function($context) {
    $context->call('rapidez:index:additionals');
});
```

### DIFF
--- a/src/Commands/IndexProductsCommand.php
+++ b/src/Commands/IndexProductsCommand.php
@@ -21,6 +21,8 @@ class IndexProductsCommand extends InteractsWithElasticsearchCommand
 
     public function handle()
     {
+        Eventy::action('rapidez.index.before', $this);
+        
         $this->call('cache:clear');
         $productModel = config('rapidez.models.product');
         $storeModel = config('rapidez.models.store');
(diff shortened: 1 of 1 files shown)


src/Commands/IndexProductsCommand.php
```diff
@@ -21,6 +21,8 @@ class IndexProductsCommand extends InteractsWithElasticsearchCommand
 
     public function handle()
     {
+        Eventy::action('rapidez.index.before', $this);
```
royduin, line 24:
```suggestion
        Eventy::action('index.before', $this);
```

src/Commands/IndexProductsCommand.php
```diff
@@ -86,6 +88,8 @@ public function handle()
             $bar->finish();
             $this->line('');
         }
+        
+        Eventy::action('rapidez.index.after', $this);
```
royduin, line 92:
```suggestion
        Eventy::action('index.after', $this);
```

royduin (changes requested):
Please add it to the docs and I also like to have this in 0.x

Jade-GG:
> Please add it to the docs and I also like to have this in 0.x

I have made a PR: https://github.com/rapidez/docs/pull/31

royduin:
Merged, could you also make a PR to the 0.x branch?

PR #228, opened 2023-03-30
Add caching to fallback routes

indykoning:
By caching fallback routes per url we can greatly reduce the response time in case of heavy controllers, like requests to external services (e.g. strapi)

Here's an example how the timeline is currently in the fallback controllers if the page is available in a later controller
image

Once a url has had it's controller cached the timeline looks like this
image

This could be further optimised by caching the response to this url to prevent the external call in the first place.
But this is the responsibility of the controller making that external call.

Customisation to the cache duration can be done via the `rapidez.fallback_routes.cache_duration` config key.


src/Http/Controllers/FallbackController.php
```diff
@@ -2,32 +2,58 @@
...
+                continue;
+            }
 
-                return $response;
-            } catch (RouteNotFoundException|NotFoundHttpException|BackedEnumCaseNotFoundException|ModelNotFoundException|RecordsNotFoundException $e) {
+            try {
+                Cache::put($cacheKey, $route, value(config('rapidez.fallback_routes.cache_duration', 3600), $request, $response, $route));
+            } catch (Exception $e) {
+                // We can't cache it, no worries.
+                // Ususally a sign it's a direct callback or caching hasn't been configured properly.
```
royduin, line 35: Why do we want to catch this?

indykoning: Because anything to do with saving the controller in cache shouldn't hinder returning a response.

Cases that i've had error pages on this were because of anonymous crontrollers `Rapidez::addFallbackRoute(fn() => 'test')` because an anonymous function can not be serialized for storage in cache.

And an improperly configured redis application, e.g. driver missing, redis full

Caching the controller for a request is not necessary but can speed up the application when one in the list has low speed

royduin: But Laravels route caching does have support for anonymous callback functions? Catching for misconfiguration isn't needed as caching is used on multiple places where we don't try/catch.

indykoning: That's true Laravel's route caching does indeed work with anonymous callback functions because it caches the routes per-route in a "routes.php" file while we're caching it in Redis per-url.

Redis can only save strings, so we need to Serialize the functions. In the case of referenced functions this is easy as it is just pointing to the function that should be executed.

When it is a direct function however you will get the following error:
image
As transforming a function into a string is not possible by default

There may be ways to hack it by working with exec with the string stored in cache, but that can open up many other issues

royduin:
I'll merge when the PR for is master is ready 🚀

PR #234, opened 2023-04-07
RAP-469 - Prefix super attributes

indykoning:
This allows for relatively good backwards compatibility with the old way of working with attributes.
All super variables are available like `super_` but if you ask for the variable without `super_` it will check if it is not null.
If it is, it will attempt to get the `super_` value and cast it according to it's non `super_` casts.

The only limitation is that this does not happen during a `toArray` or `toJson`

### DIFF
--- a/src/Models/Product.php
+++ b/src/Models/Product.php
@@ -199,6 +199,11 @@ class Product extends Model
             return $val;
         }
 
-        return $this->transformModelValue($key, parent::getAttribute('super_'.$key));
+        $val = parent::getAttribute('super_'.$key);
+        if($this->hasCast('super_'.$key)) {
+            return $val;
+        }
+
+        return $this->transformModelValue($key, $val);
     }
 }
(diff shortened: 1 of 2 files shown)


royduin:
Did you test this? As the data is added here: https://github.com/rapidez/core/blob/master/src/Http/Controllers/ProductController.php#L23 which will be transformed to an array/object. Backwards compatible is nice but maybe better to force it? Sometimes the attribute and other times the super attribute is wanted, better te be explicit I think?

royduin:
But if both are filled you're always getting the super attribute value, I think it's better to be explicit. If you need a super attribute, just prefix it so you know what you're getting?

royduin:
Awesome! Ready to merge? Another PR for master is coming?

indykoning:
This is ready to merge and ready to merge into master as well, making this PR to master would bring a lot of changes with it. It may be better to merge 0.x into master

royduin:
`0.x` is impossible to merge in `master` at this point and won't happen in the future. Everything going in `0.x` should also be added in `master`.

PR #236, opened 2023-04-11
Automatically load first childs in components folder

indykoning:


royduin:
Aren't we getting any issues where a package needs something from the core but the core is not yet imported this way?

royduin:
Closed due inactivity.

PR #241, opened 2023-04-20
Install Duster to format codestyle

indykoning:


indykoning:
I've done some configuring on the linters and fixers, these are the changes it currently makes https://github.com/indykoning/core/commit/946c7c8ac2a7fd90da2707b907cd29643e7b80ea

Notably the spaces between string concatenation 
```diff
- 'test'.'string'
+ 'test' . 'string'
```

Do we want this or should i update the rules to force no space here?

PR #242, opened 2023-04-20
Install Duster to format codestyle

indykoning:


royduin:
I prefer not but it's a PSR-12 thing I think? If so let's stick to that.

PR #244, opened 2023-04-21
Use prettier fixes, update codestyle

indykoning:


indykoning:
I've configured prettier, it now formats like this https://github.com/indykoning/core/commit/cc699c082723b0b5f7581871a109bf3a4b534d73
It also formats the blade files but should only do indenting, html and tailwind class sorting

PR #245, opened 2023-04-21
Use prettier fixes, update codestyle

indykoning:


indykoning:
I've configured prettier, it now formats like this https://github.com/indykoning/core/commit/cc699c082723b0b5f7581871a109bf3a4b534d73
It also formats the blade files but should only do indenting, html and tailwind class sorting

royduin:
Maybe disable it until it's good?

indykoning:
I've had to disable the entire prettier blade parser unfortunately.
This is the result, css, vue and js still get checked and formatted!
https://github.com/indykoning/core/commit/d8ce291cfc0bf724fb486415dc778cfd8baa21d3

.editorconfig
```diff
@@ -1,15 +1,34 @@
...
+# Unofficial config, support depends on the application reading the file.
+# https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#supported-by-a-limited-number-of-editors
+# https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#ideas-for-domain-specific-properties
+max_line_length = off
+quote_type = single
+spaces_around_operators = true
+spaces_around_brackets = none
+
+# https://en.wikipedia.org/wiki/Indent_style
+indent_brace_style = Allman
```
royduin, line 21: This isn't conform PSR? This should be K&R I think?

.editorconfig
```diff
@@ -1,15 +1,34 @@
...
 
 [*.{yml,yaml}]
+indent_style = space
 indent_size = 2
+
+[*.{blade.php,html}]
+max_line_length = 400
+
+[*.{js,vue}]
+max_line_length = 140
```
royduin, line 34: Do we really want to enforce this?

.github/workflows/pretter.yml
```diff
@@ -0,0 +1,37 @@
+name: Prettier Fix
+
+on:
+  push:
+    branches:
+      - master
+      - develop
+      - 0.x
+  workflow_run:
+    workflows: ["Duster Fix"]
```
royduin, line 10: Duster?

.github/workflows/pretter.yml
```diff
@@ -0,0 +1,37 @@
+name: Prettier Fix
+
+on:
+  push:
+    branches:
+      - master
+      - develop
```
royduin, line 7: We do not have a `develop` branch, maybe just run it on all branches?

royduin (changes requested):

.editorconfig
indykoning: You're absolutely right

.editorconfig
indykoning: Maybe not necessarily enforce it, but it instructs tools like prettier when it should move a long line of code to the next line, and this seems to match up pretty well with our code.

.github/workflows/pretter.yml
indykoning: If duster fixes code it may prevent other actions from committing their fix, due to limitations from GitHub.
So if it has run we want to run this action after duster has committed their change.
https://github.com/tighten/duster-action#:~:text=The%20resulting%20commit%20will%20not%20trigger%20another%20GitHub%20Actions%20Workflow%20run

PR #246, opened 2023-04-21
add autoscroll to master

JimmyHoenderdaal:


resources/js/components/Elements/Slider.vue
```diff
@@ -18,20 +19,48 @@
...
-                mounted: false
+                mounted: false,
+                hover: false,
+                pause: ()=>{},
+                resume: ()=>{}
             }
         },
         mounted() {
-            useEventListener(this.slider, 'scroll', this.scroll)
+            this.slider.addEventListener('scroll', this.scroll)
```
royduin, line 48: This one could stay so we don't need the `beforeDestroy()` as VueUse will take care of that automatically.

cc @indykoning

resources/js/components/Elements/Slider.vue
```diff
@@ -18,20 +19,48 @@
...
+
+            if (this.autoplay) {
+                let { pause, resume } = useIntervalFn(this.autoScroll, this.interval)
+                this.pause = pause
+                this.resume = resume
+            }
+        },
+        beforeDestroy() {
+            this.slider.removeEventListener('scroll', this.scroll)
         },
```
royduin, line 64: So we don't need this now we're using `useEventListener()` right?

PR #247, opened 2023-04-26
Add specific count parameter to add-to-cart add function

Jade-GG:
Useful for some specific cases where this is the most straightforward solution

### DIFF
--- a/resources/js/components/Product/AddToCart.vue
+++ b/resources/js/components/Product/AddToCart.vue
@@ -66,7 +66,7 @@
             },
 
             async add() {
-                return this.addCount(this.qty);
+                return this.addQty(this.qty);
             },
 
             async addQty(count) {
--- a/resources/js/components/Product/AddToCart.vue
+++ b/resources/js/components/Product/AddToCart.vue
@@ -52,6 +52,7 @@
                 options: this.options,
                 error: this.error,
                 add: this.add,
+                addQty: this.addQty,
                 qty: this.qty,
 
                 adding: this.adding,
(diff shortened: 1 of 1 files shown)


resources/js/components/Product/AddToCart.vue
```diff
@@ -64,14 +64,16 @@
                 this.qty = event.target.value
             },
 
-            async add() {
+            async add(count) {
```
royduin, line 67: Aren't we getting any errors when we don't specify a param this way?

Jade-GG: The usage of the param is checked later, defaulting to `this.qty` when it's not specified

royduin: So doing `add()` without specifying a param works without errors?

royduin: Alright; why are the tests failing?

Jade-GG: I've changed things around slightly as I'm thinking the `$event` is actually getting passed to the function as a param causing it to fail.

It should be extra safe and backwards compatible now :)

resources/js/components/Product/AddToCart.vue
```diff
@@ -65,13 +66,19 @@
             },
 
             async add() {
+                return this.addQty(this.qty);
+            },
+
+            async addQty(count) {
```
royduin, line 72:
```suggestion
            async add(count = 1) {
```

Jade-GG: This wouldn't solve the backwards compatibility issue because in some cases, `@click="add"` is used, causing the `$event` to be passed as the `count` parameter.

PR #251, opened 2023-05-08
Default to using this.$el for slider

Jade-GG:
We can't always use just `this.$el`, even in the example given in the core. However, we can default to it and let people use the ref in case it's necessary :)

### DIFF
--- a/resources/js/components/Elements/Slider.vue
+++ b/resources/js/components/Elements/Slider.vue
@@ -12,10 +12,7 @@
             })
         },
         props: {
-            reference: {
-                type: String,
-                default: 'slider'
-            },
+            reference: String,
             vertical: {
                 type: Boolean,
                 default: false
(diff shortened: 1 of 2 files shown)


resources/js/components/Elements/Slider.vue
```diff
@@ -91,6 +88,9 @@
         },
         computed: {
             slider() {
+                if(!this.reference) {
+                    return this.$el;
```
royduin, line 92:
```suggestion
                    return this.$el.firstChild;
```

This way we don't need the reference by default at all right?

Jade-GG: This is very true, although I feel like to some degree it's just moving the problem laterally 😅

It at least made sense for it to use the root node of the component when not given a reference, but to me this is a little too specific. I guess "overfitting" is the way I'd describe it? Not sure if I'm into it.

PR #252, opened 2023-05-09
Get tax display configuration from Magento

jbclaudio:
For this I made an example of how this could be implemented. It is not yet completely 100%, but let's first see if this is the right way. If this is the right way, this can continue to work and the code can be fine-tuned and finished.

### DIFF
--- a/resources/js/components/Price.vue
+++ b/resources/js/components/Price.vue
@@ -1,0 +1,92 @@
+<script>
+export default {
+    props: {
+        product: {
+            type: Object,
+            default: () => window.config.product || {},
+        },
+        type: {
+            type: String,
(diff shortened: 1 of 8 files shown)


resources/js/components/Price.vue
```diff
@@ -0,0 +1,92 @@
...
+                return this.product.special_price
+            }
+
+            return this.calculateTax(this.product.special_price, true)
+        }
+    },
+
+    computed: {
+        price() {
+            if (this.mounted) {
```
indykoning, line 73: Do we need these checks wether the component is mounted? If so i would suggest flipping the logic around to return early if it is not mounted

PR #253, opened 2023-05-11
Added popup functionality | Added cookie notice

Roene-JustBetter:


royduin:
Please fix the tests 😇

royduin:
Please fix the tests 😇

resources/js/components/Popup.vue
indykoning, line 1: How much of this component do we use for the cookie notice that the default html5 dialog element does not have? https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog
It'd save on quite a bit of JS

resources/views/components/cookie-notice.blade.php
```diff
@@ -0,0 +1,26 @@
...
+        class="fixed inset-x-0 bottom-0 z-30 pb-2"
+        slot-scope="{ show, close, accept }"
+        v-if="show"
+    >
+        <div class="container">
+            <div class="rounded bg-white p-6 shadow">
+                <div class="flex flex-wrap items-center justify-between">
+                    <div class="flex-1 items-center">
+                        <div class="cookie-consent__message text-sm text-black">
+                            @lang('This website uses cookies')
```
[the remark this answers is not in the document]
indykoning: Maybe it's a good idea to use slots for this instead in that case.
otherwise everyone using this component will have to overwrite it

resources/js/components/CookieNotice.vue
```diff
@@ -0,0 +1,45 @@
...
+            open() {
+                this.$el.show()
+            },
+            accept() {
+                document.cookie = 'accept-cookies=true'
+                this.close()
+                location.reload()
+            },
+            close() {
+                this.$el.show()
```
indykoning, line 33:
```suggestion
                this.$el.close()
```

resources/views/components/cookie-notice.blade.php
```diff
@@ -0,0 +1,14 @@
+<cookie-notice name="cookie-notice" show-until-close v-cloak>
+    <dialog slot-scope="{ close, accept }" class="container rounded bg-white p-6 shadow fixed inset-x-0 bottom-4 z-30">
+        <div class="flex flex-wrap items-center justify-between">
+            <div class="flex-1 items-center">
+                <div class="cookie-consent__message text-sm text-black">
+                    {{ $slot }}
+                </div>
+            </div>
+            <x-rapidez::button @click="accept" variant="outline" class="mt-3 w-full shrink-0 md:ml-6 md:w-auto">
+                {{ $button ?? __('Accept coockies') }}
```
indykoning, line 10:
```suggestion
                {{ $button ?? __('Accept cookies') }}
```

jordythevulder, line 10: cockies xd

royduin:
Previously this was a generic popup component? Now it's just a cookie notice component? We don't want/need the popup component anymore?

royduin:
True but; in projects we currently do have a popup component with functionality such as a delay, show once, etc. would be nice to have a generic component which can be used for all scenarios.

Roene-JustBetter:
> True but; in projects we currently do have a popup component with functionality such as a delay, show once, etc. would be nice to have a generic component which can be used for all scenarios.

That is true, but then the question is. Is a cookie notice the same as an popup component? My opinion is that a popup has more functionalities so the popup component should be a different component then the cookie notice.

royduin:
They've overlapping functionality so a multi functional popup component which can be used for multiple things would be nice

indykoning (approved):

tests/Browser/CartTest.php
```diff
@@ -10,8 +10,9 @@ class CartTest extends DuskTestCase
     public function testAddSimpleProduct()
     {
         $this->browse(function (Browser $browser) {
-            $browser->visit($this->testProduct->url)
-                    ->waitUntilIdle()
+            $browser->visit($this->testProduct->url);
+            $browser->script("localStorage['cookie-notice'] = true;");
```
royduin, line 14: Can't we do this by default somewhere else? So we don't have to add this in every test? And add a new test without it so we can test the popup?

resources/views/components/cookie-notice.blade.php
```diff
@@ -0,0 +1,24 @@
+@if(App::environment() != 'testing' || request()->has('testingcookienoticetrue'))
```
royduin, line 1:
```suggestion
@if(!App::runningUnitTests() || request()->has('show-cookie-notice'))
```

tests/Browser/DialogTest.php
```diff
@@ -0,0 +1,24 @@
...
+
+class DialogTest extends DuskTestCase
+{
+    /**
+     * @test
+     */
+    public function test()
+    {
+        $this->browse(function (Browser $browser) {
+            $browser->visit('/?testingcookienoticetrue')
```
royduin, line 16:
```suggestion
            $browser->visit('/?show-cookie-notice')
```

PR #254, opened 2023-05-12
Generalized functionality for various indexer commands

Jade-GG:
Still only a draft as I'm not sure yet what the cleanest solution will end up being. Right now this functionality can be used to write complete indexer commands like this (these are working tested examples):

```php
$this->indexAllStores('blogs',
    fn($store) => Entry::query()->where('collection', 'blog')->where('site', $store)->get(),
    fn($item) => [
        'title' => $item->title,
        'url' => $item->url,
        'image' => $item->image ? $item->image->url() : '',
        'slug' => $item->slug,
        'categories' => $item->blog_categories->map(fn ($category) => [
            'title' => $category->title,
            'slug' => $category->slug
        ]),
        'content' => $item->content,
    ]
);
```

```php
$categoryModel = config('rapidez.models.category');
$categoryModelInstance = new $categoryModel;
$categories = config('rapidez.models.category')::select($categoryModelInstance->qualifyColumns(['entity_id', 'name', 'url_path', 'children_count']))
    ->whereNotNull('url_key')->whereNot('url_key', 'default-category')
    ->where('children_count', '>', 0)
    ->get() ?? [];
    
$this->indexAllStores('categories', $categories, ['name', 'url_path', 'children_count'], 'entity_id');
```


src/Commands/ElasticsearchIndexCommand.php
```diff
@@ -0,0 +1,80 @@
...
+     */
+    public function indexStore($store, $indexName, $data, $values)
+    {
+        $this->line('Indexing `'.$indexName.'` for store: '.$store->name);
+        try {
+            [$alias, $index] = $this->prepareIndexer($store, $indexName);
+
+            $currentData = is_callable($data)
+                ? $data(config()->get('rapidez.store_code'))
+                : $data;
```
indykoning, line 42: Laravel has a function to handle this for you already 🙂 
https://github.com/laravel/framework/blob/82875a158e8f130a9103e096419f8518161b34d3/src/Illuminate/Collections/helpers.php#L186
```suggestion
            $currentData = value($data, config()->get('rapidez.store_code'));
```

src/Commands/ElasticsearchIndexCommand.php
```diff
@@ -0,0 +1,80 @@
...
+        [$alias, $index] = $this->createAlias($store, $indexName);
+        $this->createIndex($index);
+        return [$alias, $index];
+    }
+
+    public function indexItem($index, $item, $id, $values)
+    {
+        $currentValues = is_callable($values)
+            ? $values($item)
+            : collect($item)->only($values);
```
indykoning, line 72: If this is the only thing we do with the collection it may be more efficient to use the Arr::only() function which the collections only function uses as well https://github.com/laravel/framework/blob/82875a158e8f130a9103e096419f8518161b34d3/src/Illuminate/Collections/Collection.php#L912

Skips us having to put the array(able) into a collection first.

src/Commands/ElasticsearchIndexCommand.php
```diff
@@ -0,0 +1,80 @@
...
+abstract class ElasticsearchIndexCommand extends InteractsWithElasticsearchCommand
+{
+    /**
+     * Index all items in all stores
+     *
+     * @param  string $indexName
+     * @param  (callable(object): array)|array $data
+     * @param  (callable(object): array)|array $values
+     */
+    public function indexAll($indexName, $data, $values)
```
indykoning, line 18: I'm not quite convinced of the function name yet, what "All" does this index if we have not read this code?

src/Commands/InteractsWithElasticsearchCommand.php
```diff
@@ -45,4 +47,27 @@ public function switchAlias(string $alias, string $index): void
...
+
+    public function setStore(Store $store): void
+    {
+        config()->set('rapidez.store', $store->store_id);
+        config()->set('rapidez.website', $store->website_id);
+        $code = config('rapidez.models.store')::getCachedWhere(function ($store) {
+            return $store['store_id'] == config('rapidez.store');
+        })['code'];
+        config()->set('rapidez.store_code', $code);
+    }
```
indykoning, line 72: Do keep in mind that these functions do not do any interaction with Elasticsearch 😉 
Maybe this logic could become a trait, or some of this logic could be moved to a Facade(/The Rapidez facade)

src/Commands/ElasticsearchIndexCommand.php
Jade-GG: How about `indexAllStores`? 😇

src/Commands/ElasticsearchIndexCommand.php
```diff
@@ -0,0 +1,115 @@
...
+        return [$alias, $alias.'_'.Carbon::now()->format('YmdHis')];
+    }
+
+    public function prepareIndexer($store, $indexName, array $mapping = [], array $settings = []): array
+    {
+        Rapidez::setStore($store);
+        [$alias, $index] = $this->createAlias($store, $indexName);
+        $this->createIndex($index, $mapping, $settings);
+
+        return [$alias, $index];
```
royduin, line 113: Can't we just keep this in the class so we can just do something like:
- `$this->indexer->prepare('products', $store)`
- In a loop: `$this->indexer->index($data)`
- `$this->indexer->finish()`

royduin (changes requested):

src/Commands/ElasticsearchIndexCommand.php
Jade-GG: I agree, I think that something like this would be much more elegant.

src/Commands/ElasticsearchIndexCommand.php
```diff
@@ -0,0 +1,115 @@
...
+
+use Carbon\Carbon;
+use Exception;
+use Illuminate\Support\Arr;
+use Rapidez\Core\Facades\Rapidez;
+use Rapidez\Core\Jobs\IndexJob;
+
+abstract class ElasticsearchIndexCommand extends InteractsWithElasticsearchCommand
+{
+    protected bool $useJob = true;
```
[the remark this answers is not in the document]
Jade-GG: I've now changed it to just always use a job because there's no reason not to :)

src/Rapidez.php
```diff
@@ -61,4 +62,21 @@ public function fancyMagentoSyntaxDecoder(string $encodedString): object
 
         return json_decode(str_replace(array_values($mapping), array_keys($mapping), $encodedString));
     }
+
+    public function getStores(): Collection
+    {
+        $storeModel = config('rapidez.models.store');
+
+        return $storeModel::all();
```
indykoning, line 70: I'm guessing it is potentially a good idea to wrap this in an array cache in case this gets executed multiple times during the same run, so we don't need to fetch the store models each time.
```suggestion
        return cache()->store('array')->rememberForever('allStores', fn() => (config('rapidez.models.store'))::all());
```

royduin: They are probably already cached from here: https://github.com/rapidez/core/blob/8732ca140a3b899830902686531a30c66a73b64b/src/Models/Store.php#L28. Maybe create a `getCached()` there so it's also possible to get all stores.

indykoning (approved):

src/Commands/IndexProductsCommand.php
```diff
@@ -28,17 +26,10 @@ public function handle()
         IndexBeforeEvent::dispatch($this);
 
         $productModel = config('rapidez.models.product');
-        $storeModel = config('rapidez.models.store');
-        $stores = $this->argument('store') ? $storeModel::where('store_id', $this->argument('store'))->get() : $storeModel::all();
```
royduin, line 32: So we can not run the indexer per store anymore?

src/Commands/IndexProductsCommand.php
```diff
@@ -66,30 +56,31 @@ public function handle()
 
                 $showOutOfStock = (bool) Rapidez::config('cataloginventory/options/show_out_of_stock', 0);
 
-                $productQuery->chunk($this->chunkSize, function ($products) use ($store, $bar, $index, $categories, $showOutOfStock) {
-                    foreach ($products as $product) {
+                $productQuery->chunk($this->chunkSize, function ($products) use ($store, $bar, $categories, $showOutOfStock) {
+                    $this->indexer->index($products, function ($product) use ($store, $categories, $showOutOfStock) {
                         if (!$showOutOfStock && !$product->in_stock) {
-                            continue;
+                            return null;
```
royduin, line 62:
```suggestion
                            return;
```

Does the same right?

src/Commands/IndexProductsCommand.php
Jade-GG: oops, fixed along with the setStore/getStore functionality improvement 😇

src/Http/Middleware/DetermineAndSetShop.php
```diff
@@ -3,32 +3,21 @@
...
+            ? $request->get('_store')
+            : $request->server('MAGE_RUN_CODE');
 
-        // Find the store code and website by the default store id.
-        if (! isset($store)) {
-            $store = $storeModel::getCachedWhere(function ($store) {
-                return $store['store_id'] == config('rapidez.store');
-            });
-        }
+        $store = $storeCode ? Rapidez::getStore($storeCode) : Rapidez::getStore(config('rapidez.store'));
```
royduin, line 18:
```suggestion
        $store = Rapidez::getStore($storeCode ?: config('rapidez.store'));
```

PR #257, opened 2023-05-15
Directive as a simplistic way to get rid of references

Jade-GG:
This feels very dirty but I couldn't find a better way to grab the component instance in the directive.

Maybe there's a better name to use than v-element?


royduin:
We're keeping it as it, thanks for your effort 🚀

PR #260, opened 2023-05-19
Modify page size functionalitity on POP with URL parameter

jbclaudio:


resources/js/components/Listing/Listing.vue
```diff
@@ -43,6 +46,22 @@
...
+
+        methods: {
+            initPageSize() {
+                this.variables.pageSize = Turbo.navigator.location.searchParams?.get('pageSize') ?? this.$root.config.grid_per_page
+            }
+        },
+
+        created() {
+            this.initPageSize()
+            this.$watch('variables.pageSize', (pageSize) => {
```
royduin, line 61: Watchers can be defined within `watch: {}`

resources/js/components/Listing/Listing.vue
```diff
@@ -35,6 +35,9 @@
         data: () => ({
             loaded: false,
             attributes: useAttributes(),
+            variables: {
```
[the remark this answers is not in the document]
jbclaudio: If I don't add it within variables (you could also call it something else) then it's not reactive. I don't quite understand why.

resources/js/components/Listing/Listing.vue
jbclaudio: I've now put the watcher inside `watch()`. But because `pageSize` is only reactive within `variables` and `watch: {}` can't listen to `variables.pageSize`, I have now created a computed property for pageSize.

royduin:
https://github.com/rapidez/core/pull/279

PR #261, opened 2023-05-19
Improve render time, add performance variable

indykoning:
This change reduced the render function (With cpu throttling set to 8x) of ~120ms to ~20ms
most other components' render functions also take ~20ms

The ENV variable VITE_PERFORMANCE will allow us to enable additional performance markers. In this case just Vues performance markers. Adding visibility to metrics like these (Only works in Vue developer mode)
image


royduin:
Also nice to have on master?

indykoning:
Good one, see https://github.com/rapidez/core/pull/265

PR #262, opened 2023-05-23
Use turbo event instead of waiting for 1500ms for notification

Jade-GG:


royduin:
Could you also create a PR to master with this?

Jade-GG:
> Could you also create a PR to master with this?

https://github.com/rapidez/core/pull/263 😇

PR #269, opened 2023-05-25
Only refresh mask if it is missing in getMask

indykoning:


royduin:
Maybe we need an extra Dusk test for multiple products?

PR #270, opened 2023-05-25
Fix Slider

Roene-JustBetter:
Slider wasn't working correct with the EventListener.


indykoning (approved):

royduin:

Roene-JustBetter:
> Do we also need this on master?

@royduin I asked @indykoning but he told me that it is not necessary on Master because it is already set correct there.

PR #271, opened 2023-05-26
Add loggedin computed helper

Jade-GG:
Going from `$root.user` to `$root.user?.id` to check if you're logged in breaks a ton of stuff. There were even still some places in the core itself where this wasn't fixed, so I've also included those here.

### DIFF
--- a/resources/js/helpers.js
+++ b/resources/js/helpers.js
@@ -1,0 +1,9 @@
+Vue.mixin({
+    computed: {
+        loggedIn() {
+            return Boolean(this.$root.user?.id)
+        }
+    }
+})
+
+
--- a/resources/js/helpers.js
+++ b/resources/js/helpers.js
@@ -1,9 +1,0 @@
-Vue.mixin({
-    computed: {
-        loggedIn() {
-            return Boolean(this.$root.user?.id)
-        }
-    }
-})
-
-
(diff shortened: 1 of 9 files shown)


indykoning (approved):

resources/js/helpers.js
royduin, line 1: Why a helper and not directly on the Vue instance?

Jade-GG: I don't know what would be cleaner in this case. If we end up eventually having more and more little helper functions, would it maybe even be nice to have a specific helper "object" to put them all under?

royduin: "helpers" are currently here: https://github.com/rapidez/core/blob/master/resources/js/package.js#L87:L100, when this explodes we'll move it somewhere else

Jade-GG: Makes sense, I'll move it over 👍

resources/js/package.js
```diff
@@ -97,6 +97,10 @@ function init() {
         computed: {
             // Wrap the local storage in getter and setter functions so you do not have to interact using .value
             guestEmail: wrapValue(useLocalStorage('email', (window.debug ? 'wayne@enterprises.com' : ''), {serializer: StorageSerializers.string})),
+
+            loggedIn() {
+                return Boolean(this.$root.user?.id)
+            },
```
royduin, line 103:
```suggestion
            loggedIn: Boolean(this.$root.user?.id),
```

This doesn't work?

Jade-GG: Getting rid of the $root works, but it has to stay in its own function with the return unfortunately.

PR #272, opened 2023-05-26
Title helper directive

Jade-GG:


resources/views/layouts/title.blade.php
```diff
@@ -0,0 +1,9 @@
+<title>
+    @hasSection('title')
+        {{ Rapidez::config('design/head/title_prefix') }}
+        @yield('title')
+        {{ Rapidez::config('design/head/title_suffix') }}
+    @else
+        {{ Rapidez::config('design/head/default_title') }}
```
indykoning, line 7: just in case, this should not be prefixed and suffixed right?

royduin:
We don't need this on 0.x? And we need to update this table: https://docs.rapidez.io/1.x/configuration.html#magento

resources/views/layouts/title.blade.php
Jade-GG: Correct, the prefix and suffix only come into play when you have a title set. This follows the standard Magento behavior.

Jade-GG:
https://github.com/rapidez/core/pull/273
https://github.com/rapidez/docs/pull/37

😇

PR #279, opened 2023-06-08
Page size as url param

royduin:


jbclaudio (approved):

resources/js/components/Listing/Listing.vue
```diff
@@ -35,15 +35,11 @@
         data: () => ({
             loaded: false,
             attributes: useAttributes(),
+            pageSize: Turbo.navigator.location.searchParams.get('pageSize') || config.grid_per_page
```
indykoning, line 38: Maybe it works wel to use [useUrlSearchParams](https://vueuse.org/core/useUrlSearchParams), it automatically syncs the url with the data.
```suggestion
            pageSize: useUrlSearchParams('history', {'initialValue': {'pageSize': config.grid_per_page}}).pageSize
```

It uses replaceState so we might have to check wether it plays nice with Turbo
https://github.com/vueuse/vueuse/blob/main/packages/core/useUrlSearchParams/index.ts#L127

royduin: Nice! But let's stick with this for now as it doesn't need to be reactive

PR #280, opened 2023-06-19
Categories index command & small fix

Jade-GG:
As it turns out, eloquent models don't like being cast to an array explicitly with `(array)$item`, as I discovered here.

### DIFF
--- a/src/Commands/ElasticsearchIndexer.php
+++ b/src/Commands/ElasticsearchIndexer.php
@@ -25,31 +25,32 @@ class ElasticsearchIndexer
         $this->elasticsearch->indices()->delete(['index' => $index]);
     }
 
-    public function index(iterable|object $data, callable|array $mapping, callable|string $id = 'id'): void
+    public function index(iterable|object $data, callable|array|null $dataFilter, callable|string $id = 'id'): void
     {
         if (is_iterable($data)) {
-            $this->indexItems($data, $mapping, $id);
+            $this->indexItems($data, $dataFilter, $id);
         } else {
-            $this->indexItem($data, $mapping, $id);
+            $this->indexItem($data, $dataFilter, $id);
         }
     }
 
-    public function indexItems(iterable $items, callable|array $mapping, callable|string $id = 'id'): void
+    public function indexItems(iterable $items, callable|array|null $dataFilter, callable|string $id = 'id'): void
     {
         foreach ($items as $item) {
-            $this->indexItem($item, $mapping, $id);
+            $this->indexItem($item, $dataFilter, $id);
         }
     }
 
-    public function indexItem(object $item, callable|array $mapping, callable|string $id = 'id'): void
+    public function indexItem(object $item, callable|array|null $dataFilter, callable|string $id = 'id'): void
     {
         if (is_null($item)) {
             return;
         }
 
-        $currentValues = is_callable($mapping)
-            ? $mapping($item)
-            : Arr::only($item instanceof Model ? $item->toArray() : (array)$item, $mapping);
+        $arrItem = $item instanceof Model ? $item->toArray() : (array)$item;
+        $currentValues = is_callable($dataFilter)
+            ? $dataFilter($item)
+            : (is_null($dataFilter) ? $arrItem : Arr::only($arrItem, $dataFilter));
 
         if (is_null($currentValues)) {
             return;
--- a/src/Commands/ElasticsearchIndexer.php
+++ b/src/Commands/ElasticsearchIndexer.php
@@ -4,6 +4,7 @@ namespace Rapidez\Core\Commands;
 
 use Carbon\Carbon;
 use Cviebrock\LaravelElasticsearch\Manager as Elasticsearch;
+use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\Arr;
 use Rapidez\Core\Jobs\IndexJob;
 
@@ -48,7 +49,7 @@ class ElasticsearchIndexer
 
         $currentValues = is_callable($mapping)
             ? $mapping($item)
-            : Arr::only((array) $item, $mapping);
+            : Arr::only($item instanceof Model ? $item->toArray() : (array)$item, $mapping);
 
         if (is_null($currentValues)) {
             return;
--- a/src/Commands/ElasticsearchIndexer.php
+++ b/src/Commands/ElasticsearchIndexer.php
@@ -4,7 +4,7 @@ namespace Rapidez\Core\Commands;
 
 use Carbon\Carbon;
 use Cviebrock\LaravelElasticsearch\Manager as Elasticsearch;
-use Illuminate\Database\Eloquent\Model;
+use Illuminate\Contracts\Support\Arrayable;
 use Illuminate\Support\Arr;
 use Rapidez\Core\Jobs\IndexJob;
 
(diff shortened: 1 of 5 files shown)


Jade-GG:
Looks like Duster & Prettier aren't working in PRs? 🤔

src/Commands/IndexCategoriesCommand.php
```diff
@@ -0,0 +1,27 @@
...
+    protected $description = 'Index the Rapidez categories to Elasticsearch';
+
+    public function handle(): int
+    {
+        $categoryModel = config('rapidez.models.category');
+        $categoryModelInstance = new $categoryModel;
+        $categories = config('rapidez.models.category')::select($categoryModelInstance->qualifyColumns(['entity_id', 'name', 'url_path', 'children_count']))
+            ->whereNotNull('url_key')->whereNot('url_key', 'default-category')
+            ->where('children_count', '>', 0)
+            ->get() ?? [];
```
royduin, line 21: So the query currently only gets the categories for the current store? All stores will get the categories from the default one right?

royduin:
> Looks like Duster & Prettier aren't working in PRs? 🤔

@indykoning

src/Commands/IndexCategoriesCommand.php
```diff
@@ -0,0 +1,27 @@
...
+
+        $this->indexStores(Rapidez::getStores($this->argument('store')), 'categories', $categories, ['name', 'url_path', 'children_count'], 'entity_id');
```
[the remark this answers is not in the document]
royduin: 1. `$stores` = `Rapidez::getStores($this->argument('store'))`
2. `$indexName` = `'categories'`
3. `$items` = `$categories`
4. `$mapping` = `['name', 'url_path', 'children_count']`
5. `$id` = `'entity_id'`

So it's the mapping? Or isn't the name doing what it says?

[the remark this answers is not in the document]
Jade-GG: I guess it's not a good name for the variable, as it refers to which data is sent to the indexer.

src/Jobs/IndexJob.php
```diff
@@ -13,11 +14,11 @@ class IndexJob
     protected int $id;
     protected array $values;
 
-    public function __construct(string $index, int $id, array $values)
+    public function __construct(string $index, int $id, $values)
     {
         $this->index = $index;
         $this->id = $id;
-        $this->values = $values;
+        $this->values = ($values instanceof Model ? $values->toArray() : (array)$values);
```
indykoning, line 21:
```suggestion
        $this->values = ($values instanceof Arrayable ? $values->toArray() : (array)$values);
```
May be better since it explicitly ensures toArray exists without checking if it is that specific model type.

src/Commands/ElasticsearchIndexer.php
```diff
@@ -24,31 +25,32 @@ public function deleteIndex(string $index): void
...
             return;
         }
 
-        $currentValues = is_callable($mapping)
-            ? $mapping($item)
-            : Arr::only((array) $item, $mapping);
+        $arrItem = $item instanceof Model ? $item->toArray() : (array)$item;
+        $currentValues = is_callable($dataFilter)
+            ? $dataFilter($item)
+            : (is_null($dataFilter) ? $arrItem : Arr::only($arrItem, $dataFilter));
```
indykoning, line 53: This was not very readable to me at first glance, maybe we can extract it into a function or match statement e.g. 
```suggestion
        $currentValues = match (true) {
            is_callable($dataFilter) => $dataFilter($item),
            is_array($dataFilter) => Arr::only($arrItem, $dataFilter))
            default => $arrItem
         }
```

indykoning:
> Looks like Duster & Prettier aren't working in PRs? 🤔
You're right, maybe it has to do with repository access when working in forks? 
Maybe we should update the configuration to run on any branch in the core and stop it from running on PRs themselves

src/Commands/ElasticsearchIndexer.php
Jade-GG: Agreed, I've used the match statement for this now.

src/Commands/IndexCategoriesCommand.php
```diff
@@ -0,0 +1,32 @@
...
+        $this->indexStores(
+            Rapidez::getStores($this->argument('store')),
+            'categories',
+            fn() => config('rapidez.models.category')::select((new $categoryModel)->qualifyColumns(['entity_id', 'name', 'url_path', 'children_count']))
+                ->whereNotNull('url_key')->whereNot('url_key', 'default-category')
+                ->where('children_count', '>', 0)
+                ->get() ?? [],
+            fn($data) => Eventy::filter('index.category.data', $data),
+            'entity_id'
+        );
```
royduin, line 28: This isn't very readable, maybe change it to methods which can be chained or maybe typehint the parameters so we know what everything is. Like `'categories'`, that's the index name but I know that after looking at the `indexStores()` method.

```suggestion
        $this->indexStores(
            stores: Rapidez::getStores($this->argument('store')),
            indexName: 'categories',
            items: fn() => config('rapidez.models.category')::select((new $categoryModel)->qualifyColumns(['entity_id', 'name', 'url_path', 'children_count']))
                ->whereNotNull('url_key')->whereNot('url_key', 'default-category')
                ->where('children_count', '>', 0)
                ->get() ?? [],
            dataFilter: fn($data) => Eventy::filter('index.category.data', $data),
            id: 'entity_id'
        );
```

What do you think @indykoning?

Jade-GG: With regards to chaining, I don't think something like this would be that much cleaner:

```php
$this->indexItems(fn() => config('rapidez.models.category')::select((new $categoryModel)->qualifyColumns(['entity_id', 'name', 'url_path', 'children_count']))
        ->whereNotNull('url_key')->whereNot('url_key', 'default-category')
        ->where('children_count', '>', 0)
        ->get() ?? [])
    ->withStores($this->argument('store'))
    ->filter(fn($data) => Eventy::filter('index.category.data', $data))
    ->id('entity_id')
    ->doIndex('categories')
```

There's probably a slightly nicer way to structure it but this seems a bit less intuitive to me?

royduin: True true, typed params are cleaner. Any thoughts @indykoning?

indykoning: Yeah i agree typed params may be cleaner.
I think the reason readability may be struggling a little is the shorthand function with the query as function argument as it spans multiple lines since it needs to for readablility of the query itself.

As to how i'm passing the function through see: https://www.php.net/releases/8.1/en.php#first_class_callable_syntax

```suggestion
        $this->indexStores(
            stores: Rapidez::getStores($this->argument('store')),
            indexName: 'categories',
            items: $this->getCategories(...),
            dataFilter: fn($data) => Eventy::filter('index.category.data', $data),
            id: 'entity_id'
        );
...
public function getCategories()
{
    return config('rapidez.models.category')::query()
                ->select((new config('rapidez.models.category'))->qualifyColumns(['entity_id', 'name', 'url_path', 'children_count']))
                ->whereNotNull('url_key')
                ->whereNot('url_key', 'default-category')
                ->where('children_count', '>', 0)
                ->get()
}
```

However i am seeing the folowing query used in more places. Maybe it's an idea to turn it into a scope?
```php
                ->whereNotNull('url_key')
                ->whereNot('url_key', 'default-category')
                ->where('children_count', '>', 0)
```

src/Commands/IndexCategoriesCommand.php
```diff
@@ -0,0 +1,37 @@
...
+    protected $signature = 'rapidez:index:categories {store? : Store ID from Magento}';
+
+    protected $description = 'Index the Rapidez categories to Elasticsearch';
+
+    public function handle(): int
+    {
+        $this->indexStores(
+            stores: Rapidez::getStores($this->argument('store')),
+            indexName: 'categories',
+            items: $this->getCategories(...),
```
royduin, line 20: `...`?

Jade-GG: https://www.php.net/releases/8.1/en.php#first_class_callable_syntax 😇

royduin: Whoeps doh 🙃  what a magic ✨

royduin:
Thanks!! 💯 next step is a PR with the categories in the autocomplete?

PR #282, opened 2023-06-22
Custom product options

royduin:
So we've some challenges here:

1. How are we going to name models in the config?
> Fixed, we're going for underscores
2. Currently the options come later as they're loaded with GraphQL, as we've the models we could render them without? Good for CLS.
> Fixed, now loaded from the database instead of GraphQL
3. To show the selected options in the cart we're eager loading `items2.options` as it's almost impossible to join this into the `items` and because we'd like to move away from that; but: getting this with GraphQL would probably be much easier. And as you can see within the `QuoteItemOption`, we can't eagerload everything.
> For now we stick with `items2`, in the future we're migrating to GraphQL
4. Are we going to keep things backwards compatible or not?
> 

So for the cart: Our own endpoint vs GraphQL? -> Later GraphQL
> We did implement this earlier: https://github.com/rapidez/core/pull/63 but we've some challenges there as well; we can't join any data anymore with the `quote.items.select` Eventy filter + we also need to cache the response and only refresh the cart when something changes. So not on every page the minicart needs a call.

> When we stay with our own endpoint, do we keep `items` and `items2`? -> Yes `items2` for now

And for the frontend: GraphQL vs models? -> Models
> Models is faster and better for Lighthouse; GraphQL is easier.


royduin:
The Duster public/protected order could also be disabled @indykoning, bit weird to have the timestamp at the top: https://github.com/rapidez/core/pull/282/commits/e1ce4f71284ee84d5cacb03d31068aa1e1134725

resources/views/product/partials/queries/options.graphql
```diff
@@ -0,0 +1,55 @@
+{
+    products(filter: { sku: { eq: "{{ $sku }}" } }) {
```
indykoning, line 2: It's better to work with a named graphql query and variables here.
It _could_ allow the query to be [cached](https://graphql.org/learn/caching/) by Magento or Client.
Very much **could** i dont think magento does it and neither our client

royduin: The query is removed as we're getting it from the database now 🚀

resources/js/components/Product/AddToCart.vue
```diff
@@ -183,15 +199,45 @@ export default {
...
             Object.entries(this.options).forEach(([key, val]) => {
                 options.push({
                     option_id: key,
                     option_value: val,
                 })
             })
+
+            Object.entries(this.customOptions).forEach(([key, val]) => {
+                if (typeof val === 'string' && val.startsWith('FILE;')) {
+                    let values = val.split(';', 4)
```
indykoning, line 213: It might be nice for readability to use the destructuring assignment
```suggestion
                    let [prefix, name, type, data] = val.split(';', 4)
```

src/Models/ProductOptionTypeValue.php
```diff
@@ -0,0 +1,68 @@
...
+        return $this->belongsTo(config('rapidez.models.product_option'), 'option_id');
+    }
+
+    protected function title(): Attribute
+    {
+        return Attribute::make(
+            get: function () {
+                $titles = $this->titles->pluck('title', 'store_id')->toArray();
+
+                return $titles[config('rapidez.store')] ?? $titles[0];
```
indykoning, line 30: I think you could keep this as a collection and then get the first value using `$titles->first()`

src/Models/QuoteItemOption.php
```diff
@@ -0,0 +1,67 @@
...
+                ->title
+        )->shouldCache();
+    }
+
+    // It would be nice if we could make a HasMany relation from this so it's possible
+    // to eager load it but DB::raw() to do the explode within SQL can't be used.
+    protected function option(): Attribute
+    {
+        return Attribute::make(
+            get: fn () => config('rapidez.models.product_option')::find(explode('_', $this->code)[1])
```
indykoning, line 64: In this case we do not need explode specifically, we could use something like
```sql
SUBSTRING_INDEX(SUBSTRING_INDEX(`code`, '_', -2), '_', 2)
```
This removes everything before the first occurence of `_` and then removes everything after the next occurence of `_`
emulating the behavior of `explode('_', $this->code)[1]`

royduin: Sadly we can't use raw SQL within Eloquent relations + this will be removed when we migrate the cart to GraphQL

src/Models/ProductOptionTypeValue.php
royduin: Yea, doing that with `price` as we need multiple columns there but with the title, we just need the title for the current store. Besides the ID's there is nothing else in the table so that's why I choose to return the title directly.

royduin: But.. I've created a new macro for this to make it cleaner: https://github.com/rapidez/core/pull/282/commits/9cb70b1933b4eb1d1969a96a0e26444a199c7970

royduin:
One more review @indykoning? 😇

indykoning (approved):

PR #283, opened 2023-06-23
Changes to colors

JimmyHoenderdaal:


tailwind.config.js
```diff
@@ -1,41 +1,47 @@
...
+                inactive: "#64748B", // Inactive text color
+                highlight: "#F1F5F9", // Background highlight color
+                border: "#F1F5F9", // Border color
             },
             borderColor: {
-                DEFAULT: 'rgb(var(--border) / <alpha-value>)',
-                border: 'rgb(var(--border) / <alpha-value>)'
+                DEFAULT: "#F1F5F9", // Border color default so it gets used when only using border
             },
         },
```
royduin, line 35: So this is THE way to do it? We don't need the alpha values? We don't need CSS variables? Logical variabel names? Everybody agrees?

JimmyHoenderdaal: Those alpha values aren't necessary since you're defining the hexcodes in tailwind config itself which allows us to still use bg-opacity bg-primary/50 and so on. I've discussed these variable names with @jordythevulder and we think they are logical and they are pretty much the industry standard.

Roene-JustBetter (approved):

danielsmink (approved):

PR #288, opened 2023-06-27
Watch loadingCount instead of a computed prop to set loading

jbclaudio:


royduin:
This doesn't work or why the PR? Was introduced in https://github.com/rapidez/core/pull/233 @indykoning?

jbclaudio:
I looked at this along with @indykoning. The data property `loading` does not work via a computed property, so we have now thrown it into a watcher to set the variable.

royduin:

PR #289, opened 2023-06-28
Add helpers for adding domain to urls

indykoning:
Turned named urls into `route()` calls
use the `url()` helper when we're certain the url starts with a `/`
use the `to()` helper if the url might start with a relative path

How Laravel generates the domain for the url can be done in multiple ways:
```php
\Illuminate\Support\Facades\Url::formatHostUsing(fn($root, $route) => 'https://example.com');
\Illuminate\Support\Facades\Url::forceRootUrl('https://example.com');
```
or if you simply want to add a prefix to the url that is possible by using the `X-Forwarded-Prefix` header

As there are multiple ways to set this domain, like getting it from Magento, setting it in configuration, passing it from your reverse proxy i have not made any opinionated code on this.
Depending on the requirements of the project one of these can be used on a per-project basis.

Most projects will not need to change the host for url generation.

### DIFF
--- a/resources/js/app.js
+++ b/resources/js/app.js
@@ -122,7 +122,7 @@ function init() {
         methods: {
             search(value) {
                 if (value.length) {
-                    Turbo.visit('/search?q=' + encodeURIComponent(value))
+                    Turbo.visit(window.url('/search?q=' + encodeURIComponent(value)))
                 }
             },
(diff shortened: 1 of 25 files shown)


royduin:
Just need another one for master? 😇

PR #290, opened 2023-06-29
Ensure the cart has been cleared on logout

indykoning:
The cart wouldn't get cleared properly since the redirect happened before the cart had been properly cleared.
We now await the clearing before we redirect in the next tick, so watchers have time to respond to changes.

I also noticed that the onLogout event listener was added per instance of interactWithUser, causing this function to be called multiple times. This has been prevented using memoization.

And just in case we now check the current state of tokens before setting the data to the received response, to deter possible race conditions.


resources/js/components/User/mixins/InteractWithUser.js
```diff
@@ -1,6 +1,12 @@
-import { useLocalStorage } from '@vueuse/core'
+import { useLocalStorage, useMemoize } from '@vueuse/core'
```
Jade-GG, line 1: Is this being used? I don't see it in the diff

PR #296, opened 2023-07-19
Remove column aliases and qualify columns

Jade-GG:
add the `created_at` column to the global scopes


royduin:
Just add it where the defaults are instead of with a scope 😇

royduin:
So that would be in `WithProductAttributesScope` I think

src/Models/Scopes/Product/WithProductAttributesScope.php
```diff
@@ -26,7 +26,8 @@ public function apply(Builder $builder, Model $model)
             ->addSelect($builder->getQuery()->from . '.entity_id AS id')
             ->addSelect($builder->getQuery()->from . '.sku')
             ->addSelect($builder->getQuery()->from . '.visibility')
-            ->addSelect($builder->getQuery()->from . '.type_id AS type');
+            ->addSelect($builder->getQuery()->from . '.type_id AS type')
+            ->addSelect($builder->getQuery()->from . '.created_at');
```
indykoning, line 30:
```suggestion
            ->addSelect($model->getQualifiedCreatedAtColumn());
```
https://github.com/laravel/framework/blob/9c44052743b0ee7f3adea36994918b4d8019e8b3/src/Illuminate/Database/Eloquent/Concerns/HasTimestamps.php#L162

indykoning, line 30: OR nicer might be
```suggestion
            ->addSelect([
                'id' => $model->qualifyColumn('entity_id')
                'type' => $model->qualifyColumn('type_id')
            ])
            ->addSelect($model->qualifyColumns([
                'sku',
                'visibility'
                $this->getCreatedAtColumn()
            ]));
```

Jade-GG: I'm not sure on this one, it feels harder to read the code when it's split apart like this. It's less immediately obvious what's being selected in what way.

src/Models/Scopes/Product/WithProductAttributesScope.php
```diff
@@ -26,7 +26,8 @@ public function apply(Builder $builder, Model $model)
             ->addSelect($builder->getQuery()->from . '.entity_id AS id')
             ->addSelect($builder->getQuery()->from . '.sku')
             ->addSelect($builder->getQuery()->from . '.visibility')
-            ->addSelect($builder->getQuery()->from . '.type_id AS type');
+            ->addSelect($builder->getQuery()->from . '.type_id AS type')
```
indykoning, line 29: In that case i would suggest using the qualifyColumn functions, and selecting a column `AS` using the built in way in addSelect as it contains additional safeties
```suggestion
            ->addSelect(['id' => $model->getQualifiedKeyName()])
            ->addSelect($model->qualifyColumn('sku'))
            ->addSelect($model->qualifyColumn('visibility'))
            ->addSelect(['type' => $model->qualifyColumn('type_id')])
```

src/Models/Scopes/Product/WithProductAttributesScope.php
royduin: Agree with @Jade-GG

src/Models/Scopes/Product/WithProductAttributesScope.php
royduin: It's still harder to read?

royduin: It's master so here we can make bigger changes; maybe it's a good idea to get rid of the aliases? And just go with:
```php
$builder->addSelect([
    $model->getQualifiedKeyName(),
    $model->qualifyColumn('sku'),
    $model->qualifyColumn('visibility'),
    $model->qualifyColumn('type_id'),
    $model->getQualifiedCreatedAtColumn(),
]);
```

The aliases are just some simple replaces, it would be nice if we can track them down within the Rapidez repo's and where it's used also replace it with the qualify column. When we're going to update a project we'll notice it pretty fast; so no need to keep it backwards compatible.

Jade-GG: That also works for me, just gotta find and eliminate all the `id` and `type` usages which I'm sure are numerous.

src/Models/Product.php
```diff
@@ -96,7 +96,7 @@ public function views(): HasMany
         return $this->hasMany(
             config('rapidez.models.product_view'),
             'product_id',
-            'id'
+            'entity_id'
```
royduin, line 99: We don't need to specify this param for relations anymore now I think?

Jade-GG:
Looks like I had already found all of the cases where `id` and `type` were used. Makes me wonder why they were even aliased in the first place 🤔

royduin:
Haha, good question. At the time it was build we thought it was a good idea. In the future we need to refactor this also: https://github.com/rapidez/core/blob/master/src/Models/Scopes/Attribute/OnlyProductAttributesScope.php

Did you also check all packages? 😇

Jade-GG:
I can only find it in the wishlist/multiple-wishlist packages, I'll make a PR for those with a fallback for compatibility :)

Jade-GG:
@royduin Compatibility-wise, this is an issue: https://github.com/rapidez/multiple-wishlists/blob/master/resources/views/multiplewishlist/partials/wishlist.blade.php#L21

We could make this query look for both `id` or `entity_id` but that's very cumbersome. Do we just mark it as requiring the newest version of the core just for this?

royduin:
Just a PR is fine, as there are no releases for that package currently we'll keep it open until the project where the package is used will be upgraded to v1.

PR #298, opened 2023-07-20
Add categories into autocomplete

Jade-GG:
This is a simple implementation of adding categories into the standard autocomplete with a config toggle.


src/RapidezServiceProvider.php
```diff
@@ -60,6 +61,12 @@ protected function bootCommands(): self
             InstallTestsCommand::class,
         ]);
 
+        Event::listen(IndexBeforeEvent::class, function ($event) {
+            if (config('rapidez.autocomplete_categories')) {
+                $event->context->call('rapidez:index:categories');
+            }
+        });
+
```
Jade-GG, line 69: Do we want this to be dependent on the config setting? It feels to me like this could easily become spaghetti in a codebase with very custom frontend. The fact that it also auto-runs the indexer doesn't seem like it makes much sense from an outside perspective.

Jade-GG:
Will be adding a product_count to be indexed with the categories by default, as this is a very commonly recurring issue.

resources/js/components/Search/Autocomplete.vue
```diff
@@ -1,12 +1,63 @@
...
+        },
+    },
+
     render() {
-        return this.$scopedSlots.default()
+        return this.$scopedSlots.default({
+            results: this.results,
+            searchAdditionals: this.searchAdditionals,
+            additionals: this.additionals,
+        })
```
royduin, line 22:
```suggestion
        return this.$scopedSlots.default(Object.assign(this, { self: this }))
```

src/RapidezServiceProvider.php
royduin: Nope, you can remove the config. Let's just always index the categories 👍🏻

config/rapidez.php
```diff
@@ -32,6 +32,9 @@
         'customer_fields_show',
     ],
 
+    // Query the categories in the autocomplete & hook into rapidez indexer with categories index
+    'autocomplete_categories' => false,
+
```
royduin, line 37: No need to make it conditional, let's just always use it.

royduin:
@indykoning, time to review?

resources/js/components/Search/Autocomplete.vue
```diff
@@ -1,12 +1,59 @@
...
+    },
+
+    data() {
+        return {
+            results: { count: 0 },
+        }
+    },
+
+    methods: {
+        searchAdditionals(query) {
```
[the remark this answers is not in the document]
Jade-GG: It's possible, though it's a bit of a weird one. If you do it that way, the results get mixed together and the limit counts for both. This means you might just get 9 products and 0 categories, even though you're searching both.

Separating them like this beforehand means you can actually get results for both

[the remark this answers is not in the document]
royduin: True, better separated

royduin:
The is a conflict, when that's fixed and @indykoning approves we can merge.

indykoning (approved):

PR #299, opened 2023-07-24
Double check on localStorage.token existing

Jade-GG:
This is a quick fix for `useLocalStorage` not seeing a direct write to `localStorage.token`.

In my case, the issue arose stemming from https://github.com/rapidez/login-as-customer/blob/master/resources/js/components/LoginAsCustomer.vue#L45, causing the login-as-customer package to not work properly.

This can also be fixed by changing the login-as-customer package to import `useUser` and use `token.value`, however that is not backwards compatible with old rapidez.

### DIFF
--- a/resources/js/stores/useUser.js
+++ b/resources/js/stores/useUser.js
@@ -8,7 +8,11 @@ let isRefreshing = false
 
 export const refresh = async function () {
     if (!token.value) {
-        userStorage.value = {}
+        if (localStorage.token) {
+            token.value = localStorage.token
+        } else {
+            userStorage.value = {}
+        }
         return false
     }
 


indykoning:
I agree, unfortunately we cannot make those changes of using useUser just yet.
However vueUse itself is part of both branches of the core already.

A better, more 1.x approach would be to utilise vueUse's useStorage methods in order to write it to localstorage.
It will handle syncing all required values automatically.
https://github.com/vueuse/vueuse/blob/4f4141ad3ce9bbb86dbf57c57ceb3792a28ffdd6/packages/core/useStorage/index.ts#L196

So while it would be nicer to be able to use "useUser" the best solution right now is to do
```js
useLocalStorage('token', '').value = tokenResponse.data.data.generateCustomerTokenAsAdmin.customer_token
```

Which has the added benefit of automatically triggering a refresh on the 1.x branch
https://github.com/rapidez/core/blob/ffaa4c10b54b30787a166a6a8822d643e7a48196/resources/js/stores/useUser.js#L60

Jade-GG:
https://github.com/rapidez/login-as-customer/pull/5 tested and verified 😇

PR #300, opened 2023-07-25
Add race condition fallback in onOnce

Jade-GG:
Right now, logging out usually doesn't work because the `logout` event isn't captured---`window.app.$on` appears to be unavailable available in the created hook, nor in the mounted hook, of a mixin like this.

This was the cleanest workaround I could come up with, this will wait until every other thread is completed by basically putting a new thread at the bottom of the stack.

### DIFF
--- a/resources/js/components/User/mixins/InteractWithUser.js
+++ b/resources/js/components/User/mixins/InteractWithUser.js
@@ -2,13 +2,8 @@ import { useLocalStorage, useMemoize } from '@vueuse/core'
 import { user, token, refresh as refreshUser, clear as clearUser } from '../../../stores/useUser'
 
 const onOnce = useMemoize(
-    (eventName, callback) => {
-        if(window.app?.$on) {
-            window.app.$on(eventName, callback)
-        } else {
-            // App isn't available yet, wait for vue to load in and then try.
-            window.setTimeout(() => window.app.$on(eventName, callback))
-        }
+    (self, eventName, callback) => {
+        self.$root.$on(eventName, callback)
     },
     {
         getKey: (eventName, callback) => eventName + callback.toString(),
@@ -116,7 +111,7 @@ export default {
     },
 
     created() {
-        onOnce('logout', this.onLogout)
+        onOnce(this, 'logout', this.onLogout)
     },
 
     asyncComputed: {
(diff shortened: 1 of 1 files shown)


royduin:
This wasn't an issue earlier, so what's changed causing this not to work anymore?

indykoning:
Are you sure this is what is causing the problem you are experiencing?

useMemoize will only execute the function when the returned function is called.

The onOnce function is only ever called in the `created` hook, which could only be called if Vue exists and has been initialised.

Jade-GG:
You'd think Vue is already initialized, however in both using `created()` and `mounted()` I found the `app` object to be empty by the time this gets called. I haven't been able to figure out why this happens, it's very possible there's some weird reason why this happens in this case.

royduin:
Maybe investigate this further together with @indykoning?

Jade-GG:
I've made an alternative solution that's probably cleaner.

It seems as though `window.app` is not yet available in this context because this gets called on the stack where vue gets initialized. The code only sets `window.app` after it's fully initialized: https://github.com/rapidez/core/blob/master/resources/js/package.js#L53

We can also call the function by doing `window.Vue.prototype.$on`, however this of course doesn't carry the vm with it properly.

resources/js/components/User/mixins/InteractWithUser.js
```diff
@@ -111,7 +111,7 @@ export default {
     },
 
     created() {
-        onOnce('logout', this.onLogout)
+        onOnce(this, 'logout', this.onLogout)
```
indykoning, line 114: Maybe pass `this.$root` from here so we don't have to append `.$root` in the onOnce function.
This will allow the function to be used on components as well instead of just the root.

Still odd that it would be necessary though considering vues composition api and vueuse can interact with the vue container without issue

Jade-GG: > Still odd that it would be necessary though considering vues composition api and vueuse can interact with the vue container without issue

Yeah the order of operations here is weird. Something about how mixins are fully created and mounted before the Vue object is created. I debugged the actual vue code for a while and there really isn't a clean way around it without writing these kinds of mixins differently, which is a much bigger and possibly breaking update.

royduin:
So this can be merged @indykoning and @Jade-GG?

indykoning (approved):

PR #301, opened 2023-07-26
Tier pricing

Jade-GG:
Given that this has been necessary in more than one project, it might be a good idea to make this more generalized.


resources/js/components/Product/AddToCart.vue
```diff
@@ -171,6 +172,50 @@ export default {
...
+        },
+
+        tierPriceDiscount(price, tier) {
+            let discount = tier ?? this.tierPrices[this.tierPrice] ?? null;
+            if(!discount) {
+                return price
+            }
+
+            if (discount.percentage_value > 0) {
+                return price * (100 - discount.percentage_value) / 100
```
indykoning, line 197: Just to make sure, the discount should be calculated on the entire price? Not on the price of the qty defined on the tier prices?

e.g. product is 100
Tier price gives a discount of 10% on 5 qty
You order 9
Should the discount be 450 + 400 = 850 or 810?

Jade-GG: It should be 810, yes. Tier prices like this are not calculated like tax brackets 😉

indykoning:
Most of the tests are failing, could you check that out? 🙂

resources/js/components/Product/AddToCart.vue
```diff
@@ -171,6 +172,50 @@ export default {
...
+        getTierPrice(qty) {
+            qty = parseInt(qty)
+            let best = null
+            let bestQty = 0
+            Object.entries(this.tierPrices).forEach(([id, tier]) => {
+                if (tier.qty > bestQty && tier.qty <= qty) {
+                    best = id
+                    bestQty = tier.qty
+                }
+            })
```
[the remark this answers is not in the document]
Jade-GG: I was thinking about something similar for this but it really doesn't look cleaner to me at all. I feel like it makes it much harder to read what this code really does.

Jade-GG:
Now being implemented in the price component, so this PR is no longer necessary.

PR #302, opened 2023-07-27
Alternate hreflang tags

royduin:


src/Rapidez.php
```diff
@@ -96,6 +96,17 @@ public function setStore(Store|array|callable|int|string $store): void
         config()->set('rapidez.store_code', $store['code']);
         config()->set('rapidez.website', $store['website_id']);
         config()->set('rapidez.website_code', $store['website_code']);
+        config()->set('rapidez.group', $store['group_id']);
         config()->set('rapidez.root_category_id', $store['root_category_id']);
     }
+
+    public function withStore(Store|array|callable|int|string $store, callable $callback)
```
indykoning, line 103: I think `withStore` is a bit of an ambiguous name, `asStore` may be a better name.
Maybe it's nice to allow arguments to be passed in as well.

```suggestion
    public function asStore(Store|array|callable|int|string $store, callable $callback, ...$args)
```
```php
$result = $callback(...$args);
```

src/Rapidez.php
```diff
@@ -96,6 +96,17 @@ public function setStore(Store|array|callable|int|string $store): void
...
+    {
+        $initialStore = config('rapidez.store');
+        Rapidez::setStore($store);
+        $result = $callback();
+        Rapidez::setStore($initialStore);
```
indykoning, line 108: people could catch the exception later down the chain, and we want to restore the original store in that case as well
```suggestion
        try {
        	$result = $callback();
        } finally {
        	Rapidez::setStore($initialStore);
        }
```

PR #303, opened 2023-07-28
Implement tax display values from Magento settings

Jade-GG:
Based on this draft by @jbclaudio: https://github.com/rapidez/core/pull/252

Includes a pretty breaking refactor where the price section in the add-to-cart is completely replaced by the new price component.

Do we want to build a standard way to override this tax display functionality for custom logic? Given that we do customize this logic in some places it might be useful.


src/Rapidez.php
```diff
@@ -100,6 +102,38 @@ public function setStore(Store|array|callable|int|string $store): void
...
+    public function getTaxTable(): array
+    {
+        return Cache::rememberForever('tax_'.config('rapidez.store'), function () {
+            $table = [];
+            $rates = DB::table('tax_class')
+                ->join('tax_calculation', 'tax_calculation.product_tax_class_id', '=', 'tax_class.class_id')
+                ->join('tax_calculation_rate', 'tax_calculation_rate.tax_calculation_rate_id', '=', 'tax_calculation.tax_calculation_rate_id')
+                ->where('tax_country_id', Rapidez::config('tax/defaults/country', 'US'))
+                ->where('tax_region_id', Rapidez::config('tax/defaults/region', 0))
+                ->whereIn('tax_postcode', [null, '*', Rapidez::config('tax/default/postcode', null)])
```
Jade-GG, line 114: Do we want to have the full tax calculation table here instead? Right now this will only base your tax rates off of the country the shop has set as its default, which might be problematic for more international shops. I'm not well-versed enough in tax law to know how this is supposed to work 😅

indykoning:
If this does have major breaking changes please open a PR to the docs as well, describing how to upgrade from 0.X
https://github.com/rapidez/docs/blob/master/src/1.x/upgrading.md

resources/js/components/Product/Price.vue
```diff
@@ -0,0 +1,130 @@
...
+                }
+            })
+
+            return addition
+        }
+    },
+
+    computed: {
+        specialPrice() {
+            if(!this.mounted) {
```
indykoning, line 89: What purpose do all these mounted checks serve?
So long as we have the product data, and the locations we can calculate _a_ price. And i don't see any logic being executed in the mounted method other than setting mounted to true.

resources/js/components/Product/Price.vue
```diff
@@ -0,0 +1,130 @@
...
+            }
+            return this.calculatePrice(this.product, this.type, this.options)
+        },
+
+        isDiscounted() {
+            if(!this.mounted) {
+                return false;
+            }
+
+            return this.specialPrice != this.price
```
indykoning, line 107:
```suggestion
            return this.specialPrice !== this.price
```

resources/js/components/Product/Price.vue
```diff
@@ -0,0 +1,130 @@
...
+            return addition
+        }
+    },
+
+    computed: {
+        specialPrice() {
+            if(!this.mounted) {
+                return;
+            }
+            return this.calculatePrice(this.product, this.type, Object.assign(this.options, { special_price: true }))
```
indykoning, line 92: Is this what Magento defines as the special price, or the final price?

https://magento.stackexchange.com/questions/279738/what-is-difference-between-price-special-price-and-final-price-in-magento

I think final price may be more descriptive of what you're actually getting here.

resources/js/mixins.js
```diff
@@ -13,5 +13,15 @@ Vue.mixin({
...
+
+        includeTaxAt(location) {
+            return location === true || location === false
+                ? location
+                : (window.config.tax.display[location] ?? 0) >= 1
+        },
+
+        decideTax(including, excluding, location) {
+            return this.includeTaxAt(location) ? including : excluding
+        },
```
indykoning, line 25: If a Mixin is used in multiple components my preference is to load that mixin inside those components, like in https://github.com/rapidez/core/blob/master/resources/js/components/Cart/Cart.vue#L2

resources/views/components/price.blade.php
```diff
@@ -0,0 +1,30 @@
...
+                <template v-if="range.min != range.max">
+                    @{{ range.min | price }} - @{{ range.max | price }}
+                </template>
+                <template v-else>
+                    @{{ range.min | price }}
+                </template>
+            </span>
+        </template>
+        <template v-else>
+            <span v-text="$options.filters.price(specialPrice || price)">
```
indykoning, line 19: the specialPrice computed attribute already returns the price if no special price is entered.
(Which is also why it may be better to call it finalPrice)

resources/js/components/Product/Price.vue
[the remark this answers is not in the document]
Jade-GG: It's essentially the same thing, yeah. This seemed clearer to me but I'm happy with either if you have a preference.

resources/views/cart/overview.blade.php
```diff
@@ -33,7 +33,7 @@ class="mx-auto"
                     </div>
                 </div>
                 <div class="w-2/6 sm:w-1/6 lg:w-1/12 text-right pr-5">
-                    @{{ item.price | price }}
+                    @{{ $root.decideTax(item.price, item.price_excl_tax, 'cart_price') | price }}
```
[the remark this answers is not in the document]
Jade-GG: Maybe. It seemed extremely overkill to me to have the full price component for every little line just to get this exact same functionality. It'd be very verbose in the code and very inefficient, but it'd work. You'd still have to define the `cart_price` location code and essentially run the same function, but then within a component.

resources/js/mixins.js
Jade-GG: This kinda comes down to the above thing, do we want to overkill with a component above? If that's the solution we choose then I can move this to a dedicated mixin instead.

src/Rapidez.php
Jade-GG: Magento does this: https://github.com/magento/magento2/blob/2.4-develop/app/code/Magento/Tax/Model/ResourceModel/Calculation.php#L267

Basically, just calculates the rate on request in the backend by combining all the factors there (then caching it). The entire table isn't really exposed anywhere that way.

I'm not sure what the best approach for rapidez would be. Probably not exactly that?

resources/views/cart/overview.blade.php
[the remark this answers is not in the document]
indykoning: How inefficient is the code? Considering on catalog & search pages it could be ran over 50 times

It would help readability and consistency to use or rename the component.
With it's name as a new user i would assume it would handle all price changes, and it would be confusing seeing it used in one place and replaced with raw js expressions somewhere else.
And something i can see coming in the future is the functionality of the price component and this function drifting apart as  the price component receives more features

src/Rapidez.php
indykoning: That does not seem to be the bit of code responsible for the postcode it should use, that has already been placed inside of $request.
What i'm most interested in is how Magento would deal with prices on the catalog, as magento caches every page as well thus this bit of backend code would not be able to influence the prices directly.

Maybe we can place the default tax amount in the frontend, and once another postcode is added send an api request and update the tax amount based on what it responds with?

PR #304, opened 2023-08-01
Update Heroicons v2

jbclaudio:


royduin:
I think it's a good idea to add the `blade-ui-kit/blade-heroicons^2.0` requirement in the `composer.json` of all packages where this is changed. That way we notice incompatibilities directly when updating.

PR #309, opened 2023-08-03
Add slider loop option

jbclaudio:
This adds a loop capability to the slider. When you slide back to the first slide, the chunk of items is added to the front of the loop. So this also happens immediately when loading, because you start on the first slide.

Similarly, when you go to the last slide, the chunk is placed at the back of the slider. This is the same thing that, for example, Swiper also does for loop generation.

### DIFF
--- a/resources/js/components/Elements/Slider.vue
+++ b/resources/js/components/Elements/Slider.vue
@@ -15,44 +15,55 @@ export default {
     props: {
         reference: {
             type: String,
-            default: 'slider',
+            default: 'slider'
         },
         vertical: {
             type: Boolean,
-            default: false,
+            default: false
         },
         interval: {
             type: Number,
-            default: 5000,
+            default: 5000
         },
         autoplay: {
             type: Boolean,
-            default: false,
+            default: false
         },
         bounce: {
             type: Boolean,
-            default: false,
+            default: false
         },
         stopOnHover: {
             type: Boolean,
-            default: true,
+            default: true
         },
+        loop: {
+            type: Boolean,
+            default: false
+        }
     },
     data: () => {
         return {
+            slider: '',
+            slidesTotal: 0,
             position: 0,
             showLeft: false,
             showRight: false,
             mounted: false,
             hover: false,
             direction: 1,
+            chunk: '',
             pause: () => {},
-            resume: () => {},
+            resume: () => {}
         }
     },
     mounted() {
+        this.initSlider()
         useEventListener(this.slider, 'scroll', useThrottleFn(this.scroll, 150, true, true), { passive: true })
         this.$nextTick(() => {
+            if (this.loop) {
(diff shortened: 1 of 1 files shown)


royduin:
Any more info? 😇

resources/js/components/Elements/Slider.vue
```diff
@@ -102,6 +118,19 @@ export default {
                 ? this.slider.scrollTo(0, this.slider.children[index]?.offsetTop)
                 : this.slider.scrollTo(this.slider.children[0]?.offsetWidth * index, 0)
         },
+
+        handleLoop() {
+            if (this.currentSlide + 1 === this.slidesTotal - 1) {
+                Array.from(this.chunk.children).forEach((child) => {
+                    this.slider.appendChild(child.cloneNode(true))
+                })
+            }
```
Jade-GG, line 127: given that slidesTotal is static, this seems like it'll continuously create duplicates if you go back and forth over the first loop point? I feel like this should be `=== (this.slider.children?.length ?? 1) - this.slidesVisible`

It's tricky to imagine by just reading the code so I could be wrong here :)

jbclaudio: slidesTotal is no longer static. In the method initSlider this will be set: https://github.com/rapidez/core/pull/309/files#diff-2d53f035dc6e0201915f81219d324c2479c3acd61fde0b67ffbc091205965494R76. Then I added a watcher for currentSlide where initSlider will be triggered: https://github.com/rapidez/core/pull/309/files#diff-2d53f035dc6e0201915f81219d324c2479c3acd61fde0b67ffbc091205965494R144. So if currentSlide changes, the slidesTotal will be set again.

Jade-GG: Ah, the second part is what I missed 😇

Jade-GG (approved):

resources/js/components/Elements/Slider.vue
```diff
@@ -72,13 +88,13 @@ export default {
...
-            this.hover = useElementHover(this.slider)
+            this.hover = useElementHover(this.$el)
         },
 
         scroll() {
             this.position = this.vertical ? this.slider.scrollTop : this.slider.scrollLeft
-            this.showLeft = this.position
-            this.showRight = this.slider.offsetWidth + this.position < this.slider.scrollWidth - 1
+            this.showLeft = this.loop ? this.loop : this.position
+            this.showRight = this.loop ? this.loop : this.slider.offsetWidth + this.position < this.slider.scrollWidth - 1
```
royduin, line 97:
```suggestion
            this.showLeft = this.loop || this.position
            this.showRight = this.loop || this.slider.offsetWidth + this.position < this.slider.scrollWidth - 1
```
This should do the same, but test it 😇

royduin (commented):
We also need this on `master` 😇

PR #310, opened 2023-08-07
Index refactor with ~~chaining~~ DTO

Jade-GG:
Significantly changing the indexer commands one more time.

Allows it to be used as such:

```php
$this->onlyStores($this->argument('store'))
    ->withFilter(fn ($data) => Eventy::filter('index.category.data', $data))
    ->chunk(25)
    ->index(
        indexName: 'categories',
        items: $this->getCategories(...),
    );
```

The built in functionality mostly came out of the wish to make the products indexer cleaner---the result of which you can also see in this PR.


src/Commands/ElasticsearchIndexCommand.php
```diff
@@ -4,55 +4,255 @@
...
+        }
+
+        $count = is_iterable($fullItems) ? count($fullItems) : $fullItems->getQuery()->getCountForPagination();
+
+        $this->beforeIndex($count);
+        $this->tryIndexItems($fullItems, $id);
+        $this->afterIndex();
+    }
+
+    public function tryIndexItems(iterable|Builder $items, callable|string $id): void
```
indykoning, line 177: I think these functions would be better suited in a class managing elasticsearch instead of an abstract command

src/Commands/ElasticsearchIndexCommand.php
```diff
@@ -4,55 +4,255 @@
...
+    public function indexPartialIterable(iterable $items, callable|string $id): void
+    {
+        if ($this->chunkSize < 1) {
+            $this->indexPartial($items, $id);
+
+            return;
+        }
+
+        $arrItems = $items instanceof Arrayable ? $items->toArray() : (array) $items;
+        foreach (array_chunk($arrItems, $this->chunkSize) as $chunk) {
```
indykoning, line 201: I feel like these chunked indexers are not the most efficient way to go about it. 
The opposite approach might be better.
Turning everything that isn't a [Generator](https://www.php.net/manual/en/class.generator.php) into a [Generator](https://www.php.net/manual/en/language.generators.syntax.php) in their most efficient way.

In case of a Builder instance, by using lazy
in case of anything paginated/chunked by looping and then advancing the pointer
In case of an array by simply looping over it or yielding it

If we need extra functionality we could wrap it in a LazyCollection https://laravel.com/docs/10.x/collections#lazy-collections

src/Commands/ElasticsearchIndexer.php
```diff
@@ -34,19 +36,23 @@ public function index(iterable|object $data, callable|array|null $dataFilter, ca
...
 
-    public function indexItem(object $item, callable|array|null $dataFilter, callable|string $id = 'entity_id'): void
+    public function indexItem(mixed $item, callable|array|null $dataFilter = null, callable|string $id): void
     {
         if (is_null($item)) {
             return;
         }
 
+        if (! $this->prepared) {
+            throw new Exception('Attempted to index items without preparing the indexer first.');
```
indykoning, line 53: Direct exceptions are difficult to catch in an isolated way, maybe there is a more appropriate exception for this in the Elasticsearch packages

src/Commands/ElasticsearchIndexCommand.php
Jade-GG: I agree with this, however the problem I have with doing that is that the progress bar logic (coming all the way from `indexPartial`) would then be moved out of the command which doesn't make sense to me.

I might be able to add some callback parameter to advance the bar? Doesn't feel very clean to me either. Could also just get rid of the bar entirely but it's nice to have.

indykoning: You're right in that when using this in the current way. However when using a lazy collection we can cheat a little bit by using the map method on the lazy collection.

The map method on a Lazy collection is different than that of a regular collection. 
It will get executed once the data gets accessed, instead of being executed once called.

So each iteration of the foreach loop would call the map function

Jade-GG: Ah, a LazyCollection is just essentially a generator wrapper I see. Should make it a lot easier to get a clean solution here 🤔

src/Commands/ElasticsearchIndexCommand.php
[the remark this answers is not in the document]
indykoning: Yes, the variables could be set using the constructor as well but i'm talking about a full DTO where we have a class with it's only purpose being keeping the settings & data. 
https://medium.com/@eloufirhatim/what-is-dto-and-how-to-use-it-in-a-laravel-application-7ca1e9045985

The command is becoming quite bloated with things that shouldn't be its responsibility.
What i'm thinking is something like

```php
public function handle()
{
    $settings = new ElasticsearchIndexSettings(
        indexName: '…',
        items: …,
        mapping: [ ... ],
        filter: fn ($data) => ...
    );

    $settings->addMapping([ ... ]);
    $settings->setChunkSize(1000)

    $this->index($settings);
}
```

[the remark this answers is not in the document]
Jade-GG: Hmm, I'd at least keep the `items` inside of the index function itself I think, but I can give that a whirl.

indykoning (approved):

royduin:
Please create a new PR with just the synonyms within the current way of indexing. We'll do the refactor later as this isn't how I like to see it. Reference: RAP-446

PR #311, opened 2023-08-10
Fix for saving 'same_as_billing' is always set to 1

jbclaudio:


royduin:
> When fixed please also create a PR for the `master` branch.

😇

PR #315, opened 2023-08-17
Add VAT field to address based on magento config

Jade-GG:


resources/views/checkout/partials/address.blade.php
```diff
@@ -122,4 +122,15 @@
             />
         </div>
     @endif
+    @if(Rapidez::config('customer/address/taxvat_show', 0))
```
royduin, line 125: `0` is the default value if nothing is configured? Or `opt` like below?

Jade-GG: The default is indeed to not show this field, hence the 0.

The one below can also have default '0' and it'd work the same, if that would make it cleaner.

PR #320, opened 2023-08-23
Add ErrorCallback to GraphqlMutation

jbclaudio:
This adds the ability to execute a callback when there's an error in the GraphQL mutation response.

### DIFF
--- a/resources/js/components/GraphqlMutation.vue
+++ b/resources/js/components/GraphqlMutation.vue
@@ -39,6 +39,9 @@ export default {
         callback: {
             type: Function,
         },
+        errorCallback: {
+            type: Function,
+        },
         beforeRequest: {
             type: Function,
         },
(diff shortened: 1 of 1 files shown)


jbclaudio:
I've also added this to `master`: https://github.com/rapidez/core/pull/321

resources/js/components/GraphqlMutation.vue
```diff
@@ -128,6 +131,11 @@ export default {
                 )
 
                 if (response.data.errors) {
+                    if (this.errorCallback) {
+                        await this.errorCallback(this.data, response)
+                        return
```
indykoning, line 136: Do we want to block the logout redirect and notifications when we have an error callback added?

Do note that it might also be possible to watch `this.error` depending on your goal

PR #324, opened 2023-08-24
Fixed wrong step number event being emitted

Jade-GG:
This is also hardcoded and we need to get rid of hardcoded stuff, but for now this is a solution to incorrect dataLayer information.

### DIFF
--- a/resources/js/components/Checkout/Checkout.vue
+++ b/resources/js/components/Checkout/Checkout.vue
@@ -37,7 +37,7 @@ export default {
         this.setCheckoutCredentialsFromDefaultUserAddresses()
         this.getShippingMethods()
         this.getTotalsInformation()
-        this.$root.$emit('checkout-step', 1)
+        this.$root.$emit('checkout-step', 2)
     },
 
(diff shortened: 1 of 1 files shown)


royduin:
Sure? As of: https://github.com/rapidez/core/blob/master/config/rapidez.php#L51 it's the correct number

Jade-GG:
> Sure? As of: https://github.com/rapidez/core/blob/master/config/rapidez.php#L51 it's the correct number

At least when combined with the checkout theme (and any other checkout that follows the same steps), the emitted step number is incorrect. We get step 1 on the credentials page, where it should normally be step 2.

It's only being noticed now because this event hadn't been captured for a while until the recent inclusion of GTM in certain projects. This caused the GTM data to be incorrect.

indykoning:
Does this cause Rapidez without the checkout theme to still work as intended? 
i expect this would mean you skip the login step by default which would be unintended for projects not using the checkout theme https://github.com/rapidez/core/blob/master/config/rapidez/frontend.php#L22

PR #326, opened 2023-08-25
Make and use "filterable" attribute

Jade-GG:
When magento doesn't let you set a field as "filter", and yet you still want to filter something. With this update, it's actually possible.

### DIFF
--- a/config/rapidez.php
+++ b/config/rapidez.php
@@ -145,6 +145,10 @@ return [
         'slideover'        => 'z-50',
     ],
 
-    // Attribute codes of attributes that need to be filtered but can't be set as filterable in magento
-    'extra-filters' => [],
+    // From Magento only "Yes/No, Dropdown, Multiple Select and Price" attribute types
+    // can be configured as filter. If you'd like to have a filter for an attribute
+    // with, for example, the type of "Text", you can specify the attribute code here.
+    'additional_filters' => [
+        // Attribute codes
+    ],
 ];
--- a/config/rapidez.php
+++ b/config/rapidez.php
@@ -145,4 +145,6 @@ return [
         'slideover'        => 'z-50',
     ],
 
+    // Attribute codes of attributes that need to be filtered but can't be set as filterable in magento
+    'extra-filters' => [],
 ];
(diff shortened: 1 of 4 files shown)


src/Models/Attribute.php
```diff
@@ -12,11 +13,20 @@ class Attribute extends Model
...
     {
         static::addGlobalScope(new OnlyProductAttributesScope);
     }
 
+    protected function filterable(): CastsAttribute
+    {
+        return CastsAttribute::make(
+            get: fn () => $this->filter || in_array($this->code, config('rapidez.extra-filters')),
+        )->shouldCache();
+    }
```
royduin, line 28: We can just use an accessor on `filter`: https://laravel.com/docs/master/eloquent-mutators#defining-an-accessor

config/rapidez.php
```diff
@@ -145,4 +145,6 @@
         'slideover'        => 'z-50',
     ],
 
+    // Attribute codes of attributes that need to be filtered but can't be set as filterable in magento
+    'extra-filters' => [],
```
royduin, line 149:
```suggestion
    // From Magento only "Yes/No, Dropdown, Multiple Select and Price" attribute types
    // can be configured as filter. If you'd like to have a filter for example an attribute
    // with the type of "Text" you can specify the attribute code here.
    'additional_filters' => [
        // Attribute codes
    ],
```

src/Models/Attribute.php
```diff
@@ -17,6 +18,13 @@ protected static function booting()
         static::addGlobalScope(new OnlyProductAttributesScope);
     }
 
+    protected function filter(): CastsAttribute
+    {
+        return CastsAttribute::make(
+            get: fn ($value) => $value || in_array($this->code, config('rapidez.extra-filters')),
+        )->shouldCache();
+    }
```
royduin, line 26: `Attribute` vs `CastsAttribute`? Did you check the link? https://laravel.com/docs/master/eloquent-mutators#defining-an-accessor

Jade-GG: Yeah see the thing is that I can't just use the class using its name `Attribute` because, well, the class we're on is already called that :)

PR #327, opened 2023-08-25
Fix error when adding certain kinds of item to the cart

Jade-GG:
...and avoid hard-erroring out of adding items in the future. In theory this should catch everything.

This broke the rapidez demo because it tried to split "attributes"

### DIFF
--- a/src/Models/QuoteItemOption.php
+++ b/src/Models/QuoteItemOption.php
@@ -25,6 +25,7 @@ class QuoteItemOption extends Model
         return Attribute::make(
             get: fn (string $value) => match ($this->code) {
                 'info_buyRequest' => json_decode($value),
+                'attributes'      => json_decode($value),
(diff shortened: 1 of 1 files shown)


src/Models/QuoteItemOption.php
```diff
@@ -57,7 +58,7 @@ protected function label(): Attribute
     protected function option(): Attribute
     {
         return Attribute::make(
-            get: fn () => config('rapidez.models.product_option')::find(explode('_', $this->code)[1])
+            get: fn () => config('rapidez.models.product_option')::find(explode('_', $this->code)[1] ?? '')
```
royduin, line 61: I think this is a better fallback
```suggestion
            get: fn () => config('rapidez.models.product_option')::find(explode('_', $this->code)[1] ?? $this->code)
```

PR #330, opened 2023-09-04
Remove useMemoize for event, use hacky solution instead

Jade-GG:
useMemoize caused a problem because the events got unloaded when switching pages (using Turbo), but the memory of useMemoize did not.

The workaround here is to just check if the event already exists, and if not, add it.


resources/js/components/User/mixins/InteractWithUser.js
```diff
@@ -111,7 +102,9 @@ export default {
     },
 
     created() {
-        onOnce(this.$root, 'logout', this.onLogout)
+        if(!this.$root._events?.logout?.length || !this.$root._events.logout[0].name.includes('onLogout')) {
+            this.$root.$on('logout', this.onLogout)
+        }
```
indykoning, line 107: Many other events could hook into the logout event, so it might be better to use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some instead of only checking the first listener

PR #331, opened 2023-09-04
Clear addresses along with cart clear

Jade-GG:
So that addresses don't stay alive when you log out


indykoning:
Dusk tests seem to fail, could you check that out?

Jade-GG:
> Dusk tests seem to fail, could you check that out?

Looking into it, not sure why they're failing as it works fine locally 😇

PR #341, opened 2023-09-21
Implement healthchecks

indykoning:
This is what validate looks like if it is successfull with only a warning
image
image

This is what it looks like if it encounters an error:
image
image

If another module needs to add checks that should be showed when running validate or checking the healthcheck page it can be done by listening to 
```
Event::listen('rapidez:health-check', static::class);
```
and returning an array with `healthy` and `messages`


src/RapidezServiceProvider.php
```diff
@@ -183,6 +185,8 @@ protected function bootTranslations(): self
     protected function bootListeners(): self
     {
         Event::listen(ProductViewEvent::class, ReportProductView::class);
+        MagentoSettingsHealthcheck::register();
+        ElasticsearchHealthcheck::register();
```
royduin, line 189: Maybe keep this the same? I haven't seen the registration in the listener yet, is that a new thing?

PR #343, opened 2023-09-25
Prevent the indexer from indexing invisible products

Jade-GG:
In theory, you'd think that magento should automatically remove products with visibility 1 from the flat table. However, there are still some circumstances where it does get put in there.

I'm uncertain about the exact specifics of this as I've had trouble consistently reproducing the different scenarios.


Jade-GG:
For 0.x see https://github.com/rapidez/core/pull/344

src/Commands/IndexProductsCommand.php
```diff
@@ -59,6 +59,10 @@ public function handle()
 
                 $productQuery->chunk($this->chunkSize, function ($products) use ($store, $bar, $categories, $showOutOfStock) {
                     $this->indexer->index($products, function ($product) use ($store, $categories, $showOutOfStock) {
+                        if ($product->visibility == 1) {
```
indykoning, line 62:
```suggestion
                        if ($product->visibility === 1) {
```

PR #345, opened 2023-09-26
Refresh user before retrieving addresses

Jade-GG:


indykoning (changes requested):
Should this be the responsibility of the function to set the addresses? i can see it happening that we'll end up re-retrieving the user data multiple times because we might need it earlier in the components running this function.

Jade-GG:
> Should this be the responsibility of the function to set the addresses?

It's probably a little overkill to put it specifically here in this case, but I do think Rapidez currently relies a bit too much on doing only a single request to magento and clinging on to that data for the rest of time. It happens way too often that rapidez assumes incorrect information and magento ends up rejecting everything.

Maybe there's some more ideal solution here with a more generalized approach to refreshing and invalidating data?

indykoning:
Absolutely the 0.x version of Rapidez caches some things too long, 1.x does resolve this to a point by moving the Cart & Users data to the sessionstorage instead of localstorage. This way reopening your browser will always get fresh info.

However i agree that we do want the latest user info in the checkout, and for this function.
But we shouldn't refresh the user info in this function itself, the function does not describe that it would refresh the user info. And it is not the only function needing this refreshed user info.

Plus the race condition bugs this would introduce due to the async nature of this function is why i would much rather refresh the user info in the place where we call this function, so we can call all functions that deal with the user's info in the `.then`
I do not want to have to implement https://github.com/rapidez/core/blob/98c2a6fbc170739bf4f05056e7d9dbe63ade039b/resources/js/stores/useCart.js#L18 on the userStore because every function dealing with user info does a network request

Jade-GG:
This version should be a bit cleaner then, also fixing the await issue :)

indykoning:
LGTM 🚀 
However all tests do seem to be failing consistently at the same test and same step.
Could you take a look?

Jade-GG:
> all tests do seem to be failing consistently at the same test and same step. Could you take a look?

I don't really get why because it works perfectly fine (i.e. the `dusk="email"` element shows up like it should) on my machine when I follow the exact same steps that the dusk test does. Very strange.

indykoning:
You can download the artifact under summary of the failing tests, and view the screenshots + logs

https://github.com/rapidez/core/actions/runs/6768576468?pr=345#:~:text=Show%20more-,Artifacts,-Produced%20during%20runtime

This might help with debugging

From what i can see the user gets redirected to login, even if it is a guest which doesn't have a user to refresh
https://github.com/rapidez/core/blob/008bb4a7adcfed989c4b1187831265e09faad3a4/resources/js/components/User/mixins/InteractWithUser.js#L10
Also indicates that it should redirect if the user is not logged in, do we want this in this care or should we check if the user info should be updated before requesting it to be updated

resources/js/components/Checkout/Checkout.vue
```diff
@@ -27,13 +27,20 @@ export default {
...
 
     methods: {
+        async setupCheckout() {
+            await this.refreshUser()
+
+            this.setCheckoutCredentialsFromDefaultUserAddresses()
+            this.getShippingMethods()
+            this.getTotalsInformation()
+
+            this.$root.$emit('checkout-step', 1)
```
indykoning, line 41: Do we need all these functions to wait until refreshUser is done, or are there some functions which can run independently of the refreshUser function?

If so it may be better to use .then and add the functions dependent on user info within the .then.
And remove the await.

This way no functions are obstructed from running until the user info call has finished

Jade-GG: Unfortunately we need the default credentials function to run before getting the shipping information and totals information, and for the default credentials we need this user refresh. We also don't want to emit the `checkout-step` event before these addresses are set correctly, so this is about as good as it gets here.

indykoning (approved):

PR #347, opened 2023-09-29
Add SearchQuery compatbility (0.x version)

rbnmulder:


src/Http/Controllers/SearchController.php
```diff
@@ -0,0 +1,28 @@
...
+use Illuminate\Support\Str;
+
+class SearchController
+{
+    public function __invoke(Request $request)
+    {
+        if ($searchQuery = Cache::rememberForever('search.query.' . Str::slug($request->q), function () use ($request) {
+            $searchQueryModel = config('rapidez.models.search_query');
+
+            return $searchQueryModel::where('query_text', $request->q)
```
indykoning, line 16: Does magento do an exact match by default or some kind of `LIKE` query?

src/Http/Controllers/SearchController.php
indykoning, line 1: How do these searchQuery results enter the database?
Does Magento create them immediately when a request is made to the search page, or are they created manually?

If Magento creates one if it does not exist, it may be a good idea to create them as well in Rapidez using 
https://laravel.com/docs/10.x/eloquent#retrieving-or-creating-models:~:text=Or%20Creating%20Models-,The,method,-will%20attempt%20to

Sadly this does mean we cannot filter the query to the existence of a redirect

src/Http/Controllers/SearchController.php
```diff
@@ -0,0 +1,28 @@
...
+                ->where('store_id', config('rapidez.store'))
+                ->whereNotNull('redirect')
+                ->first();
+        })) {
+            $searchQuery->increment('popularity');
+
+            return redirect($searchQuery->redirect);
```
indykoning, line 23: The redirect function without additional params is a temporary redirect, should we do a temporary or permanent redirect here?

If we want it permanent we could make it
```suggestion
            return redirect($searchQuery->redirect, 301);
```

src/Http/Controllers/SearchController.php
rbnmulder: I added this functionality, great tip! 👍🏼

src/Http/Controllers/SearchController.php
rbnmulder: Exact. When there is no `search_query` found it creates it in default Magento. We should have this covered with my latest change.

src/Http/Controllers/SearchController.php
```diff
@@ -0,0 +1,28 @@
...
+
+use Illuminate\Http\Request;
+use Illuminate\Support\Facades\Cache;
+use Illuminate\Support\Str;
+
+class SearchController
+{
+    public function __invoke(Request $request)
+    {
+        if ($searchQuery = Cache::rememberForever('search.query.' . Str::slug($request->q), function () use ($request) {
```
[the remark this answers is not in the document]
rbnmulder: @indykoning it seems the trimming of whitespace already happens by default.
The query should not be case sensitive. I will make it lowercase by default. This helps with the creation of `search_query`.

indykoning (approved):
Could you create a PR for the master version as well? :)

src/Http/Controllers/SearchController.php
```diff
@@ -0,0 +1,31 @@
...
+use Illuminate\Http\Request;
+use Illuminate\Support\Facades\Cache;
+use Illuminate\Support\Str;
+
+class SearchController
+{
+    public function __invoke(Request $request)
+    {
+        $query = Str::lower($request->q);
+        if ($searchQuery = Cache::remember('search.query.' . Str::slug($query), Carbon::now()->addHour(), function () use ($query) {
```
royduin, line 15: This could fill up Redis really quickly, do we really need the caching here? Nothing wrong with just a MySQL lookup?

src/Models/SearchQuery.php
```diff
@@ -0,0 +1,56 @@
...
+        'redirect',
+        'store_id',
+        'display_in_terms',
+        'is_active',
+        'is_processed',
+    ];
+
+    public static function boot()
+    {
+        parent::boot();
```
royduin, line 53: If you use booted you don't need to call the parent boot.

src/Models/SearchQuery.php
```diff
@@ -0,0 +1,56 @@
...
+    protected $fillable = [
+        'query_text',
+        'num_results',
+        'popularity',
+        'redirect',
+        'store_id',
+        'display_in_terms',
+        'is_active',
+        'is_processed',
+    ];
```
royduin, line 49: We're not using any of this; to keep it clean just remove it

src/Models/SearchQuery.php
```diff
@@ -0,0 +1,56 @@
...
+ * @property string|null $query_text
+ * @property int $num_results
+ * @property int $popularity
+ * @property string|null $redirect
+ * @property int $store_id
+ * @property int $display_in_terms
+ * @property int|null $is_active
+ * @property int|null $is_processed
+ * @property Carbon $updated_at
+ */
```
royduin, line 22: We don't follow this within Rapidez.

src/Http/Controllers/SearchController.php
```diff
@@ -0,0 +1,26 @@
...
+class SearchController
+{
+    public function __invoke(Request $request)
+    {
+        $query = Str::lower($request->q);
+        $searchQueryModel = config('rapidez.models.search_query');
+        if ($searchQuery = $searchQueryModel::firstOrCreate(
+            ['query_text' => $query],
+            ['store_id' => config('rapidez.store')]
+        )) {
```
royduin, line 17:
```suggestion
        if ($searchQuery = $searchQueryModel::firstOrCreate([
            'query_text' => $query,
            'store_id' => config('rapidez.store'),
        ])) {
```
The `store_id` should also be in the where right?

src/Http/Controllers/SearchController.php
```diff
@@ -0,0 +1,26 @@
...
+            $searchQuery->increment('popularity');
```
royduin, line 18: So when the query doesn't exists we're first inserting it and then we update it?

src/Http/Controllers/SearchController.php
```diff
@@ -0,0 +1,37 @@
...
+            return view('rapidez::search.overview');
+        }
+
+        $searchQuery->increment('popularity');
+        if ($searchQuery->is_active === 1 && $searchQuery->redirect) {
+            return redirect($searchQuery->redirect, 301);
+        }
+
+        return view('rapidez::search.overview');
+    }
```
royduin, line 36:
```suggestion
        $searchQuery = config('rapidez.models.search_query')::firstOrNew([
            'query_text' => Str::lower($request->q),
            'store_id'   => config('rapidez.store'),
        ], ['popularity' => 1]);

        if (! $searchQuery->exists) {
            $searchQuery->save();

            return view('rapidez::search.overview');
        }

        $searchQuery->increment('popularity');

        if ($searchQuery->is_active === 1 && $searchQuery->redirect) {
            return redirect($searchQuery->redirect, 301);
        }

        return view('rapidez::search.overview');
    }
```

PR #348, opened 2023-10-04
Fix slider reactivity when resized

Jade-GG:
Because Vue is not very smart and doesn't update a computed property when the bounds of an element changes, we need to go back to the stone age and use a ResizeObserver instead.

### DIFF
--- a/resources/js/components/Elements/Slider.vue
+++ b/resources/js/components/Elements/Slider.vue
@@ -55,6 +55,9 @@ export default {
             chunk: '',
             pause: () => {},
             resume: () => {},
+
+            childSpan: 0,
+            sliderSpan: 0,
         }
     },
     mounted() {
@@ -69,6 +72,10 @@ export default {
 
             this.initAutoPlay()
         })
+
+        this.updateSpan()
+
+        new ResizeObserver(this.updateSpan).observe(this.slider)
     },
     methods: {
         initSlider() {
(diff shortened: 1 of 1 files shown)


Jade-GG:
master branch counterpart: https://github.com/rapidez/core/pull/349

indykoning:
If you want the width and height of an element to be reactive you should probably [use the function for that: useElementSize](https://vueuse.org/core/useElementSize/)

This way your function can still be computed (Of course you'll need to create these sizes in the mounted function instead of the computed function itself otherwise Vue still can't watch it 😉)

Jade-GG:
One problem I'm having is that if I only use `useElementSize` on the slider, it won't update the `childSpan` computed property because it doesn't end up using the slider width, but the width of the first child.

I could then also attach a `useElementSize` to the child but that seems a bit excessive. Ideally I'd trick vue into just reacting to the slider change to update the computed value for childSpan :)

Jade-GG:
I ended up finding a way to trick Vue without it looking weird, however it turns out that I do need access to the actual callback to fix a bug to do with keeping currentSlide intact.

I've now changed it around slightly to use useResizeObserver to fix that bug.

indykoning:
I would definitely throttle or debounce the observer for performance reasons as we are doing a lot of DOM operations.
Though reading the code, aren't we really only interested in the childs dimensions?

The sliders dimensions only really serve as a fallback when no child exist, and all calculations wouldn't make sense without children anyways.

Jade-GG:
> aren't we really only interested in the childs dimensions

I figured we can reasonably assume that the parent slider being resized is enough to determine that we need to recalculate everything. Trying to aim for the child elements seems a little more risky.

The parent element size does actually matter here though, as the "position" of the slider is based in pixels which is relative to the size of the parent.

indykoning (approved):

PR #353, opened 2023-10-06
Product relations for categories

Jade-GG:
As it turns out, we can use the `catalog_category_product_index_storeX` tables which contains the visibilities of the products. This gives a ~4x speedup in the query if you just want a products count, as joining the whole flat table is pretty slow™. To make this easy I've just added a `productIndices` relation along with the `products` index to use in your withCount.


royduin:
Maybe `belongsToMany` and add the columns with `withPivot`?

royduin:
Mm, maybe a rename, `productRelations`, `productLinks` or `productPivot`? @indykoning thoughts?

src/Models/CategoryProduct.php
indykoning, line 1: CategoryProduct may conflict in name as there, i personally tend to stick close to the original table name like seen here https://github.com/Genaker/laragento/tree/main/src/Models

src/Commands/IndexCategoriesCommand.php
```diff
@@ -25,11 +25,11 @@ public function handle(): int
...
     public function getCategories()
     {
-        return config('rapidez.models.category')::query()
-            ->select((new (config('rapidez.models.category')))->qualifyColumns(['entity_id', 'name', 'url_path', 'children_count']))
+        return config('rapidez.models.category')::select((new (config('rapidez.models.category')))->qualifyColumns(['entity_id', 'name', 'url_path']))
+            ->withCount('productIndices')
             ->whereNotNull('url_key')
             ->whereNot('url_key', 'default-category')
-            ->where('children_count', '>', 0)
+            ->having('product_indices_count', '>', 0)
```
indykoning, line 32: In this case you probably do not have to go this complex.
By simply querying a relationships existence Laravel already creates a good query for you
https://laravel.com/docs/10.x/eloquent-relationships#querying-relationship-existence

Where it uses subselects in a single query, we do not need to know the count. Only that there is at least 1 visible product.

Considering you've already added your scope for visibility 
`->has('products')` should suffice.

Do be sure to dump the query just to make sure it gets built into a good query

Jade-GG: Ah, didn't know about this one. Looks like it works great & it's efficient enough that the `productIndices` relation is probably no longer necessary :)

src/Models/CategoryProduct.php
Jade-GG: It's tough because then we also should rename the Product and Category models to CatalogProduct and CatalogCategory. With how it's set up right now, I feel like this one specifically is more consistent with the others and if this is something we want to uphold in the future we should probably change this in a dedicated PR so it shows up in the changelog.

Jade-GG:
I've removed `productIndices` for now as the `has()` version of the query is much more efficient regardless of whether or not the products table is joined.

indykoning (approved):

src/Models/Category.php
```diff
@@ -65,6 +66,20 @@ public function subcategories()
...
+        return $this->hasManyThrough(
+            config('rapidez.models.product'),
+            config('rapidez.models.category_product'),
+            'category_id',
+            'entity_id',
+            'entity_id',
+            'product_id'
+        )
+            ->withoutGlobalScopes()
+            ->whereIn((new(config('rapidez.models.category_product')))->qualifyColumn('visibility'), [2, 4]);
```
royduin, line 80:
```suggestion
        return $this
            ->hasManyThrough(
                config('rapidez.models.product'),
                config('rapidez.models.category_product'),
                'category_id',
                'entity_id',
                'entity_id',
                'product_id'
            )
            ->withoutGlobalScopes()
            ->whereIn((new(config('rapidez.models.category_product')))->qualifyColumn('visibility'), [2, 4]);
```

PR #354, opened 2023-10-06
Use checkout-theme improvements in improve-frontend

Jade-GG:
Also a tiny merge of the master branch

### DIFF
--- a/resources/views/product/partials/options.blade.php
+++ b/resources/views/product/partials/options.blade.php
@@ -1,7 +1,9 @@
-<div class="flex flex-col space-y-3">
-    @foreach($product->options->sortBy('sort_order') as $option)
-        <div>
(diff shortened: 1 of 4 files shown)


resources/views/product/partials/options.blade.php
```diff
@@ -1,7 +1,9 @@
-<div class="flex flex-col space-y-3">
-    @foreach($product->options->sortBy('sort_order') as $option)
-        <div>
-            @include('rapidez::product.partials.options.'.$option->type_id)
-        </div>
-    @endforeach
-</div>
+@if(($product->options ?? false) && $product->options->count() > 0)
```
royduin, line 1: Does this need to be so explicit? Just checking the count should be enough right?
```suggestion
@if($product->options->isNotEmpty())
```

PR #355, opened 2023-10-06
Add auth guard for authorized customers

indykoning:
This adds a new "Auth Guard" to ensure logged in customers by either a JWT or customer oauth token.
It'll make it easier to create guarded routes only available to customers, it'll also add this customer to the context of the request.
usage:
```php
    Route::post('protected-route-for-customers', Controller::class)->middleware('auth:magento-customer');
```
after which you can use
```php
$customer = auth()->user();
```
to get the customer's info.

Alternatively if you do not want to guard the route against guests but would want the logged in customer's info you can use
```php
$customer = auth('magento-customer')->user();
```

When either of these are used, the user is added to the context in case errors are thrown:
Screenshot 2023-10-09 at 11 22 53

This uses the Bearer token, or "token" field on the request.
the "token" field can be any type of input, see: https://laravel.com/docs/10.x/requests#retrieving-an-input-value as it uses https://github.com/laravel/framework/blob/cddb4f3bb5231f44f18fce304dbf190ad8348ddc/src/Illuminate/Auth/TokenGuard.php#L97

PS: The fact that this is not the default auth provider, and you need to explicitly ask for it is on purpose.
Otherwise customer tokens could be used to log in to Laravel backends like Statamic.


src/Auth/MagentoCustomerTokenGuard.php
```diff
@@ -0,0 +1,58 @@
...
+    {
+        if (empty($credentials[$this->inputKey])) {
+            return false;
+        }
+
+        if ($this->retrieveByToken($credentials[$this->inputKey])) {
+            return true;
+        }
+
+        return false;
```
Jade-GG, line 51: What about:

```php
return boolval($this->retrieveByToken($credentials[$this->inputKey]));
```

indykoning:
In Rapidez not _really_ anything as of yet, we could use this to secure (mainly API) endpoints by customer.

We are using an alternative way for another project to authorize for Realtime orders/invoices & customer specific price sheets. Which do not pass these tokens to a Magento API thus we need to validate the customer ourselves.
After the addition of JWTs this needed to be overhauled.

I could see this being used for more in the future.

PR #356, opened 2023-10-09
Move config to files in Rapidez folder

indykoning:
This is a POC of moving the config to the Rapidez folder.
It works the same way as statamic does it, however for the base config there is still a config/rapidez.php


royduin:
Maybe move `config/rapidez.php` to `config/rapidez/system.php`? But that will cause a lot of replaces... `rapidez.store` => `rapidez.system.store`. Maybe just include `config/rapidez/system.php` in `config/rapidez.php`? That way we can still use `rapidez.store` and we can edit everything from `config/rapidez/*`

indykoning:
Because Laravel builds the config based on the paths in the config folder once published we will need to copy this rapidez.php over to the `/config` folder as well in order to load the `/config/rapidez/system.php`

royduin:
Just double checking; works with config caching?

indykoning:
Config caching (and optimize in general) seem to be working without issue

royduin:
Nice, just some conflicts 😇 and we also need to check all modules: https://github.com/search?q=org%3Arapidez%20config(%27rapidez.&type=code

indykoning:
From the search, the only packages being incompatible seem to be the 2 with the PRs above this comment

royduin:
Last thing to do is add a upgrade note?

PR #360, opened 2023-10-16
Implement basic version of input field component

Jade-GG:
Draft PR until the new version is implemented everywhere

Note that this is not 1-to-1 backwards compatible with previous installations of rapidez, not sure what the best approach is as you probably don't want `w-full` as a standard class on every input.

### DIFF
--- a/resources/views/checkout/partials/address.blade.php
+++ b/resources/views/checkout/partials/address.blade.php
@@ -15,7 +15,7 @@
     </graphql>
 </div>
 
-<div class="contents" v-if="!$root.loggedIn || !checkout.{{ $type }}_address.customer_address_id">
+<div class="contents [&>div>input]:w-full" v-if="!$root.loggedIn || !checkout.{{ $type }}_address.customer_address_id">
     <div class="col-span-12 {{ Rapidez::config('customer/address/middlename_show', 0) ? 'sm:col-span-4' : 'sm:col-span-6' }}">
         <x-rapidez::input
             label="Firstname"
@@ -86,6 +86,7 @@
     </div>
     <div class="col-span-12 sm:col-span-6">
         <x-rapidez::country-select
+            class="w-full"
             name="{{ $type }}_country"
             label="Country"
             v-model="checkout.{{ $type }}_address.country_id"
--- a/resources/views/components/input-field/checkbox.blade.php
+++ b/resources/views/components/input-field/checkbox.blade.php
@@ -1,0 +1,1 @@
+@include('rapidez::components.input-field.index', ['type' => 'checkbox'])
--- a/resources/views/components/input-field/index.blade.php
+++ b/resources/views/components/input-field/index.blade.php
@@ -1,0 +1,27 @@
+@props(['name', 'label', 'type' => 'text', 'disabled' => false, 'required' => false, 'placeholder' => false, 'dusk' => null])
+@slots(['input', 'label'])
+
+@php
+    $componentType = match($type) {
+        'select', 'textarea', 'checkbox' => 'rapidez::' . $type,
+        default => 'rapidez::input',
+    };
+    $labelClasses = match($type) {
+        'radio', 'checkbox' => '!flex-row',
+        default => '',
+    }
+@endphp
+
+<x-rapidez::label :attributes="$attributes->class($labelClasses)" :$label>
+    <x-dynamic-component :component="$componentType" :attributes="$input->attributes->merge([
+        'name' => $name ?? null,
+        'type' => $type,
+        'placeholder' => $placeholder,
+        'dusk' => $dusk,
+        'disabled' => $disabled,
+        'required' => $required,
+    ])">
+        {{ $input }}
+    </x-dynamic-component>
+    {{ $slot }}
+</x-rapidez::label>
--- a/resources/views/components/select.blade.php
+++ b/resources/views/components/select.blade.php
@@ -1,17 +1,9 @@
-@props(['name', 'label', 'wrapperClass', 'labelClass'])
+@props(['name' => null])
 
-<div class="{{ $wrapperClass ?? '' }}">
-    @if(!isset($label) || (isset($label) && $label))
-        <x-rapidez::label for="{{ $name ?? '' }}" class="{{ $labelClass ?? '' }}">
-            @lang($label ?? ucfirst($name ?? ''))
-        </x-rapidez::label>
-    @endif
-    <select {{ $attributes->merge([
-        'id' => $name ?? null,
-        'name' => $name ?? null,
-        'dusk' => $attributes->get('v-bind:dusk') ? null : $name ?? null,
-        'class' => 'w-full py-2 pl-3 pr-8 border-gray-200 rounded focus:ring-neutral focus:border-neutral sm:text-sm'
-    ]) }}>
-        {{ $slot }}
-    </select>
-</div>
+<select {{ $attributes->merge([
+    'class' => 'peer rounded border border-border bg-white py-4 px-5 text-sm outline-none !ring-0 transition-all placeholder:text-inactive disabled:bg-gray-50 disabled:cursor-not-allowed font-medium',
+    'id' => $name,
+    'name' => $name,
+]) }}>
+    {{ $slot }}
+</select>
--- a/resources/views/checkout/steps/login.blade.php
+++ b/resources/views/checkout/steps/login.blade.php
@@ -4,6 +4,7 @@
             <h1 class="font-bold text-4xl text-center mb-5">@lang('Checkout')</h1>
 
             <x-rapidez::input
+                class="w-full"
                 :label="false"
                 name="email"
                 type="email"
@@ -13,6 +14,7 @@
                 required
             />
             <x-rapidez::input
+                class="w-full"
                 v-if="!emailAvailable"
                 :label="false"
                 class="mt-3"
--- a/resources/views/components/input-field/radio.blade.php
+++ b/resources/views/components/input-field/radio.blade.php
@@ -1,0 +1,1 @@
+@include('rapidez::components.input-field.index', ['type' => 'radio'])
--- a/resources/views/components/input-field/select.blade.php
+++ b/resources/views/components/input-field/select.blade.php
@@ -1,0 +1,1 @@
+@include('rapidez::components.input-field.index', ['type' => 'select'])
--- a/resources/views/components/input-field/textarea.blade.php
+++ b/resources/views/components/input-field/textarea.blade.php
@@ -1,0 +1,1 @@
+@include('rapidez::components.input-field.index', ['type' => 'textarea'])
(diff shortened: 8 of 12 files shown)


resources/views/checkout/partials/address.blade.php
```diff
@@ -86,6 +86,7 @@
     </div>
     <div class="col-span-12 sm:col-span-6">
         <x-rapidez::country-select
+            class="w-full"
```
Roene-JustBetter, line 89: Do you need w-full on all inputs?

Roene-JustBetter (changes requested):

resources/views/components/input-field/index.blade.php
```diff
@@ -0,0 +1,27 @@
+@props(['name', 'label', 'type' => 'text', 'disabled' => false, 'required' => false, 'placeholder' => false, 'dusk' => null])
+@slots(['input', 'label'])
+
+@php
+    $componentType = match($type) {
+        'select', 'textarea', 'checkbox' => 'rapidez::' . $type,
+        default => 'rapidez::input',
+    };
+    $labelClasses = match($type) {
+        'radio', 'checkbox' => '!flex-row',
```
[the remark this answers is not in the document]
Jade-GG: It's to overwrite the flex-col-reverse, which unfortunately doesn't seem to happen on its own. But maybe if we [used twMerge instead...](https://github.com/gehrisandro/tailwind-merge-laravel) 😇

resources/views/checkout/partials/address.blade.php
```diff
@@ -15,7 +15,7 @@
     </graphql>
 </div>
 
-<div class="contents" v-if="!$root.loggedIn || !checkout.{{ $type }}_address.customer_address_id">
+<div class="contents [&>div>input]:w-full" v-if="!$root.loggedIn || !checkout.{{ $type }}_address.customer_address_id">
```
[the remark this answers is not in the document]
Jade-GG: It could probably just be a `template` to keep the `v-if` working, but I didn't want to mess with the current structure just yet.

royduin:
With https://github.com/gehrisandro/tailwind-merge-laravel it's alright to have `w-full` on the input as it's possible to overwrite

JimmyHoenderdaal (approved):

Roene-JustBetter (approved):

Roene-JustBetter:
Do you need to check the failed tests?

Jade-GG:
> Do you need to check the failed tests?

Can't do much about those, dusk tests have been broken for a bit and in this case specifically it looks like the php version that is used for them has to be updated. Fixing these should already be in the pipeline at some point in the near future :)

resources/views/components/input-field/checkbox.blade.php
```diff
@@ -0,0 +1 @@
+@include('rapidez::components.input-field.index', ['type' => 'checkbox'])
```
jordythevulder, line 1: Is the name "input" not enough? why add "field" behind it.

indykoning:
> > Do you need to check the failed tests?
> 
> Can't do much about those, dusk tests have been broken for a bit and in this case specifically it looks like the php version that is used for them has to be updated. Fixing these should already be in the pipeline at some point in the near future :)

I've opened the test results and these tests are failing because tailwind merge is not compatible with all testing scenarios.
Could you update the requirements or testing scenarios to resolve this?

When i look in the latest commits on master the tests don't seem to be broken at all.

Jade-GG:
> I've opened the test results and these tests are failing because tailwind merge is not compatible with all testing scenarios.

Yeah the tests use a version of php that's too old. I tried updating the composer.json but I don't think I have the ability to update the php version used in these tests 😇

royduin:
You can find the tests in `/.github/workflows` but it doesn't look like that's the problem?

Jade-GG:
`gehrisandro/tailwind-merge-laravel v0.2.1 requires php ^8.1.0 -> your php version (8.0.30) does not satisfy that requirement.`

Not sure what else to make of that 🤔

Jade-GG:
Ah right, these are specific php 8.0 tests, I guess those will have to be replaced by php8.2 tests 😇

resources/views/components/input-field/checkbox.blade.php
Jade-GG: Two reasons:
 - We wanted to have a separate input/label component as well as this one which combines both, giving people the freedom to build things up on their own
 - If you call a composite component like this just `input` then it becomes confusing, as you'd expect the base element of this component to be an input.

Roene-JustBetter (approved):

Jade-GG:
Usage:

```blade
<div class="container flex flex-col gap-8">
    <x-rapidez::input-field
        name="supersimpel"
        disabled
    />

    <x-rapidez::input-field
        name="simpel"
        label="simpel"
        v-model="username"
        required
    />

    <x-rapidez::input-field
        name="placeholder"
        label="placeholder"
        placeholder="text"
        v-model:value="username"
    />

    <x-rapidez::input-field
        name="advanced"
        type="email"
    >
        <x-slot:label><em>Advanced</em> shit</x-slot:label>
        <x-slot:input
            required
            placeholder="advanced"
        ></x-slot:input>
    </x-rapidez::input-field>

    <x-rapidez::input-field
        name="expert"
        type="email"
    >
        <x-slot:label class="text-xl"><em>Expert level</em> shit</x-slot:label>
        <x-slot:input
            class="pl-12"
            v-bind:required="$root.loggedIn"
            v-bind:disabled="$root.loggedIn"
            v-bind:value="$root.user.email"
        ></x-slot:input>
        <x-heroicon-o-user class="pointer-events-none absolute left-3 bottom-3 h-8 w-8 peer-enabled:opacity-0" />
        <a class="hidden absolute right-5 top-5 w-5 peer-disabled:block" href="#">
            <x-heroicon-o-logout/>
        </a>
    </x-rapidez::input-field>
</div>
```

With specific types:

```blade
<x-rapidez::input-field.select
    class="w-auto"
    name="qty"
    dusk="qty"
    label="Quantity" sr-only-label
    v-model="addToCartSlotProps.qty"
>
    <x-slot:input>
        @for ($i = $product->qty_increments; $i <= $product->qty_increments * 10; $i += $product->qty_increments)
            <option value="{{ $i }}">{{ $i }}</option>
        @endfor
    </x-slot:input>
</x-rapidez::input-field.select>


<x-rapidez::input-field.textarea
    label="{{ $option->title }} {{ $option->price_label }}"
    name="option_{{ $option->option_id }}"
    dusk="option_{{ $option->option_id }}"
    :required="$option->is_require"
    :maxlength="$option->max_characters ?: false"
    v-model="customOptions[{{ $option->option_id }}]"
/>


<x-rapidez::input-field.checkbox
    v-model="checkout.create_account"
    dusk="create_account"
    :label="__('Create an account')"
/>


<x-rapidez::input-field.radio
    v-model="checkout.shipping_method"
    v-bind:value="method.carrier_code+'_'+method.method_code"
    v-bind:dusk="'method-'+index"
>
    <x-slot:label>@{{ method.method_title }}</x-slot:label>
</x-rapidez::input-field.radio>
```

Jade-GG:
Docs update: https://github.com/rapidez/docs/pull/42

resources/views/checkout/steps/payment.blade.php
```diff
@@ -26,16 +26,18 @@ class="ml-auto w-8 h-8"
                         <a class="text-gray-700" href="#" v-on:click.prevent="toggle" v-if="agreement.mode == 'AUTO'">
                             @{{ agreement.checkbox_text }}
                         </a>
-                        <div v-else>
-                            <x-rapidez::checkbox
+                        <div v-else>he
```
royduin, line 29: Typo?

Jade-GG: You saw nothing... :)

resources/views/components/select.blade.php
```diff
@@ -1,17 +1,9 @@
...
-        'name' => $name ?? null,
-        'dusk' => $attributes->get('v-bind:dusk') ? null : $name ?? null,
-        'class' => 'w-full py-2 pl-3 pr-8 border-gray-200 rounded focus:ring-neutral focus:border-neutral sm:text-sm'
-    ]) }}>
-        {{ $slot }}
-    </select>
-</div>
+<select
+    {{ $attributes->twMerge('w-full peer rounded border border-border bg-white py-4 px-5 text-sm outline-none !ring-0 transition-all placeholder:text-inactive disabled:bg-gray-50 disabled:cursor-not-allowed font-medium') }}
+    id="{{ $name }}"
```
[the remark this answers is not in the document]
Jade-GG: Fixed in both this file and the other relevant ones!

JimmyHoenderdaal (approved):

Roene-JustBetter (approved):

jordythevulder (approved):

JimmyHoenderdaal (changes requested):
Renaming the component to "input-field" poses a significant risk of backward compatibility issues, necessitating the renaming of numerous existing input components. I recommend retaining the name "input" to avoid the need for extensive search and replace operations across projects.

Jade-GG:
> Renaming the component to "input-field" poses a significant risk of backward compatibility issues, necessitating the renaming of numerous existing input components. I recommend retaining the name "input" to avoid the need for extensive search and replace operations across projects.

We never should have originally used something named `input` for this purpose in the first place. It's a bad name for what it is in my opinion, and to keep that mistake around forever only for the sake of backwards compatibility is asinine to me.

Given that it would take a code migration regardless of whether or not we change the name, we might as well kill two birds with one stone here.

JimmyHoenderdaal:
> > Renaming the component to "input-field" poses a significant risk of backward compatibility issues, necessitating the renaming of numerous existing input components. I recommend retaining the name "input" to avoid the need for extensive search and replace operations across projects.
> 
> We never should have originally used something named `input` for this purpose in the first place. It's a bad name for what it is in my opinion, and to keep that mistake around forever only for the sake of backwards compatibility is asinine to me.
> 
> Given that it would take a code migration regardless of whether or not we change the name, we might as well kill two birds with one stone here.

I see your point, Let's try out the changes and reflect on our practices afterward. But let's not harm our avian friends. I'd rather kill 2 flies with one 👏  :)

resources/views/components/country-select.blade.php
```diff
@@ -1,9 +1,16 @@
...
-            <option v-for="country in data.countries" :value="country.two_letter_abbreviation.toUpperCase()">
-                @{{ country.full_name_locale }}
-            </option>
-        </x-rapidez::select>
+        <x-rapidez::input-field.select {{ $attributes }}>
+            <x-slot:input>
+                <option
+                    v-for="country in data.countries.toSorted((a, b) =>
+                        (a.full_name_locale ?? '').localeCompare(b.full_name_locale ?? '')
+                    )"
```
Jade-GG, line 8: `toSorted` is still a very new function and, although it has universal support now, will cause the site to break on slightly outdated browser versions that are still relevant.

royduin:
Closing as we're going to split this in a components package later based on: https://github.com/rapidez/core/pull/532

PR #361, opened 2023-10-16
Retrieve payment icons dynamically based on payment method name

Jade-GG:
Maybe this should be a package? The only issue with that would be having it hook into the payment method data nicely.

I'm using a slightly complex method of retrieving the icon because it has to match any possible payment provider as best as possible. For example, you might have:
- `creditcard` (will output `creditcard`)
- `paymentprovider_ideal` (will output `ideal`)
- `visa_creditcard` (will output `visa`, first match will count)
- `theidealprovider_visa` (will output `visa`, ignores the partial "ideal" in the payment provider code)


Jade-GG:
Dusk test seem to be timing out a lot recently 🤔

royduin:
It feels like a bit of overkill, I think it's cleaner to remove the config, route and controller. Don't think it's a good idea to route all payment icons through PHP. I've also created a new repository: https://github.com/rapidez/payment-icons which can be used to store the icons and with just a publish in the service provider. Optionally you could borrow some icons from here: https://github.com/Siteation/magento2-hyva-icons-payment. After that we could also remove the icons from the checkout theme and refactor that package to use this new repo.

royduin:
And for the Dusk tests I've created a story for the next sprint 👌🏻

Jade-GG:
> It feels like a bit of overkill, I think it's cleaner to remove the config, route and controller.

The tricky part is matching up the icons for different payment providers. I feel like it's wasteful to have an icon for, for example, `multisafepay_ideal` and `ideal`. I could probably get the icons list from the folder listings and then not make the fallback/regexes configurable, and if you really need to you can just override the vue file. As for the route & controller:

> Don't think it's a good idea to route all payment icons through PHP.

What would you suggest as a good way to have the icons be publicly available yet also make it possible to add extra icons as package user? I could make them publish directly to the public folder, but that would then cause some issues with if you want to add more icons yourself, as these would be gitignored.

With this setup---given the storage is linked---it will only go through the php route once to add a new icon to the storage, which doesn't seem too problematic to me and reserves the option to add more icons through the `payment-icons` folder. But I understand it does add complexity 😅

royduin:
In the package just add the icons once, for example: `ideal.svg`, for all variations for providers you could add relative symlinks; `multisafepay_ideal.svg` links to `ideal.svg`

For publishing just publish it to the public directory just like Laravel Horizon, etc does. If someone wants different icons in their project they can just add/change it there and commit it. If someone doesn't want the payment icons in the project they can gitignore it and add the payment icons publish in the postupdate/deploy script.

The simplest and still flexible solution I think.

Jade-GG:
> In the package just add the icons once, for example: `ideal.svg`, for all variations for providers you could add relative symlinks; `multisafepay_ideal.svg` links to `ideal.svg`

As far as I'm aware symlinks aren't universally supported, for example I'm pretty sure you could not do this at all on an AWS instance. I also can't imagine maintaining a big folder with >50% symlinks, that doesn't feel like a good solution to me.

Naively matching up the existing icons list with whatever payment code you get seems simple enough and plenty flexible to me.

royduin:
What about removing the controller; you just have to publish the icons to your project and listing the icons automatically instead of specifying it in the config with `Storage::disk(...)->files('payment-icons')`?

@indykoning maybe other ideas?

indykoning:
I would definitely publish the images to the public folder in the same way horizon does as it is a massive difference in speed.
i'm pretty sure the bugfix in rapidez/image-resizer that allowed it to skip the controller made a difference of 100+ ms vs <10ms

And of course it reduces server load since php and the framework does not have to be booted for each icon requested.

You're right symlinks aren't supported by the disk class for that reason as only the local driver _could_ support it.
but if you try to put the public folder on another disk than local, like s3 you will have much bigger problems to solve 😉 

personally i am a fan of rapidez/core publishing the icons base magento would use.
And letting other payment providers publish their own icons to the folder, sure if you add multiple payment providers supporting the same methods you might end up with the same icon under a different name.

But i think most people prefer to lose a couple more bytes of disk space considering the performance benefit

royduin:
Good one! I removed the `rapidez/payment-icons` repo. Looking forward to this approach.

PR #362, opened 2023-10-18
Price component

royduin:


resources/js/components/Product/AddToCart.vue
```diff
@@ -304,14 +272,5 @@ export default {
...
     },
-
-    watch: {
-        customOptions: {
-            handler() {
-                this.calculatePrices()
-            },
-            deep: true,
-        },
-    },
```
Jade-GG, line 315: Todo: Will probably need to re-implement such a watcher to have the options update in the price component

resources/js/components/Product/Price.vue
```diff
@@ -0,0 +1,124 @@
...
+
+        getTaxPercent(product) {
+            let country_id = window.config.tax.defaults.country_id
+            let region_id = window.config.tax.defaults.region_id
+            let postcode = window.config.tax.defaults.postcode
+
+            if (['shipping', 'billing'].includes(window.config?.tax?.calculation.based_on)) {
+                country_id = window.app?.$data?.checkout?.[window.config?.tax?.calculation.based_on + '_address']?.country_id || country_id
+                region_id = window.app?.$data?.checkout?.[window.config?.tax?.calculation.based_on + '_address']?.region_id || region_id
+                postcode = window.app?.$data?.checkout?.[window.config?.tax?.calculation.based_on + '_address']?.postcode || postcode
```
Jade-GG, line 53:
```js
let basedOn = window.app?.$data?.checkout?.[window.config?.tax?.calculation.based_on + '_address']
```

And then use that instead of repeating it 3 times, would probably be much cleaner here.

src/Models/Product.php
```diff
@@ -116,6 +108,27 @@ public function rewrites(): HasMany
...
+    {
+        return $this->hasMany(
+            config('rapidez.models.product_tier_price'),
+            'entity_id'
+        )
+            ->whereIn('website_id', [0, config('rapidez.website')])
+            ->where(fn ($query) => $query
+            ->where('all_groups', 1)
+            ->orWhere('customer_group_id', auth('magento-customer')->user()?->group_id ?? 0)
+            );
```
Jade-GG, line 121: Formatting?

Jade-GG (changes requested):

resources/js/components/Product/Price.vue
```diff
@@ -0,0 +1,139 @@
...
+
+    computed: {
+        specialPrice() {
+            JSON.stringify(this.options) // Hack: make vue recognize reactivity within the options object
+            return this.calculatePrice({ ...this.options, ...{ special_price: true } })
+        },
+
+        price() {
+            return this.calculatePrice(this.options)
+        },
```
Jade-GG, line 113: This needs to be thought out again, Magento doesn't work with prices this way.

Prices are calculated as follows: `Final Price=Min(Regular(Base) Price, Group(Tier) Price, Special Price, Catalog Price Rule) + Sum(Min Price per each required custom option)`

Note that prices are clamped to a minimum of 0 before and after the custom options are added. Also, the price gets rounded before getting multiplied by the qty.

PR #365, opened 2023-10-20
Avoid error in console

Jade-GG:


royduin:
In which scenario does this occur?

Jade-GG:
This happens essentially always. Because the order data comes from the request done on mount, it'll take a bit before that data is available but the checkout success component will try to render the items regardless.

It won't kill the page, but it'll throw a vue warning.

PR #369, opened 2023-10-30
Add config flag to let the indexer index invisible products

Jade-GG:


config/rapidez/system.php
```diff
@@ -22,6 +22,9 @@
     // With this token you can run commands from an url.
     'admin_token' => env('RAPIDEZ_TOKEN', env('APP_KEY')),
 
+    // Whether or not to index products with visibility 1 into ElasticSearch
+    'index_invisible' => false,
```
indykoning, line 26: If we want the setting to differ for other store views that's not possible.
I suggest wrapping the config getter in [value()](https://laravel.com/docs/10.x/helpers#method-value)

This way a closure could be used as well, allowing for customisation of the return value

royduin:
Maybe do the same as we did with the Statamic package? https://github.com/rapidez/statamic/blob/master/config/rapidez-statamic.php#L28-L33

Maybe create a new `config/rapidez/indexer.php` with this config 
+ `expose_stock` from `config/rapidez/frontend.php`
+ `searchable` and `additional_filters` from `config/rapidez/system.php`

config/rapidez/system.php
royduin: I don't think this setting will be different per store

royduin:
Just for reference; this was added in: https://github.com/rapidez/core/pull/343

config/rapidez/system.php
Jade-GG: Might still be worth having the option though, I can imagine some crazy multistore setup could make use of it

(but then maybe we could think about making everything like that and that might be very overkill 🤔 )

Jade-GG:
>     * `expose_stock` from `config/rapidez/frontend.php`

I don't know if this one would make much sense to be in the indexer config as it's also used in the frontend on PDPs, and splitting it is pretty weird because there'd have to be some ugly workarounds for when one is set to true and the other to false.

But I'll move the others over to a new config :)

indykoning:
I agree, when i'm looking through the code it seems more like an indexer thing.
If we find a term to encompass both while being broad enough that would be great, but if not i think system is the better place for it as well

config/rapidez/indexer.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+return [
+    // Which product visibilities should be indexed?
+    // VISIBILITY_NOT_VISIBLE    = 1
+    // VISIBILITY_IN_CATALOG     = 2
+    // VISIBILITY_IN_SEARCH      = 3
+    // VISIBILITY_BOTH           = 4
+    'index_visibility' => [2, 3, 4],
```
royduin, line 9:
```suggestion
    'visibility' => [2, 3, 4],
```

src/Commands/IndexProductsCommand.php
```diff
@@ -56,10 +56,11 @@ public function handle()
                     ->pluck('name', 'entity_id');
 
                 $showOutOfStock = (bool) Rapidez::config('cataloginventory/options/show_out_of_stock', 0);
+                $indexVisibility = config('rapidez.indexer.index_visibility');
```
royduin, line 59:
```suggestion
                $indexVisibility = config('rapidez.indexer.visibility');
```

PR #370, opened 2023-10-31
Use data_get to allow both stdClass & arrays to be used

Jade-GG:
Cute lil helper function that laravel has, also allows dot notation to be used for deeper nesting if necessary 😇


royduin:
Use case? 😇

PR #372, opened 2023-11-03
GraphQL cart

royduin:
So for now just the cart with a lot of todo's. The main difference is that we can get all cart information from everything that deals with the cart. Like; adding products, changing quantities, coupons, etc. so we don't have to do 2 calls anymore: action + cart refresh.

But we also have some challenges like:
- ~~Where do we store the cart? Currently I'm using the GraphQL Vue component cache but that's not reactive so when something changes we're firing a callback which can be picked up by the new `rerunEvent` prop.~~ => Global store
- ~~How are we going to get the `min_sale_qty` and `qty_increments` values as default Magento GraphQL doesn't expose it?~~ => Added using an optional Compadre module

And some questions:
- ~~Are we also going to migrate the add to cart calls to GraphQL? As we can get all cart information directly from that call. Maybe a separated story.~~ => Yes
- ~~Can we remove all the quote related models? Are they used somewhere for something else?~~ => They're still in use

Things left to do:
- ~~Cart store; fix the current "reactivity" or move to `window.app.cart`~~ => Fixed; global store
- ~~Test everything with a logged in user~~
- ~~Implement the minicart~~ => Implemented
- ~~Test/implement expired carts~~
- ~~Test/implement link user to cart~~
- ~~Migrate add to cart to GraphQL?~~ => Yes, done
- Frontend like cart items in a table
- ~~Implement `min_sale_qty` and `qty_increments` (Update the docs to install rapidez/magento2-compadre)~~

I think it's a good idea to finish the cart first before we start with the checkout.

https://github.com/rapidez/account/pull/43
https://github.com/rapidez/login-as-customer/pull/7
https://github.com/rapidez/amasty-automatic-related-products/pull/17
https://github.com/rapidez/docs/pull/39
https://github.com/rapidez/amasty-custom-form/pull/7


src/RapidezServiceProvider.php
```diff
@@ -12,6 +12,8 @@
 use Illuminate\Support\Facades\View;
 use Illuminate\Support\Facades\Vite;
 use Illuminate\Support\ServiceProvider;
+// TODO: Anyone a better name for this?
+use Illuminate\View\View as ViewView;
```
Jade-GG, line 16: FacadeView for the first one and then just View for this one? 😇

royduin: Yeah but are we going to do that for all facades? Facades are used more compared to macroable classes

indykoning:
> Where do we store the cart? Currently I'm using the GraphQL Vue component cache but that's not reactive so when something changes we're firing a callback which can be picked up by the new rerunEvent prop.
i think we want to keep this in de useCart Store, it supports a getter and a setter so when we have new info that was gathered outside of the cart storage (e.g. addtocart) we push it into the storage and everything gets updated.

The reason why we want to keep the useCart store, and update the refresh function to use graphql is that we cannot only get the cart info after an action. We need to do it on e.g. page load as well to make sure we have correct information.

The useCart store should still be the place responsible for managing data (This is from vuex docs https://reintech.io/blog/working-with-advanced-state-management-in-vuejs#:~:text=This%20store%20would%20be%20responsible%20for%20managing%20the%20data%2C%20ensuring%20it%20is%20always%20up%2Dto%2Ddate%2C%20and%20providing%20access%20to%20the%20data%20when%20needed.)

> Are we also going to migrate the add to cart calls to GraphQL? As we can get all cart information directly from that call. Maybe a separated story.
I think we should want that as it could greatly improve the speed of the cart updating, and solves the problem of cart refreshes clashing because they were executed close together.
The last request can update the store with the latest info.

src/RapidezServiceProvider.php
Jade-GG: Could always rename it to `MacroView` if it's only used as a macroable

src/Models/Quote.php
```diff
@@ -9,6 +9,7 @@
 use Rapidez\Core\Models\Scopes\IsActiveScope;
 use TorMorten\Eventy\Facades\Eventy;
 
+// TODO: Do we still need this model?
```
indykoning, line 12: This model is still being used for the order controller, to get the order by quote hash

src/RapidezServiceProvider.php
```diff
@@ -243,6 +245,10 @@ protected function bootMacros(): self
             })->sortByDesc('store_id')->first();
         });
 
+        ViewView::macro('renderOneliner', function () {
+            return trim(preg_replace('/\s+/', ' ', $this->render()));
+        });
+
```
indykoning, line 251: I think this might not be necessary if we use https://laravel.com/docs/10.x/strings#method-str-squish.
`Str::squish(view('rapidez::cart.queries.cart')->render())`
Depends on if we want to call it on the view or if using it more verbose is okay

royduin:
The GraphQL component callback params changed to be inline with the GraphQL Mutation component, maybe we should split this from this PR to have it in the changelog / upgrade notes as we don't know where this is used in projects. Within packages this is used here:

- https://github.com/rapidez/amasty-custom-form/pull/7
- https://github.com/rapidez/account/pull/43

resources/js/stores/useCart.js
```diff
@@ -55,19 +44,27 @@ export const clearAddresses = async function () {
 
 export const cart = computed({
     get() {
-        if (!cartStorage.value?.entity_id && mask.value) {
-            refresh()
+        if (!cartStorage.value?.id && mask.value) {
+            // TODO: Check as we don't want this all the time.
```
indykoning, line 48: I think we do want this all the time, since this only gets executed when we do not have a cart but do have a mask.
Which means either we **should** have a cart or we **shouldn't** have a mask.

These cases get sorted out by the refresh function.
 - If we have no cart but we have a valid mask, we retrieve the cart
 - If we have no cart and we have an invalid mask, the mask gets removed with the notification that the cart has expired.
 - If we have a cart and a mask this will not get triggered

PR #374, opened 2023-11-06
Use publish for payment icons

Jade-GG:
Simpler solution for https://github.com/rapidez/core/pull/361

Do we really want to have the specific icons re-copy-pasted in every payment provider package? If one has to change, say, the `ideal` icon to `wero`, it would be annoying to have to update every repo.

Either way, this works fine :)


royduin:
These icons are not default Magento and should go to the payment provider packages. In the core we only want the payment icons for the default payment options.

Jade-GG:
> These icons are not default Magento and should go to the payment provider packages.

I get that, but do we really want to have to maintain that list in every package? We'll have several "sources of truth" so to speak with the same icons and it might get very annoying.

indykoning:
They may have the same icons, but they may also have different branded icons with different names.
When a payment provider adds/removes/changes a payment method we should change the icon in the respective package instead of requiring the user to update rapidez/core to update the icons.

Of course we might create an integration with a different payment provider as well, where the package should be responsible for the payment icons as it is impossible for us to know all icons used by all providers

PR #376, opened 2023-11-07
Add magentoGraphQL axios instance

jbclaudio:
This adds an instance of Axios that should be used for GraphQL requests. This allows for example the following to be used to send GraphQL requests to Magento:
```
magentoGraphQL.post('graphql', {
query:
`mutation StartTransaction(
    $order_id: String
    $return_url: String
) {
  paynlStartTransaction (
    order_id: $order_id
    return_url: $return_url
  ) {
    redirectUrl
  }
}`,
variables: {
    order_id: data.order.id,
    return_url: window.url('/paynl/finish')
}}).then(response => {
    window.location.replace(response.data.data.paynlStartTransaction.redirectUrl)
})
```

This now always includes a Store header by default and possibly an Authorization Bearer token if the user is logged in.

### DIFF
--- a/resources/js/axios.js
+++ b/resources/js/axios.js
@@ -10,6 +10,13 @@ window.magentoUser = axios.create()
 window.magentoUser.defaults.baseURL = config.magento_url + '/rest/' + config.store_code + '/V1/'
 window.magentoUser.defaults.headers.common['Authorization'] = `Bearer ${localStorage.token}`
 
+window.magentoGraphQL = axios.create()
+window.magentoGraphQL.defaults.baseURL = config.magento_url
+if (localStorage.token) {
+    window.magentoGraphQL.defaults.headers.common['Authorization'] = `Bearer ${localStorage.token}`
+}
+window.magentoGraphQL.defaults.headers.common['Store'] = config.store_code
+
 // It's not possible to set global interceptors like headers
 // or the base url can; so we set them for all instances.
 // See: https://github.com/axios/axios/issues/510


royduin:
As the endpoint for GraphQL is always the same maybe we can skip the location part, just pass the data and change the base url to include `/graphql`? The `Graphql.vue` and `GraphqlMutation.vue` components can use this already?

As all projects are migration to 1.x do we still need this in 0.x?

resources/js/axios.js
```diff
@@ -10,6 +10,13 @@ window.magentoUser = axios.create()
 window.magentoUser.defaults.baseURL = config.magento_url + '/rest/' + config.store_code + '/V1/'
 window.magentoUser.defaults.headers.common['Authorization'] = `Bearer ${localStorage.token}`
 
+window.magentoGraphQL = axios.create()
+window.magentoGraphQL.defaults.baseURL = config.magento_url
+if (localStorage.token) {
+    window.magentoGraphQL.defaults.headers.common['Authorization'] = `Bearer ${localStorage.token}`
```
indykoning, line 16: It's better to use token.value from the userStore here, as it allows much more flexibility where we store the token in the future.

indykoning (commented):
This will still require us to do a lot ourselves when we want to run a Graphql query. Maybe it's better to create an own import so we can add additional functionality to running the query.
Like handling graphql error responses which coincidentally i've had to add here https://github.com/rapidez/core/pull/372/commits/beb26d30184237c92d419e1a7905d0e9c9ff459c#diff-078e7249289d8b23d2867fbd0c68c26c25a9b6798a800dfe6122dfc578fc4a1aR82

Here's a quick setup of what i mean with a new file
```js
import axios from 'axios'
import { token } from './stores/useUser'


graphqlAxios = axios.create()
graphqlAxios.defaults.baseURL = config.magento_url + '/graphql'

function initAuthorization() {
    if (token.value) {
        graphqlAxios.defaults.headers.common['Authorization'] = `Bearer ${token.value}`
    } else {
        graphqlAxios.defaults.headers.common['Authorization'] = undefined;
    }
}

function initStore() {
    if (window.config.store_code) {
        graphqlAxios.defaults.headers.common['Store'] = window.config.store_code
    } else {
        graphqlAxios.defaults.headers.common['Store'] = undefined;
    }
}

export async function getInstance() {
    initAuthorization();
    initStore();
    return graphqlAxios;
}

export async function run(query, variables = {}, options = {}) {
    instance = await getInstance();

    return await instance.post(
        '/graphql',
        {
            query: query,
            variables: variables,
        },
        options,
    ).then((response) => {
        if (response.data.errors) {
            throw new axios.AxiosError('Graphql Errors', null, response.config, response.request, response)
        }

        return response;
    })
}
```

This way we can just execute 
```
run(
  'query cart ($cart_id: String!) { cart (cart_id: $cart_id) { etc.. }}', 
  {'cart_id': 'sjdghsakdhsakjdjksah'}
)
.then((response) => {
  // logic with the response
})
.catch((error) => {
  // Any http or graphql error occured
})
```

indykoning:
> The Graphql.vue and GraphqlMutation.vue components can use this already?
I also think it is important to run all graphql queries in the frontend using this, it makes it much more consistent and easier to debug.

PR #379, opened 2023-11-10
Support MSP way of saving method_title

indykoning:
Multisafepay saves the method title under `raw_details_info`
If we cannot get the method title from the default place, we'll try and get it from there.


royduin:
Do we also need this in the checkout theme?

indykoning:
Yes, we do. Good catch https://github.com/rapidez/checkout-theme/pull/52

PR #382, opened 2023-11-16
Respect the position per category

indykoning:
Reason i'm using the function_score instead of sorting on the field directly is so relevancy will still be used when no positions are given or positions are the same.

### DIFF
--- a/src/Commands/IndexProductsCommand.php
+++ b/src/Commands/IndexProductsCommand.php
@@ -88,7 +88,7 @@ class IndexProductsCommand extends ElasticsearchIndexCommand
                         $data['positions'] = $product->categoryProducts
                             ->pluck('position', 'category_id')
                             // Turn all positions positive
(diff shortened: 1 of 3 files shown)


src/Commands/IndexProductsCommand.php
```diff
@@ -78,6 +85,11 @@ public function handle()
 
                         $data = $this->withCategories($data, $categories);
 
+                        $data['positions'] = $product->categoryProducts
+                            ->pluck('position', 'category_id')
+                            // Turn all positions positive
+                            ->mapWithKeys(fn ($position, $category_id) => [$category_id => (($position * -1) + $maxPositions[$category_id])]);
```
Jade-GG, line 91: `$maxPositions[$category_id] - $position` seems more logical to me here

Jade-GG (dismissed):

PR #383, opened 2023-11-21
Clear cart storage

jbclaudio:
When an order is placed as a logged-in user, the shopping cart is not emptied. This fix ensures that it does get emptied after placing the order.

### DIFF
--- a/resources/js/stores/useCart.js
+++ b/resources/js/stores/useCart.js
@@ -31,6 +31,7 @@ export const refresh = async function () {
     } catch (error) {
         if (error.response.status == 404) {
             mask.value = null
+            cartStorage.value = {}
             return false
         }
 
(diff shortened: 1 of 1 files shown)


indykoning (dismissed):

resources/js/stores/useCart.js
```diff
@@ -45,6 +46,7 @@ export const clear = async function () {
     await clearAddresses()
     await clearMask()
     await refresh()
+    cartStorage.value = {}
```
royduin, line 49: Are we sure we needs this? As `refresh()` is called, if the mask/token isn't there it will be cleared; otherwise a request will be made and the storage will be updated.

indykoning: In this case the token will still be there when the customer is logged in.
So it will trigger the 404 and empty the cart after the request finishes, this 404 is intended as we do not know wether the user has a cart

royduin, line 49:
```suggestion
```

PR #385, opened 2023-11-24
Add check if already fetched

indykoning:
This helps prevent repeating request when the API returns an empty array


royduin:
FYI @Jade-GG

PR #390, opened 2023-12-04
Add config option for category positions

Jade-GG:
Ran into an issue with a certain project where we ended up having more than 1000 individual mappings in elasticsearch, because this creates an individual mapping for each category---almost two thirds of the mapping table was just category positions.

### DIFF
--- a/config/rapidez/indexer.php
+++ b/config/rapidez/indexer.php
@@ -19,4 +19,9 @@ return [
     'additional_filters' => [
         // eav_attribute attribute_code. e.g. brand
     ],
+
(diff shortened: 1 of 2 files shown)


indykoning (approved):

royduin:
This isn't really a fix right?

royduin:
So what's next? Is there a decent fix for this?

Jade-GG:
> So what's next? Is there a decent fix for this?

We could always just tell people in the docs to flatten the positions data in the case that it takes up too much real estate in elasticsearch... :)

PR #391, opened 2023-12-05
Changed cart into a table

JimmyHoenderdaal:


Roene-JustBetter (approved):

resources/views/cart/item.blade.php
```diff
@@ -1,25 +1,44 @@
...
+                        @{{ option.option_label }}: @{{ option.value_label }}
+                    </div>
+                    <div v-for="option in item.customizable_options">
+                        @{{ option.label }}: @{{ option.values[0].label || option.values[0].value }}
+                    </div>
+                    @include('rapidez::cart.item.remove')
+                </div>
+            </td>
+            <td class="justify-center text-right max-md:flex max-md:w-full">
+                <div class="inline-flex flex-1 items-center justify-between gap-2 md:gap-5">
```
jordythevulder, line 36: inline-flex and a flex-1?

jordythevulder (changes requested):

indykoning:
> I assume the dusk tests fail because of the getcart thing that prevents it from performing yarn dev, is this right? @royduin @indykoning?

Yes, we'll fix the tests in the graphql-checkout branch 👍

indykoning (approved):

PR #397, opened 2023-12-11
Update Lazy component

Jade-GG:
This is a slightly breaking change, because it changes the `intersected` slot prop to actually do what it says, and adds `loaded` to cover that original use case.

I don't think there's a clean way to make this functionality without also breaking backwards compatibility? The name `intersected` kinda assumes that you've actually intersected this component.


royduin:
Why do we needs this change?

Jade-GG:
> Why do we needs this change?

Having access to just the intersection can be useful at times, this change is specifically because it was really useful for what I was working on. I also have a slight vendetta against variables that don't do what their name says they should, like this one did :)

The `idleTimeout` prop is just one I happened to notice while I made this change so I added it along with it.

indykoning:
I agree that this is much better naming, intersected should not be set to true if the element.

Do note that the observer will stop observing once it has been triggered, so it might be better to have a component to handle intersection instead of abusing the lazy component for it

royduin:
Status?

Jade-GG:
> Status?

I guess the question is, is it worth breaking backwards compatibility a little bit to fix the not-so-consequential variable name being wrong?

indykoning:
Do we still need the changes made in here?
The lazy component is made to lazy load vue components, not to act like an intersection observer.
If we don't need that we can close this

PR #400, opened 2024-01-04
GraphQL components store prop

royduin:
So it's possible to override this. Useful for example with the reviews; they are store specific. In case you want to show product reviews from another store.


royduin:
Merged but still needs a review @indykoning, most likely we need to reimplement this on the `graphql-checkout` branch as we've some merge conflicts there now: https://github.com/rapidez/core/pull/372

PR #402, opened 2024-01-11
Fall back to other default address if only one is set

Jade-GG:
This caused issues in the checkout when only default billing or only default shipping were set.


indykoning:
Does this happen in default Rapidez as well? This fault has occurred in the checkout theme but has been fixed there.
https://github.com/rapidez/checkout-theme/pull/64

Jade-GG:
> Does this happen in default Rapidez as well? This fault has occurred in the checkout theme but has been fixed there. [rapidez/checkout-theme#64](https://github.com/rapidez/checkout-theme/pull/64)

Should do, yes, though I'm surprised we haven't seen it happen before.

PR #403, opened 2024-01-11
Always set step in checkout success to success step

Jade-GG:
This is relevant for when external payment providers are used---the checkout step value is lost, causing it to say "step 1 out of X".

We can reasonably assume the success step is always step 4 as it's forced this way with the rest API stuff, but I've overengineered it anyway to not have it be hardcoded.

### DIFF
--- a/resources/js/components/Checkout/CheckoutSuccess.vue
+++ b/resources/js/components/Checkout/CheckoutSuccess.vue
@@ -29,6 +29,11 @@ export default {
         this.token ??= localStorage.token
         this.mask ??= localStorage.mask
 
+        let successStep = this.$root.config.checkout_steps[config.store_code]?.indexOf('success')
+        if (successStep > 0) {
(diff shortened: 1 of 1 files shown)


indykoning:
Actually it's good it's not hardcoded to 4 as we've made it easy to add/remove steps in https://github.com/rapidez/core/blob/master/config/rapidez/frontend.php#L22
We could also reasonably assume the last step is the success step but this works just as well.

We have a merge conflict and failing tests, could you take a look? 🙂

Jade-GG:
> we've made it easy to add/remove steps

You say this but there are currently plenty of reasons why this will break the checkout :)

Should be fine after the GraphQL update though, which is part of the reason why I made sure to do it this way.

Jade-GG:
Looks like dusk test are failing since [this commit](https://github.com/rapidez/core/commit/d1764012714a20073bbb0a6282b3c6d8251cb283).

PR #404, opened 2024-01-18
Add product parent

Jade-GG:


royduin:
[Has One Through](https://laravel.com/docs/master/eloquent-relationships#has-one-through)?

Jade-GG:
> [Has One Through](https://laravel.com/docs/master/eloquent-relationships#has-one-through)?

It technically doesn't quite fit the bill, but in a slightly hacky way it seems it does work if used like 

```
return $this->hasOneThrough(config('rapidez.models.product'), config('rapidez.models.product_link'), 'product_id', 'entity_id', 'entity_id', 'parent_id')->withoutGlobalScopes();
```

So I guess that's a good replacement.

Jade-GG:
It's very funny(read: frustrating) that the dusk tests are failing here, given that this PR changes literally nothing until you call `->parent` on a product, which is a new functionality that's not used anywhere.

src/Models/Product.php
```diff
@@ -134,6 +136,16 @@ public function rewrites(): HasMany
...
     }
 
+    public function parent(): HasOneThrough
+    {
+        return $this->hasOneThrough(
+            config('rapidez.models.product'),
+            config('rapidez.models.product_link'),
+            'product_id', 'entity_id',
+            'entity_id', 'parent_id'
+        )->withoutGlobalScopes();
```
royduin, line 146: Maybe if we follow some more key conventions and some relations on the link model we can remove some of those keys.

Jade-GG: The third key is technically unnecessary in the current state, but if we turn the ProductLink model into a pivot with `asPivot` and manually set $foreignKey to `entity_id` in the constructor we can also remove the second key, turning it into:

```
return $this->hasOneThrough(
    config('rapidez.models.product'),
    config('rapidez.models.product_link),
    firstKey: 'product_id',
    secondLocalKey: 'parent_id',
)->withoutGlobalScopes();
```

I've tried some other things but can't really find anything else possibly worthwhile. The Laravel documentation on this stuff seems honestly kind of lacking and hard to parse.

indykoning (approved):
Should be good for now, i'm seeing some failing tests. Could you pull in Master? 🙂

indykoning (approved):

PR #406, opened 2024-01-19
Add events to handle order placing independently

indykoning:
These changes were required to support https://github.com/rapidez/paynl/pull/14


royduin:
Backwards compatible? Do we need to document the new events?

indykoning:
If any of these are not used the old functionality will still work as expected.
The only thing that might be considered backwards incompatible, is the fact that `order.id` can be `null` in `checkout-payment-saved`

I'll create a PR to update the docs

PR #407, opened 2024-01-22
Improve flexibility of additional search queries in autocomplete

Jade-GG:


royduin:
Could you provide some more information? The current query is taken from ReactiveSearch and how should the newly added parts be used? Maybe write some docs for it?

resources/js/components/Search/Autocomplete.vue
```diff
@@ -12,43 +12,75 @@ export default {
...
 
-    methods: {
-        searchAdditionals(query) {
-            if (!this.additionals) {
+    mounted() {
+        this.$nextTick(() => this.$emit('mounted'))
+        let self = this
+
+        // Define function here to gain access to the debounce prop
+        this.searchAdditionals = useDebounceFn(function (query) {
```
indykoning, line 42: Are we sure we want to use Debounce? Or do we want to use throttle?
What does reactivesearch do?

resources/js/components/Search/Autocomplete.vue
```diff
@@ -12,43 +12,75 @@ export default {
...
+            if (!self.additionals) {
                 return
             }
 
-            this.results = { count: 0 }
+            // Initialize with empty data to preserve additionals order
+            self.results = Object.fromEntries(Object.keys(self.additionals).map((e) => [e, []]))
```
indykoning, line 48: What is `e` in this context? Maybe update the variable name to reflect that

resources/js/components/Search/Autocomplete.vue
Jade-GG: ReactiveSearch (at least the DataSearch component) also does a debounce, this mirrors the functionality so that the search queries are always executed at the same time.

Jade-GG:
> Could you provide some more information? The current query is taken from ReactiveSearch and how should the newly added parts be used? Maybe write some docs for it?

Docs is probably the best approach, there is one example given in the `autocomplete.blade.php` on how it should be used but it could definitely have more documentation.

Basically, this allows you to add any extra index query to the autocomplete by adding it to the `additionals` prop like before, but now it's also structured in a more modular and extendable way.

There's also a highlight function now which---and I'm amazed that we've not done it this way before in implementations of the autocomplete---uses the standard highlighting feature from elasticsearch, with a little extra workaround to fit the ReactiveSearch-data-shaped-peg into a round hole.

resources/views/layouts/partials/header/autocomplete.blade.php
```diff
@@ -6,16 +6,17 @@
...
-    v-if="!$root.loadAutocomplete">
+    v-if="!$root.loadAutocomplete"
+>
 
 <autocomplete
     v-if="$root.loadAutocomplete"
     v-on:mounted="() => window.document.getElementById('autocomplete-input').focus()"
-    v-cloak
-    :additionals="{ categories: ['name^3', 'meta_description^1'] }"
+    v-bind:additionals="{ categories: ['name^3', 'meta_description^1'] }"
```
Jade-GG, line 15: Do we maybe want this list to come from a config file by default?

indykoning (approved):

PR #413, opened 2024-01-30
Seperate check whether flat tables are active, and if they exist

indykoning:


src/Listeners/MagentoSettingsHealthcheck.php
```diff
@@ -63,7 +73,7 @@ public function handle()
         }
 
         if (! $response['healthy']) {
-            $response['messages'][] = ['type' => 'info', 'value' => __('The of Magento settings have been checked, there were some errors; fix them before you continue. See: https://rapidez.io/docs/0.x/installation#flat-tables')];
+            $response['messages'][] = ['type' => 'info', 'value' => __('The Magento settings have been checked, there were some errors; fix them before you continue. See: https://rapidez.io/docs/0.x/installation#flat-tables')];
```
Jade-GG, line 76: While we're here could you change this url to the [1.x version](https://docs.rapidez.io/1.x/installation.html#flat-tables)?

indykoning: You're right! I've updated the url to the correct docs

Jade-GG (approved):

PR #416, opened 2024-02-01
Use lowercase & asciifolding filters in default analyzer & in synonym analyzer

Jade-GG:
- `lowercase` filter makes the searching case insensitive
- `asciifolding` allows the search to ignore accents on letters and other weird unicode stuff


Jade-GG:
I've also added the `keyword` field on the mappings that get synonym-ized, because this caused the alphabetical sorting to cease metabolic function.

indykoning (approved):

PR #417, opened 2024-02-07
Use Arr::wrap()

Jade-GG:


indykoning:
this changes the behaviour from what has been described here https://github.com/rapidez/core/pull/411#issue-2100576552
Is there a problem we're trying to solve?

Jade-GG:
> this changes the behaviour from what has been described here [#411 (comment)](https://github.com/rapidez/core/pull/411#issue-2100576552) Is there a problem we're trying to solve?

Ah fair, just seemed like an obvious code cleanup :)

PR #418, opened 2024-02-08
Allow store-specific searches & ignore failed requests

Jade-GG:


indykoning (approved):

royduin:
Conflicts 😇

Jade-GG:
> Conflicts 😇

Not anymore 😇

resources/js/components/Search/Autocomplete.vue
```diff
@@ -87,11 +92,12 @@ export default {
                     method: 'POST',
                     headers: { 'Content-Type': 'application/json', Authorization: auth },
                     data: JSON.stringify(esQuery),
+                    validateStatus: false,
                 }).then(async (response) => {
                     const responseData = await response.json()
 
                     self.results[name] = responseData?.hits ?? []
-                    self.results.count += self.results[name]?.hits?.length ?? 0
+                    self.results.count += self.results[name]?.hits?.length ??
```
royduin, line 100: No fallback?

resources/js/components/Search/Autocomplete.vue
```diff
@@ -87,6 +92,7 @@ export default {
                     method: 'POST',
                     headers: { 'Content-Type': 'application/json', Authorization: auth },
                     data: JSON.stringify(esQuery),
+                    validateStatus: false,
```
royduin, line 95: Isn't that an Axios thing?

royduin:
Any updates?

royduin:
Any info?

Jade-GG:
When working with a multistore setup, sometimes some of the stores only require some of the indexes to be searched. For example, if you have a blog on only one of your three multistore sites, you wouldn't want that blog index to also be searched on the other two. Thus, with this change we can choose which indexes get searched per store.

royduin:
And where do we configure that? Do we provide an example and/or some info in the docs?

royduin:
And the docs? Currently there isn't mentioned anything about indexing "additionals"

royduin:
Follow-up with internal reference: RAP-1054

PR #419, opened 2024-02-09
Support catalog urls

indykoning:
Magento has these urls built in by default, and these are always available in Magento.
When no url rewrite has been made for the specified product or category Magento will automatically publish the `catalog/<x>/view/id/<id>` url to the sitemap. Causing 404s to be found on the sitemap prevously


indykoning:
You're right! Default Magento always sends the canonical url to the url_rewrite if it has one

PR #420, opened 2024-02-12
Reduce the usage of hardcoded step numbers

indykoning:
This new way of getting the next step number allows us to place the login step on the same step as another simply by extending the Login.vue and overriding the computed nextStep function

### DIFF
--- a/resources/js/components/Checkout/Login.vue
+++ b/resources/js/components/Checkout/Login.vue
@@ -67,7 +67,7 @@ export default {
                 .then((response) => {
                     if ((this.emailAvailable = response.data)) {
                         this.$root.guestEmail = this.email
-                        this.$root.checkout.step = 2
+                        this.$root.checkout.step = this.nextStep
                     } else {
                         this.$nextTick(function () {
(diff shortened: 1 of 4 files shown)


resources/js/components/Checkout/Login.vue
```diff
@@ -86,11 +86,19 @@ export default {
...
+                this.$root.checkout.step = this.nextStep
             } else if (this.redirect) {
                 Turbo.visit(window.url(this.redirect))
             }
         },
     },
+    computed: {
+        loginStep: function () {
+            return config.checkout_steps[config.store_code]?.indexOf('login')
+        },
```
Jade-GG, line 98: We're using this `indexOf` construction quite a bit, maybe it's worth making a function out of it?

Could make the nextStep function become something like:

```js
return checkoutStep('login') + 1
```

And also simplify the v-ifs in the blade file below.

indykoning: You're right, that's a good idea to create a simplified function. Especially for template work.

PR #421, opened 2024-02-12
Check to make sure currentShippingMethod exists

Jade-GG:


indykoning:
Will be fixed in another way in #432

PR #422, opened 2024-02-15
Add weee_tax support in cart

Jade-GG:
There's no cleaner way to retrieve the weee_tax directly from the cart itself, unfortunately.

As we can't be sure whether or not it ends up getting included in the subtotal (because that's a magento setting), the only way to get the correct value is to sum all the individual values from the items2 array.

I don't know how this will work with the GraphQL cart.


indykoning:
> I don't know how this will work with the GraphQL cart.

closely the same way
https://github.com/magento/magento2/blob/2.4-develop/app/code/Magento/WeeeGraphQl/etc/schema.graphqls

weee will add the `fixed_product_taxes` field to the product prices field we already use, with the label and value

royduin:
Status?

royduin:
So this is for 1.x? Do we need it there? As 2.x is refactored to GraphQL?

resources/views/cart/overview.blade.php
```diff
@@ -53,6 +53,8 @@
                             <dd v-if="cart.tax > 0" class="text-right">@{{ cart.tax | price }}</dd>
                             <dd v-if="cart.shipping_amount > 0">@lang('Shipping')<br><small>@{{ cart.shipping_description }}</small></dd>
                             <dd v-if="cart.shipping_amount > 0" class="text-right">@{{ cart.shipping_amount | price }}</dd>
+                            <dd v-if="weeeTax">@lang('Disposal fee')</dd>
```
royduin, line 56: Weee = Waste Electrical and Electronic Equipment

royduin, line 56:
```suggestion
                            <dd v-if="weeeTax">@lang('Fixed product tax')</dd>
```

PR #427, opened 2024-02-19
Unmake positions flattened

Jade-GG:
This breaks the default magento store for some reason, maybe it has an attribute named `positions`? It doesn't seem to be a problem elsewhere.

If that's the issue, we could also possibly just rename `positions` to something else instead, like `category_positions`.


Jade-GG:
https://github.com/rapidez/core/pull/390 this once again might end up being a decent fix instead?

PR #431, opened 2024-02-22
Add config change event

Jade-GG:
There are a few rare cases where, for example, you want to clear some Rapidez cache when some config is changed in Magento. If we run a job hourly, we can look at the `updated_at` field and see if any configurations have been changed within the past hour. (This would obviously also be put into the docs if this is the solution we go for)

Another option would be to keep track of the last time this job was run in the cache, and then using that to compare. However if you ever clear cache (especially in response to some config change), you would run into problems.

There's already an event that gets sent out within Magento whenever you change a config setting, but I don't think there's any feasible way for us to hook into that specific event?

### DIFF
--- a/src/Events/ConfigChangeEvent.php
+++ b/src/Events/ConfigChangeEvent.php
@@ -1,0 +1,17 @@
+<?php
+
+namespace Rapidez\Core\Events;
+
+use Illuminate\Foundation\Events\Dispatchable;
+
+class ConfigChangeEvent
+{
+    use Dispatchable;
+
+    public $config;
+
+    public function __construct($config)
+    {
+        $this->config = $config;
+    }
+}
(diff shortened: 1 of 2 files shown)


royduin:
When the indexer is scheduled hourly it will already clear the cache every time it runs. Don't think we should poll for changes; how often is the config changed? Mostly manually and in that case the enduser will notice it and should clear the cache.

Jade-GG:
> When the indexer is scheduled hourly it will already clear the cache every time it runs. Don't think we should poll for changes; how often is the config changed? Mostly manually and in that case the enduser will notice it and should clear the cache.

Clearing the cache was the example the original story came from so it's the one I have on hand, but I can imagine there may end up being other reasons to want to hook into configuration changes.

That's fair though and that makes this kind of moot in this situation---a solution in search of a problem.

royduin:
Maybe something for https://github.com/rapidez/magento2-compadre so we can push instead of poll.

royduin:
Closing this for now.

PR #433, opened 2024-02-26
Always add store & authorization header to rapidez API calls

Jade-GG:
This is useful for, for example, the multiple wishlists package which needs both.


resources/js/fetch.js
```diff
@@ -37,6 +37,8 @@ window.rapidezAPI = async (method, endpoint, data = {}, options = {}) => {
     let response = await rapidezFetch(window.url('/api/' + endpoint), {
         method: method.toUpperCase(),
         headers: {
+            Store: window.config.store_code,
+            Authorization: token.value ? `Bearer ${token.value}` : null,
```
indykoning, line 41: Can this still be overwritten using the options parameter?
In some cases we still need to send the cart hash via the authorization header right?

PR #434, opened 2024-02-26
Notify correctly in interactWithUser

Jade-GG:


resources/js/components/User/mixins/InteractWithUser.js
```diff
@@ -27,8 +27,7 @@ export default {
                     }
                 })
                 .catch((error) => {
-                    console.error(error)
```
indykoning, line 30: Do we no longer want to log these errors?

Jade-GG: [They already get logged with the notification](https://github.com/rapidez/core/blob/master/resources/js/package.js#L135) so this was just doubled up

PR #437, opened 2024-02-28
Feature/title component

JimmyHoenderdaal:


royduin:
Why do we need this?

JimmyHoenderdaal:
It is one of the base components that we implement in all Rapidez projects. Now, we can take this base from Rapidez and update it further if there is ever a need for more logic to be added. Otherwise, we would have to remake this component in every project, and maintaining updates would suffer.

royduin:
Feels like a separated boilerplate package is a better place for this as it's currently not actually used in the core.

PR #438, opened 2024-02-29
Removed the catalog/seo/product_url_suffix config default

royduin:
As the default in Magento is none.


royduin:
Seems like this isn't true: https://github.com/magento/magento2/blob/a2519ca33d89b974d929cf3ffcee142135e1f00e/app/code/Magento/Catalog/etc/config.xml#L44

PR #443, opened 2024-03-05
Move resizeobserver to nextTick function

indykoning:
In later versions of VueUse the old situation causes:
> TypeError: Cannot read properties of undefined (reading 'mounted')

or

> [Vue warn]: onMounted is called when there is no active component instance to be associated with. Lifecycle injection APIs can only be used during execution of setup().

Executing this in the nexttick ensures VueUse can resolve the current Vue instance

### DIFF
--- a/resources/js/components/Elements/Slider.vue
+++ b/resources/js/components/Elements/Slider.vue
@@ -53,12 +53,12 @@ export default {
     },
     mounted() {
         this.initSlider()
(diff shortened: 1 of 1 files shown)


royduin:
Could you also create a PR for master?

indykoning:
https://github.com/rapidez/core/pull/444

PR #448, opened 2024-03-12
Use script to get the positions from the flattened type

Jade-GG:
Supersedes #427.

We use a script here because the position from the flattened type, for some reason, cannot be implicitly cast to a number type *even though that's literally what it contains if you retrieve it directly from the index*.

By using a script, we can directly parse (not explicitly cast, that also doesn't work) it into a number. ES scripts get cached so this doesn't cause any sort of noticeable performance hit.

### DIFF
--- a/resources/views/components/listing.blade.php
+++ b/resources/views/components/listing.blade.php
@@ -39,7 +39,7 @@
                                 function_score: {
                                     script_score: {
                                         script: {
-                                            source: 'Integer.parseInt(doc[\'positions.'+(window.config.category.entity_id)+'\'].value)',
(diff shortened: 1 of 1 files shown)


jbclaudio (approved):

royduin:
So this is THE fix? We don't need any monkey patches?

Jade-GG:
> So this is THE fix? We don't need any monkey patches?

Correct, it's not pretty but it works flawlessly.

indykoning (approved):

royduin:
Please create another PR for master 😇

BobWez98:
Can be closed because of #467

PR #449, opened 2024-03-12
Load Magento urls from website/default base_url

indykoning:
This allows custom magento backend urls
Especially useful for multistore multidomain setups.
Where you may want to obfuscate that different sites belong to the same Magento backend.

### DIFF
--- a/src/Rapidez.php
+++ b/src/Rapidez.php
@@ -8,6 +8,7 @@ use Illuminate\Support\Collection;
 use Illuminate\Support\Facades\App;
 use Illuminate\Support\Facades\Event;
 use Rapidez\Core\Models\Config;
+use Rapidez\Core\Models\ConfigScopes;
 use Rapidez\Core\Models\Store;
 
 class Rapidez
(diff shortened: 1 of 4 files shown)


royduin:
With this we're getting multiple options to get config values: `getCachedByPath()` and `getValue()`, maybe merge it together?

Jade-GG (approved):

src/Rapidez.php
```diff
@@ -110,6 +112,26 @@ public function setStore(Store|array|callable|int|string $store): void
         config()->set('rapidez.root_category_id', $store['root_category_id']);
         config()->set('frontend.base_url', url('/'));
 
+        $magentoUrl = trim(
+            Config::getValue('web/secure/base_url', ConfigScopes::SCOPE_WEBSITE) ?? config()->get('rapidez.magento_url'),
+            '/'
+        );
+        if ($magentoUrl !== url('/')) {
```
Jade-GG, line 119: Any value in macro'ing the `UrlGenerator` helper for a function that compares urls like this?

indykoning: In most cases using a [Request::fullUrlIs()](https://github.com/laravel/framework/blob/28de931377aed9fc2935b62f4bbd4bedab3eca30/src/Illuminate/Http/Request.php#L235) or [Request::is()](https://github.com/laravel/framework/blob/28de931377aed9fc2935b62f4bbd4bedab3eca30/src/Illuminate/Http/Request.php#L211) is the way to go for comparing urls. 

Of course that only works with the current request and url, not one like `/` and the Magento url.
If you have any uses in mind i'll be happy to add it, if not i think it's use is probably too niche and we can add it later.

royduin:
Do we still need this? It's for 1.x which is pretty EOL. What about 2.x?

indykoning:
Yes this is still relevant, though solved in other ways per project we need an ability to define custom magento routes.
More importantly the new config system allows for flexibility i've seen lacking in the past.

Not relevant for 1.x anymore so moved to 2.x https://github.com/rapidez/core/pull/567

PR #450, opened 2024-03-13
Percy

royduin:


royduin:
This was a cool start, but has some disadvantages: Percy can get pretty expensive and it's only testing in 1 browser on 1 screensize. We could add a bulk of Dusk tests for different screen sizes but this seems like a better alternative which can test in multiple browsers: https://mmazzarolo.com/blog/2022-09-09-visual-regression-testing-with-playwright-and-github-actions/

PR #452, opened 2024-03-15
[2.x] Find flat attributes by checking the actual table schema

Jade-GG:
This way, it's impossible to have an attribute labelled as flat without actually being in the flat table.

PR #454, opened 2024-03-20
[1.x] Find flat attributes by checking the actual table schema

Jade-GG:
Forgot to make this one for 1.x (see also #452)


royduin:
Thanks! Could you check the Dusk tests? 😇

PR #458, opened 2024-03-26
[2.x] Implement weee_tax in 2.x

Jade-GG:
As far as I can tell we can always assume this data exists in the query, but I'm not 100% confident in this.

### DIFF
--- a/resources/js/mixins.js
+++ b/resources/js/mixins.js
@@ -16,4 +16,20 @@ Vue.mixin({
             }
         },
     },
+
+    computed: {
+        fixedProductTaxes() {
+            let taxes = {}
+            this.$root.cart.items.forEach(item => {
+                item.prices.fixed_product_taxes.forEach(tax => {
+                    if(tax.label in taxes) {
+                        taxes[tax.label] += tax.amount.value
+                    } else {
+                        taxes[tax.label] = tax.amount.value
+                    }
+                })
+            })
+            return taxes
+        }
+    }
 })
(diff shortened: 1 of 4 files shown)


resources/views/product/partials/options.blade.php
```diff
@@ -2,7 +2,7 @@
     <div class="flex flex-col space-y-3">
         @foreach ($product->options->sortBy('sort_order') as $option)
             <div>
-                @include('rapidez::product.partials.options.'.$option->type)
+                @includeIf('rapidez::product.partials.options.'.$option->type)
```
royduin, line 5: Sure? All product option types are implemented?

Jade-GG: I guess you probably do want an error if you haven't implemented one?

royduin: On a test environment; not on production. Maybe use this one: https://github.com/rapidez/blade-directives#includefirstsafe which does that.

resources/js/mixins.js
```diff
@@ -16,4 +16,20 @@ Vue.mixin({
...
+                item.prices.fixed_product_taxes.forEach((tax) => {
+                    if (tax.label in taxes) {
+                        taxes[tax.label] += tax.amount.value
+                    } else {
+                        taxes[tax.label] = tax.amount.value
+                    }
+                })
+            })
+            return taxes
+        },
```
indykoning, line 33: Is this something we'd want to move to the cartStorage instead?
Like we have done with virtualitems computed properties
https://github.com/rapidez/core/blob/master/resources/js/stores/useCart.js#L99

PR #460, opened 2024-03-27
Removed classes from base button | Placed paddings on buttons where needed

Roene-JustBetter:
We don't need al of this classes on the button. 

- I removed classes from the base button that we don't need.
- Placed paddings on the buttons where we need it.

### DIFF
--- a/resources/views/components/button/base.blade.php
+++ b/resources/views/components/button/base.blade.php
@@ -8,7 +8,7 @@
 <x-tag
     is="{{ $tag }}"
     {{ $attributes->merge([
-        'class' => 'flex items-center justify-center font-semibold py-2 px-4 border rounded disabled:opacity-50 disabled:cursor-not-allowed hover:opacity-75 whitespace-nowrap transition',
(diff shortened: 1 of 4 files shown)


Roene-JustBetter:
@jordythevulder @JimmyHoenderdaal

JimmyHoenderdaal (approved):

jordythevulder:
Looks good! It has my approval

indykoning:
What is the impact of this? It looks to me like it isn't breaking as long as you've been using the normal button components.
Do we often use the base button?

indykoning (dismissed):

royduin:
What's changed visually?

Roene-JustBetter:
> What's changed visually?

Nothing changed visually. What I did is removed & replaced classes from the button/base to keep this clean and placed classes on buttons where it needed. For example not all buttons need the same padding (buttons used for slider) and only the outline button needs a border instead of every button.

royduin:
Did you test this? I'm reverting it with #499 as this is the result now:

image

PR #461, opened 2024-04-02
Fix adding to cart

BobWez98:


indykoning (approved):

BobWez98:
Yes, the simpleProduct is a computed property and will update on a super attribute change. https://github.com/rapidez/core/blob/master/resources/js/components/Product/AddToCart.vue#L208

If you don't use the simple product it will try to add the configurable product itself to the cart which will always be out of stock.

PR #462, opened 2024-04-02
[1.x] Clear all data on token invalidation

Jade-GG:
This caused lingering user data like addresses with a `customer_address_id` to persist in localstorage under certain conditions, blocking the guest checkout from being used.

### DIFF
--- a/resources/js/stores/useUser.js
+++ b/resources/js/stores/useUser.js
@@ -8,7 +8,7 @@ let isRefreshing = false
 
 export const refresh = async function () {
     if (!token.value) {
-        userStorage.value = {}
+        clear()
         return false
     }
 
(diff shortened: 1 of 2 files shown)


indykoning (approved):

indykoning (approved):

resources/js/stores/useUser.js
```diff
@@ -8,7 +8,7 @@ let isRefreshing = false
 
 export const refresh = async function () {
     if (!token.value) {
-        userStorage.value = {}
+        clear()
```
indykoning, line 11: Looks like a good fix, i just want to make sure we don't cause an infinite loop with this as:
1. Clear sets token.value to empty
2. we watch token, and call refresh when it changes
3. we check if the token is empty, then we call Clear

As this doesn't do any api calls and should be done quick it's fine if it calls twice. Just not infinitely.
You can test this out by adding a console.log here and logging out/removing your token from localstorage

royduin:
Any updates?

PR #463, opened 2024-04-08
Reset styling of the listing component

indykoning:
During render when vue hasn't replaced the components yet some browsers add styling to the listing component.

There is barely any information on it online however it seems that this was an unused, never implemented component that has been long deprecated.
https://html.com/tags/listing/
however some browsers still come with their own styling for it.

Especially the whitespace: pre; and margin: 1em 0px; is problematic.

display: contents; is because we don't really want vue components that get removed anyways to interact with the layout

### DIFF
--- a/resources/css/app.css
+++ b/resources/css/app.css
@@ -15,7 +15,7 @@
     -webkit-tap-highlight-color: transparant;
 }
 listing {
-/*  Reset browser added styling causing unexpected behavior */
+    /*  Reset browser added styling causing unexpected behavior */
     display: contents;
     font-family: unset;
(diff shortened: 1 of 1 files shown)


royduin:
Maybe add some more info in the comment about where it is used.

indykoning:
Listing is used due to our [Listing.vue](https://github.com/rapidez/core/blob/a48673af095e723372834e53da59b439d719c38f/resources/views/components/listing.blade.php#L6) on category pages.
Until Vue is able to replace the <listing><listing/> tag your browser adds this styling, causing layout shifts.

PR #464, opened 2024-04-08
[2.x] Rework slideover

Jade-GG:
This reworks the slideover to be simpler and work with forms/checkboxes instead of js. Also adds the toggleScroll functionality.

Note that with this change we cannot use vue-bound IDs for slideovers. The checkout agreements section had to be slightly reworked to avoid having vue load multiple slideovers into the dom.


resources/views/checkout/steps/payment.blade.php
```diff
@@ -19,33 +19,33 @@ class="ml-auto w-8 h-8"
...
+                </label>
+                <div v-else>
+                    <x-rapidez::checkbox
+                        name="agreement_ids[]"
+                        v-bind:value="agreement.agreement_id"
+                        v-model="checkout.agreement_ids"
+                        dusk="agreements"
+                        required
+                    >
+                        <label class="text-gray-700 cursor-pointer underline hover:no-underline" for="checkoutAgreement" v-on:click="window.Vue.set($root.custom, 'openAgreement', agreement)">
```
Jade-GG, line 44: Vue was being a jerk about reactivity here, this was the only approach that worked.

JimmyHoenderdaal (approved):

resources/js/package.js
```diff
@@ -103,6 +103,10 @@ function init() {
             setSearchParams(url) {
                 window.history.pushState(window.history.state, '', new URL(url))
             },
+            toggleScroll(bool) {
```
indykoning, line 106: It might be a nice idea to implement https://vueuse.org/core/useScrollLock/ so locked state can be read as well as written, and it seems to have a fix for IOS supposedly not listening to scroll locks properly

resources/views/checkout/steps/payment.blade.php
indykoning: Do you know why reactivity wasn't functioning as expected?
Are you trying to set a string/null value to an object value without a deep?

I know vue is not always a fan of set via `window.app.custom`, even `$root` sometimes. But that is not necessary, it's just more verbose.

Jade-GG: I tried using:
- Regular assignment and Vue.set() on the `data` object we have, adding an extra custom key to use
- Same as above but also using `_renderProxy`
- Regular assignment and Vue.set() on a variable that I added in the slot-scope and then initialized via the classic `:set` hack
- Regular assignment and Vue.set() on `$root.custom`

Of all these, only the Vue.set() on $root.custom worked. Everything else didn't react properly and didn't have their values updated until you clicked a checkbox (which is very strange in and of itself). I tried to debug Vue itself with this but debugging Vue in-vivo is a pain and I didn't really get anywhere.

indykoning: I think i see your issue. this value is never defined on custom, so reactivity is never created for this nested value.

You should create it in https://github.com/rapidez/core/pull/464/files/0c8a29224561f26fc5cdfd2dbdd38284d1973b40#diff-251b99eac1abebedfe294a2334032b29b5f843d0b5d0834c32c055635f79c935R59 first

resources/views/components/slideover/index.blade.php
```diff
@@ -0,0 +1,24 @@
+@props(['id', 'title', 'hasParent' => false, 'right' => false, 'isForm' => false])
```
indykoning, line 1: I'm not too sure what hasParent means, maybe we can clear this name up?
But i think using position, and tag or type div or form will make the usage easier to understand
```suggestion
@props(['id', 'title', 'hasParent' => false, 'position' => 'left', 'tag/type' => 'div'])
```

indykoning (changes requested):

resources/js/package.js
```diff
@@ -56,7 +56,7 @@ function init() {
     window.app = new Vue({
         el: '#app',
         data: {
-            custom: {},
+            custom: { openAgreement: null },
```
royduin, line 59: This is for custom stuff people build, not for the core itself. Can we do this without this?

resources/views/components/slideover/partials/header.blade.php
```diff
@@ -0,0 +1,21 @@
+<div class="slideover-header bg-primary py-5">
+    <div class="px-5">
+        <div class="relative flex items-center justify-center">
+            @if ($hasParent)
+                <label class="absolute left-0 top-1/2 -translate-y-1/2 cursor-pointer text-white" for="{{ $id }}">
```
Roene-JustBetter, line 5: When this label is clicked the 
`<input @checked($open) id="{{ $id }}" v-on:change="toggleScroll($event.target.checked)" class="peer hidden" type="checkbox"> `
Will change and it will remove the scrolllock on the body so you are be able to scroll again.

resources/js/package.js
```diff
@@ -61,10 +61,12 @@ function init() {
             loadingCount: 0,
             loading: false,
             loadAutocomplete: false,
+            openAgreement: null,
```
royduin, line 64: Can we move this somewhere else? As it isn't a global shop thing.

resources/views/components/slideover/index.blade.php
```diff
@@ -0,0 +1,25 @@
+@props(['id', 'title', 'hasParent' => false, 'position' => 'left', 'tag' => 'div', 'open' => false])
```
royduin, line 1: Maybe add some info on all options?

resources/js/package.js
Jade-GG: What would you suggest? Could turn it into a vue component but that seems unnecessarily bulky.

Roene-JustBetter (approved):

resources/js/package.js
royduin: Cant you dynamically handle this in the template?

JimmyHoenderdaal (approved):

resources/js/package.js
Jade-GG: Nope, not without breaking reactivity. The entire reason I need to define it somewhere is so that vue can pick up on the variable and make it properly reactive.

royduin: Maybe store it in a hidden input? Or use a radio to know which one is open?

Jade-GG: What I've done for now is change the slideover so that it accepts `v-bind:id`, which means we can use one slideover per agreement again.

resources/views/checkout/steps/payment.blade.php
```diff
@@ -18,21 +18,22 @@ class="ml-auto w-8 h-8"
         </x-rapidez::radio>
     </div>
 
-    <graphql query="{ checkoutAgreements { agreement_id name checkbox_text content is_html mode } }">
+    <graphql query="{ checkoutAgreements { agreement_id name checkbox_text content is_html mode } }" cache="ca">
```
royduin, line 21:
```suggestion
    <graphql query="{ checkoutAgreements { agreement_id name checkbox_text content is_html mode } }" cache="checkout.agreements">
```

PR #466, opened 2024-04-09
[1.x] Fix :key

Jade-GG:


indykoning (approved):

resources/views/layouts/partials/header/autocomplete/products.blade.php
```diff
@@ -2,7 +2,7 @@
     <li
         v-for="suggestion in suggestions"
         :key="suggestion.source._id">
-        <a :href="suggestion.source.url | url" class="flex flex-wrap flex-1" key="suggestion.source._id">
+        <a :href="suggestion.source.url | url" class="flex flex-wrap flex-1" :key="suggestion.source._id">
```
indykoning, line 5: I see the same suggestion key on the line above, is this right?
Can we remove the key on this line without problem, as the looping `li` already provides a unique key

indykoning (changes requested):

PR #471, opened 2024-04-15
Remove unnecessary toArray calls in product indexer

Jade-GG:
If you're using a callable dataFilter, this toArray call was being done unnecessarily.

The specific reason I'm doing this now is because Statamic has a bug creating an infinite loop calling `toArray()` on entries under certain specific conditions. This is really just a workaround for that, but I can imagine other situations where the extraneous `toArray()` call is also annoying.

### DIFF
--- a/src/Commands/ElasticsearchIndexer.php
+++ b/src/Commands/ElasticsearchIndexer.php
@@ -49,8 +49,8 @@ class ElasticsearchIndexer
 
         $currentValues = match (true) {
             is_callable($dataFilter) => $dataFilter($item),
-            is_null($dataFilter)     => $this->makeArray($item),
-            default                  => Arr::only($this->makeArray($item), $dataFilter),
+            is_null($dataFilter)     => $item,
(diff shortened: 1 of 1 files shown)


src/Commands/ElasticsearchIndexer.php
```diff
@@ -47,11 +47,10 @@ public function indexItem(object $item, callable|array|null $dataFilter, callabl
             return;
         }
 
-        $arrItem = $item instanceof Arrayable ? $item->toArray() : (array) $item;
         $currentValues = match (true) {
             is_callable($dataFilter) => $dataFilter($item),
-            is_null($dataFilter)     => $arrItem,
-            default                  => Arr::only($arrItem, $dataFilter),
+            is_null($dataFilter)     => $this->makeArray($item),
```
indykoning, line 52: Do we need this in the first place? 
The job already ensures the values is an array https://github.com/rapidez/core/blob/c5724460e7ae83b7b5d6f18cc47766ff4ea1fade/src/Jobs/IndexJob.php#L21

indykoning (approved):

PR #473, opened 2024-04-16
Add category filter styling

JimmyHoenderdaal:


resources/js/components/Listing/Filters/CategoryFilterCategory.vue
```diff
@@ -31,12 +31,7 @@ export default {
             type: Object,
             default: function () {
                 return {
```
indykoning, line 33: This looks breaking if it has been used to style the categories, don't we want to maintain that compatibility?

Roene-JustBetter (approved):

resources/views/listing/partials/filter/category.blade.php
```diff
@@ -10,9 +10,7 @@
     <div slot-scope="{ aggregations, setQuery, value }">
         <category-filter :aggregations="aggregations" :value="value" :set-query="setQuery">
             <div slot-scope="{ hasResults, results }">
-                <strong v-if="hasResults">@lang('Category')</strong>
-                <input class="hidden" type="radio" name="category" value="" :checked="!value" v-on:change="setQuery({})"/>
```
indykoning, line 14: This might hinder functionality if removed. I suggest leaving it in.

As for the category title; every other filter has a title, i think the default should be to have that title. With the option to remove it

resources/css/components/category.css
```diff
@@ -0,0 +1,40 @@
+.pop-category {
```
indykoning, line 1: I think something along ".listing-category" or ".filter-category"/".category-filter" would be better naming, considering POP is Dutch

resources/js/components/Listing/Filters/CategoryFilterCategory.vue
Roene-JustBetter: For me I think it is not necessary to keep te compatibility if we merge this in 2.x

indykoning (dismissed):

indykoning:
The tests are currently failing because of
> The `bg-inactive-110` class does not exist. If `bg-inactive-110` is a custom class, make sure it is defined within a `@layer` directive.

PR #478, opened 2024-04-23
[2.x] Fix ForCurrentStoreScope

Jade-GG:
Using a lateral join, which limits the SQL compatibility to [PostgreSQL, MySQL >= 8.0.14, and SQL Server](https://laravel.com/docs/11.x/queries#lateral-joins).

Basically this is fixing the fact that the previous query could only select one thing (caused by the `->limit(1)`). This makes it useful for stuff like grabbing the checkoutagreements.


indykoning (approved):

royduin:
I think we should update the requirements page in the docs with this change.

PR #480, opened 2024-04-25
[2.x] Round prices before multiplying by qty

Jade-GG:
Magento rounds prices to 2 decimals before they get multiplied by qty. Not doing this can cause some slightly incorrect prices to be shown on the PDP, especially in combination with tier price discounts.

(More specifically, it converts to the given currency before multiplying by qty by calling `currencyByStore`. This isn't always 2 decimals, see [the ISO 4217 specification](https://en.wikipedia.org/wiki/ISO_4217#List_of_ISO_4217_currency_codes)---9 currencies have more than 2 decimals, and 17 currencies have no decimals at all)

### DIFF
--- a/resources/js/components/Product/AddToCart.vue
+++ b/resources/js/components/Product/AddToCart.vue
@@ -137,7 +137,9 @@ export default {
 
         calculatePrices: function () {
             this.price = Math.round((parseFloat(this.simpleProduct.price) + this.priceAddition(this.simpleProduct.price)) * 100) / 100
-            this.specialPrice = Math.round((parseFloat(this.simpleProduct.special_price) + this.priceAddition(this.simpleProduct.special_price)) * 100) / 100
+            this.specialPrice =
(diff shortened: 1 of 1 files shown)


royduin (approved):

Jade-GG:
To get around the different currencies problem we could use something like

`+(new Intl.NumberFormat('en-US', {style: 'currency', currency: config.currency}).format(price).replace(/[^0-9.]/g,''))`

to let the browser handle it, but it's incredibly ugly.

royduin:
We already have a price filter: https://github.com/rapidez/core/blob/master/resources/js/filters.js

Jade-GG:
> We already have a price filter: https://github.com/rapidez/core/blob/master/resources/js/filters.js

This is true, however this also uses the locale, which changes the decimal separator and doesn't let us easily convert the number back into a float 🙃

royduin:
Waiting on a review from @indykoning

indykoning (approved):

PR #486, opened 2024-05-02
Fixed parent product link for configurable products

indykoning:
Without this the product is not linked to it's parent, meaning that product options of the parent aren't considered.
Nor are the selected configurable options shown in the cart as you've added a simple product instead of a configurable product

### DIFF
--- a/resources/js/components/Product/AddToCart.vue
+++ b/resources/js/components/Product/AddToCart.vue
@@ -89,7 +89,7 @@ export default {
                         config.queries.cart +
                         ` } user_errors { code message } } }`,
                     {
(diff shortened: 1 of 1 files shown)


royduin:
Are we sure this time? https://github.com/rapidez/core/pull/461

cc @BobWez98

BobWez98:
I think i made a mistake back when i changed it to SimpleProduct. We were thinking back then every variant was in stock because the disabledOptions weren't working properly. I tested this and i agree 👍 @royduin

BobWez98 (approved):

PR #489, opened 2024-05-03
Add health check for usage of Rapidez model

indykoning:
This checks if any of the models written (or overwritten) in `config/rapidez/models.php` are not extending Rapidez' Model.
And warns if it has found any models.

As this is not breaking to the application the healthcheck will not fail.

Sample output:
image

### DIFF
--- a/src/Listeners/Healthcheck/Base.php
+++ b/src/Listeners/Healthcheck/Base.php
@@ -1,0 +1,34 @@
+<?php
+
+namespace Rapidez\Core\Listeners\Healthcheck;
+
+use Illuminate\Support\Facades\Event;
+
+class Base
+{
+    public function handle()
+    {
+        return [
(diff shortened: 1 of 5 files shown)


src/Listeners/Healthcheck/Base.php
```diff
@@ -0,0 +1,34 @@
...
+                //     'value' => ''.
+                // ],
+                // [
+                //     'type'  => 'warn',
+                //     'value' => ''.
+                // ],
+                // [
+                //     'type'  => 'info',
+                //     'value' => ''.
+                // ],
```
royduin, line 25: This is just here as an example?

indykoning: Exactly, it is just an example of the expected output. As the handle function of Base is made to be overwritten

PR #494, opened 2024-05-10
[1.x] Add backorders

Jade-GG:
Will wait to do the 2.x version of this until this is approved.

### DIFF
--- a/resources/views/product/partials/addtocart.blade.php
+++ b/resources/views/product/partials/addtocart.blade.php
@@ -1,7 +1,7 @@
 <add-to-cart :default-qty="{{ $product->min_sale_qty > $product->qty_increments ? $product->min_sale_qty : $product->qty_increments }}">
     <form slot-scope="{ _renderProxy: addToCartSlotProps, options, customOptions, error, add, disabledOptions, simpleProduct, adding, added, price, specialPrice, setCustomOptionFile }" v-on:submit.prevent="add">
         <h1 class="mb-3 text-3xl font-bold" itemprop="name">{{ $product->name }}</h1>
-        @if (!$product->in_stock)
+        @if (!$product->in_stock && Rapidez::config('cataloginventory/item_options/backorders', 0) == 0)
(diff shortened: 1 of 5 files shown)


resources/views/product/partials/addtocart.blade.php
```diff
@@ -18,6 +18,14 @@
             </div>
 
             @include('rapidez::product.partials.options')
+
+            @if ($product->qty <= 0 && Rapidez::config('cataloginventory/item_options/backorders', 0) == 2)
```
royduin, line 22: If a product can be backordered can also be configured on the product

src/RapidezServiceProvider.php
```diff
@@ -105,6 +107,23 @@ protected function bootCommands(): self
...
     }
 
+    protected function bootFilters(): static
+    {
+        if (app()->runningInConsole()) {
+            return $this;
+        }
+
+        if (RapidezFacade::config('cataloginventory/item_options/backorders', 0) == 2) {
+            Eventy::addFilter('quote.items.select', function ($query) {
```
royduin, line 117: Maybe it's cleaner to have this in the quote model?

src/Models/Quote.php
```diff
@@ -62,8 +64,9 @@ protected static function booting()
                         "total_excl_tax", quote_item.row_total,
                         "attributes", quote_item_option.value,
                         "type", quote_item.product_type
-                    QUERY) . '
-                )), "$.null__") AS items')
+                    QUERY
+                        . '"backorder_count", IF(IF(stock.use_config_backorders, ' . $configBackorder . ', stock.backorders) = 2, GREATEST(0, quote_item.qty - stock.qty), 0)'
```
Jade-GG, line 68: You'd think that running the magento indexer will set the `stock.backorders` value to whatever the config setting is when `use_config_backorders` is enabled.

Not the case. If you change this setting on a product the value will be set correctly, but if you change the global config setting they won't be. Hence, I have to do a gross query like this.

indykoning (approved):

PR #496, opened 2024-05-14
[2.x] Enable highlighting in autocomplete products

Jade-GG:
This PR enables highlighting from elasticsearch in the product listing, which puts the `<mark></mark>` tags around matched text on every product title. This was already enabled by default on categories and any other extra indices, so this is for consistency.


royduin:
Could you provide some more info so @jordythevulder can make a decision if we want this by default on all shops?

royduin:
It's used on everything but the products; so yes: we want this.

PR #512, opened 2024-06-05
[1.x] Don't expose negative stock values

Jade-GG:
Adding a GREATEST here to make sure you don't get to see what the "negative stock" may be on a product.


royduin:
What about 2.x?

Jade-GG:
> What about 2.x?

2.x already works correctly with the way the compadre module for this was written 😇

PR #516, opened 2024-06-14
Add backorders to rapidez 2.0

BobWez98:


resources/views/cart/queries/cart.graphql
```diff
@@ -3,6 +3,7 @@ total_quantity
 items {
     id
     quantity
+    qty_backordered
```
indykoning, line 6: Is this the [value added in Compadre](https://github.com/rapidez/magento2-compadre/pull/3)? 
If so it must be wrapped in the Compadre version check with the [module.xml version](https://github.com/rapidez/magento2-compadre/blob/master/src/etc/module.xml#L4) to prevent errors when it is not installed/updated.

In your code you'll need to account for the key being missing

resources/views/cart/item.blade.php
```diff
@@ -28,6 +28,18 @@ class="mx-auto"
                         </template>
                     </div>
                     @include('rapidez::cart.item.remove')
+                    <div v-if="item.backorder_count" class="flex gap-2">
```
indykoning, line 31: Shouldn't this be `qty_backordered`?

resources/views/cart/item.blade.php
```diff
@@ -28,6 +28,18 @@ class="mx-auto"
...
+                        <x-heroicon-o-exclamation-circle class="mt-px w-5" />
+                        <span>
+                            <template v-if="item.backorder_count == item.qty">
```
indykoning, line 34: It looks like this if case is flipped, or could be `item.backorder_count < item.qty`

indykoning (changes requested):

resources/views/cart/item.blade.php
BobWez98: I'd say the other way around, if the qty is the stock and it's lower than the backorder count, there are products to be backordered.

resources/views/cart/queries/cart.graphql
```diff
@@ -3,6 +3,9 @@ total_quantity
 items {
     id
     quantity
+    @if (Rapidez::checkCompadreVersion('0.0.4'))
```
indykoning, line 6: It's unfortunate that it does not match the current release but it should match the version of the module
https://github.com/rapidez/magento2-compadre/blob/master/src/etc/module.xml#L4
not the release

resources/views/cart/item.blade.php
```diff
@@ -28,6 +28,18 @@ class="mx-auto"
                         </template>
                     </div>
                     @include('rapidez::cart.item.remove')
+                    <div v-if="item.qty_backordered" class="flex gap-2">
+                        <x-heroicon-o-exclamation-circle class="mt-px w-5" />
+                        <span>
+                            <template v-if="item.qty_backordered > item.qty">
```
indykoning, line 34: I'm not sure of the format qty_backordered comes in so i could be wrong.
Wouldn't it be impossible for the qty_backordered to be higher than the qty ordered?

PR #517, opened 2024-06-20
[2.x] Add fixed product taxes in cart

Jade-GG:
As the fixed product tax [can not be fetched on the total prices](https://github.com/magento/magento2/blob/2.4-develop/app/code/Magento/WeeeGraphQl/etc/schema.graphqls), we have to individually sum all products together :)


royduin:
Do we also need to show this somewhere in the checkout?

Jade-GG:
> Do we also need to show this somewhere in the checkout?

Good point, we also have the cart data there---I'll make a PR to add this.

PR #518, opened 2024-06-26
Add the ability to attach notifications to the session

jbclaudio:
This adds the ability to attach notifications to the session.

As an example, you could add the following to a controller function:
```
return redirect('/')->with(['notification' => [
    'message' => 'Lorem ipsum dolor sit amet, consectetur adipiscing elit.',
    'type' => 'success',
    'show' => true
]]);
```

When the redirect is done and `notification` is sent along with it, the notification will be displayed.

### DIFF
--- a/.github/workflows/duster-fix-blame.yml
+++ b/.github/workflows/duster-fix-blame.yml
@@ -17,6 +17,7 @@ jobs:
       - uses: actions/checkout@v3
         with:
           ref: ${{ github.head_ref }}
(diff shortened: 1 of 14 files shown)


indykoning (approved):

indykoning:
Don't forget to update your description for `notifications` in case we look back here in the future.
Maybe add something in the https://github.com/rapidez/docs

royduin:
Some info here: https://docs.rapidez.io/2.x/package-development.html would be nice.

jbclaudio:
I've added some info to the Rapidez Docs in https://github.com/rapidez/docs/pull/46.

PR #519, opened 2024-06-26
[1.x] Always attempt to decode attribute value

Jade-GG:
I've only just run into this being a problem by using a package that adds custom options here that aren't represented by magento options.

### DIFF
--- a/src/Models/QuoteItemOption.php
+++ b/src/Models/QuoteItemOption.php
@@ -30,7 +30,7 @@ class QuoteItemOption extends Model
                 default           => (function () use ($value) {
                     if (! $this->option) {
                         if (is_string($value)) {
-                            return json_decode($value);
+                            return json_decode($value) ?? $value;
                         }
 
                         return null;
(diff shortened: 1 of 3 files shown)


Jade-GG:
> should we add a way to extend the capabilities here so this can be added and transformed correctly?

Perhaps? I'm not sure how it'd work, possibly by way of eventy filter at the very start of the default case, but then actually returning would be a little weird.

Maybe for now it's ok to have it like this, and make it more extendable if we run into another case that would need this?

src/Models/QuoteItemOption.php
```diff
@@ -23,26 +23,20 @@ public function quote_item()
...
-                        return json_decode($value);
-                    }
+                if ($this->code === 'option_ids') {
+                    $value = explode(',', $value);
+                }
 
-                    return $value;
-                })()
-            })->shouldCache();
+                return json_decode($value) ?? $value;
```
indykoning, line 37: Just to be sure php won't complain about types
```suggestion
                return is_string($value) ?  (json_decode($value) ?? $value) : $value;
```

PR #520, opened 2024-06-26
[2.x] At least return something when there is a value attached to the option

Jade-GG:
see #519


royduin:
Do you have a bit more info and a good squash commit message?

Jade-GG:
> Do you have a bit more info and a good squash commit message?

Some custom types of options (e.g. custom product options created by expivi) would not work well in the old situation---see the old PR linked in the message. `Always attempt to decode attribute value` is what that one was described as, which makes sense to me.

PR #530, opened 2024-07-15
Fix missing attributeValues and autoselect config errors

BobWez98:
When updating a project to Rapidez 2.x i came across these problems. the `auto_select_*` gave an error and the cart items were not reactive in the callback.js.

### DIFF
--- a/resources/js/stores/useCart.js
+++ b/resources/js/stores/useCart.js
@@ -101,13 +101,49 @@ export const cart = computed({
         return cartStorage.value
     },
     set(value) {
-        cartStorage.value = value
-        age = Date.now()
+        getAttributeValues().then((response) => {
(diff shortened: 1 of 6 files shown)


resources/js/stores/useCart.js
```diff
@@ -101,13 +101,49 @@ export const cart = computed({
...
+                return cartItem
+            })
+
+            cartStorage.value = value
+            age = Date.now()
+
+            if (value.id && value.id !== mask.value) {
+                // Linking user to cart will create a new mask, it will be returned in the id field.
+                mask.value = value.id
+            }
```
indykoning, line 145: Can we place this in a `.finally()`? so if the `getAttributeValues` fails we still have a functioning cart

indykoning (approved):

PR #532, opened 2024-07-16
Feature/inputs

Roene-JustBetter:
What we have changed:

- Visual small changes
- Moved inputs to the components/input
- Inside components/input we have made folders for example **radio** in there we have _base.blade.php_ & _index.blade.php_

So now we can use the inputs in different ways. The most common way is to use it with a label for example: 
```
<x-rapidez::input
    :label="__('Firstname')"
    name="{{ $type }}_firstname"
    v-model.lazy="checkout.{{ $type }}_address.firstname"
    required
/>
```
This will give you an input with a label see below:
Scherm­afbeelding 2024-07-29 om 15 40 49

If you want to use an input without a label you can use the base variant
```
  <x-rapidez::input.base
      name="email"
      type="email"
      placeholder="Email"
      v-bind:value="email"
      v-on:input="loginInputChange"
      required
  />
```
This will result in an input without a label:
Scherm­afbeelding 2024-07-29 om 15 42 34

All the inputs we have in **components/input**:
- Input (type="text", type="email", ect")
- select
- radio
- checkbox
- textarea
- password

Scherm­afbeelding 2024-07-16 om 11 05 16
Scherm­afbeelding 2024-07-16 om 11 05 25
Scherm­afbeelding 2024-07-16 om 11 05 38

**If we want to use this inputs we should change the inputs in Rapidez/account as well.**


Roene-JustBetter:
@jordythevulder @JimmyHoenderdaal

jordythevulder:
Looks great! I only have a few doubts about the focus coloring. Looks fancy, but will be probably be overwritten in our projects. Maybe not bad at all but it's just worth mentioning. It's a good base and easy to use!

Also doubting about adding a base.blade.php inside the input directory with base classes for all inputs
- index/ text
- checkbox
- radio
- password
- select
- textarea

With base classes like: border rounded padding-x padding-y hover/focus states. So when you want to make a small change to all of the input components, you only have to overwrite 1 file instead of them all when you want to add a shadow for example.

What do you think?

royduin:
Seems like the single checkbox/radio component is removed, how can I get just a checkbox or radio without a label? I think we now are getting a double label in `resources/views/checkout/steps/payment.blade.php` with the radios.

resources/views/components/input/checkbox.blade.php
```diff
@@ -1,5 +1,40 @@
...
+            Accept!
+        </span>
+    </x-rapidez::input.checkbox>
+</template>
+```
 --}}
-<input type="checkbox" {{ $attributes->twMerge('focus:ring-neutral h-4 w-4 text-neutral border rounded') }}>
+
+<label {{ $attributes->only('class')->twMerge('inline-flex text-neutral text-sm cursor-pointer has-[:required]:after:content-[\'*\']') }}>
+    <input type="checkbox" class="cursor-pointer border rounded-md focus:ring-neutral size-5 mr-2.5 text-neutral border-border focus:outline-none focus:ring-0 focus:ring-offset-0" {{ $attributes->except('class') }}>
```
JimmyHoenderdaal, line 38: if the slot is empty there is a empty mr-2.5, not sure if this would ever occur

resources/views/components/input/label.blade.php
```diff
@@ -0,0 +1,19 @@
...
+    <x-rapidez::input name="firstname"/>
+</label>
+```
+By default a label has an asterix when the input is required. If you don't want an asterix
+you can use the label like this:
+```
+<x-label class="after:hidden" />
+```
+--}}
+<span {{ $attributes->twMerge("font-medium text-inactive text-sm inline-block mb-2 duration-300 transition-colors has-[~_*_:focus,~:focus]:text-primary has-[~_*_:required,~:required]:after:content-['*'] has-[~:user-invalid:not(:placeholder-shown)]:text-red-500") }}>
```
JimmyHoenderdaal, line 17: 🤯

JimmyHoenderdaal (approved):

jordythevulder (approved):

resources/views/components/input/label.blade.php
```diff
@@ -0,0 +1,19 @@
+{{--
+It's called a label but it isn't? 🤯 Yup, as it's just for styling 🎨
+The benefit of this approach is that you don't have to deal with
+"for" attributes on labels and matching input names. Usage:
```
indykoning, line 4: My worry here is that the naming may by confusing leaving users to use this component incorrectly.
Have other names been considered?

Roene-JustBetter: I think nobody will use a label without an input. We also moved the label to the input folder because you have to use it like this: `<x-rapidez::input.label>` instead of: `<x-rapidez::label>` see also: [#451](https://github.com/rapidez/core/pull/451/files#diff-38d7765c3a5fdea6a907b6d43c615d9d1e0a9bcadca894e86bfa6dc3d3e14d37)

JimmyHoenderdaal: @indykoning I get your concern about the potential confusion with the naming. However, I prefer using label because it accurately describes the component’s function. This component will always be wrapped around an input, reinforcing its role as a label. I think that using this term helps maintain readability and prevents misuse, making the component’s purpose immediately recognizable.

Typically, a label sits above an input, which is why we call it a label. By wrapping the text inside the label, we simplify it a bit and eliminate the need to add a “for” attribute.

resources/views/components/input/textarea/base.blade.php
```diff
@@ -0,0 +1,23 @@
+{{--
+Examples:
+
+1. Just an textarea without a label:
```
jordythevulder, line 4: typo: just a textarea

resources/views/components/input/textarea/base.blade.php
```diff
@@ -0,0 +1,23 @@
...
+```
+
+2. With a label:
+```
+<x-rapidez::input.textarea label="something"/>
+```
+
+3. With a label as a slot:
+```
+<x-rapidez::input">
```
jordythevulder, line 16: .textarea missing here

resources/views/product/partials/options/file.blade.php
```diff
@@ -1,13 +1,12 @@
...
-            class="px-0"
-        />
-    </label>
+    <x-rapidez::input
+        type="file"
+        :required="$option->is_require"
+        v-on:change="addToCart.setCustomOptionFile($event, {{ $option->option_id }})"
+        class="px-0"
+    >
+        {{ $option->title }} {{ $option->price_label }}
```
jordythevulder, line 8: is this correct? shouldn't this be the label? now it's underneath your input as the slot is at the bottom

resources/views/product/partials/options/field.blade.php
```diff
@@ -1,8 +1,7 @@
...
-        :maxlength="$option->max_characters ?: false"
-        v-model="addToCart.customOptions[{{ $option->option_id }}]"
-    />
-</label>
+<x-rapidez::input
+    :required="$option->is_require"
+    :maxlength="$option->max_characters ?: false"
+    v-model="addToCart.customOptions[{{ $option->option_id }}]"
+>
+    {{ $option->title }} {{ $option->price_label }}
```
jordythevulder, line 6: same here

resources/views/product/partials/options/field.blade.php
```diff
@@ -1,8 +1,7 @@
...
-        :maxlength="$option->max_characters ?: false"
-        v-model="addToCart.customOptions[{{ $option->option_id }}]"
-    />
-</label>
+<x-rapidez::input
+    :required="$option->is_require"
+    :maxlength="$option->max_characters ?: false"
+    v-model="addToCart.customOptions[{{ $option->option_id }}]"
+>
+    <x-slot:label>{{ $option->title }} {{ $option->price_label }}</x-slot:label>
```
royduin, line 6: Attributes on this slot doesn't have effect? When I do `<x-slot:label class="text-red-500">`

jordythevulder (approved):

royduin:
Closing as we're going to split this in a components package later.

PR #533, opened 2024-07-22
Order item product relation

BobWez98:
On the checkout succes page we sometimes need information or attributes of the product, that's why the relationship is made and included in the response of the order api endpoint.


resources/views/cart/queries/estimate-shipping-methods.graphql
royduin, line 1: We're using 4x space indenting.

resources/views/cart/sidebar.blade.php
```diff
@@ -11,20 +11,24 @@
         <dt>@{{ name }}</dt>
         <dd>@{{ value | price }}</dd>
     </div>
-    <div v-if="cart.shipping_addresses.length && cart.shipping_addresses[0].selected_shipping_method">
```
royduin, line 14: `cart.shipping_addresses[0].selected_shipping_method` should stay the primary value I think; `cart.shipping_method` as fallback?

PR #542, opened 2024-08-15
[2.x] Fix casts in quote model

Jade-GG:
Not sure what happened here but these were undefined classes, and you'd get an error when trying to grab the child items or crosssells. Not that we really use this model anymore, so I guess nobody noticed...


royduin:
Thanks! cc @indykoning, seems some left over stuff from the GraphQL cart migration.  A review would be nice, despite it's already merged.

PR #544, opened 2024-08-16
[2.x] Don't always assume api call returns json

Jade-GG:
For some API calls, like the `DELETE` call on a wishlist in the multiple-wishlists package, no data is returned. Of course we could also add some return data to the package there, but we might as well fix this to not be an issue.

The code for this isn't great---it's a little frustrating that there doesn't seem to be a cleaner reliable way to test for json-ness of a response (we could check for the response header but that's not bulletproof).

### DIFF
--- a/resources/js/fetch.js
+++ b/resources/js/fetch.js
@@ -62,8 +62,7 @@ export const rapidezAPI = (window.rapidezAPI = async (method, endpoint, data = {
     let responseData = await response.text()
 
     try {
-        let json = JSON.parse(responseData)
-        return json
+        return JSON.parse(responseData)
     } catch (e) {
         return responseData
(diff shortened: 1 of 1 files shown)


resources/js/fetch.js
```diff
@@ -59,7 +59,14 @@ export const rapidezAPI = (window.rapidezAPI = async (method, endpoint, data = {
         throw new FetchError(window.config.translations.errors.wrong, response)
     }
 
-    return await response.json()
+    let responseData = await response.text()
+
+    try {
+        let json = JSON.parse(responseData)
+        return json
```
royduin, line 66:
```suggestion
        return JSON.parse(responseData)
```

PR #545, opened 2024-08-16
[2.x] Allow string IDs in ES

Jade-GG:


royduin:
Why? 😇

Jade-GG:
> Why? 😇 

some IDs, like ones you'll find coming from Statamic, are UUIDs that aren't just numbers. ES has no problem indexing that but this code currently does :)

PR #553, opened 2024-08-30
Redirect when the order can not be fetched from the success page

royduin:
So when you're on the success page and the order can not be found there is nothing to show so we can redirect away.

Normally you won't get here as the success page is wrapped by the checkout component, but for example from payment providers the page is rendered standalone, see:

- https://github.com/rapidez/paynl/blob/master/resources/views/success.blade.php
- https://github.com/rapidez/core/blob/master/resources/views/checkout/steps/success.blade.php

cc https://github.com/rapidez/core/pull/503, where there is already taken care of this. Correct @indykoning?

Internal reference: FSM2-1401 and HDB-792

### DIFF
--- a/resources/js/components/Checkout/CheckoutSuccess.vue
+++ b/resources/js/components/Checkout/CheckoutSuccess.vue
@@ -47,7 +47,7 @@ export default {
                 {
                     headers: { Authorization: 'Bearer ' + (this.token || this.mask) },
                 },
-            )
+            ).catch((error) => Turbo.visit(window.url('/')))
         },
 
         serialize(address) {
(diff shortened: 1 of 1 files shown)


royduin:
Aaaah, it was Prettier haha. Tried to change it back where it went wrong 😅

Jade-GG (approved):

indykoning:
> cc #503, where there is already taken care of this. Correct @indykoning?
Correct, since #503 the success page is ALWAYS a new page no matter the payment method.

indykoning (approved):

PR #554, opened 2024-09-03
[2.x] Slideover header button fix

Jade-GG:
The back arrow button used `id` instead of `for` and used a weird attributes merge (that didn't include the class in the merge, which broke it). There may have been a use case for it while developing this slideover, but it sure as heck isn't used now.


Jade-GG:
> So we don't want any attribute merging?

Not on that label, no. This header is used via an `@include` in the index, where `$attributes` is actually used one DOM level up for the styling of the whole slideover.

PR #559, opened 2024-09-06
[2.x] Filter out double slashes

Jade-GG:
Sometimes, when using the new fetch.js functionality, a doubled `/` will slip through and break some requests. We can filter these out before it happens, just gotta make sure we don't filter out the `://` at the start of a url 😉

### DIFF
--- a/resources/js/filters.js
+++ b/resources/js/filters.js
@@ -16,7 +16,7 @@ Vue.filter('price', function (value) {
 window.url = function (path = '') {
     // Replace any double slashes not preceded by a : with a single slash
     path = path.replaceAll(/([^:])\/\//g, '$1/')
-    
+
     // Transform urls starting with / into url with domain
     if (!path.startsWith('/')) {
         return path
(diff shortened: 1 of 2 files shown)


resources/js/filters.js
```diff
@@ -15,11 +15,12 @@ Vue.filter('price', function (value) {
...
     // Transform urls starting with / into url with domain
-    if (!path.startsWith('/')) {
-        return path
+    if (path.startsWith('/')) {
+        path = (window.config.base_url || window.origin) + path
     }
 
-    return (window.config.base_url || window.origin) + path
+    // Replace any double slashes not preceded by a : with a single slash
+    return path.replaceAll(/([^:])\/\//g, '$1/')
```
indykoning, line 23: We could still pass in a relative url.
This could be problematic when passing in external urls (which this function is made to accept)
as https://www.ietf.org/rfc/rfc2396.txt defines the text between `/` are completely optional, thus it's up to the server wether `https://example.com/test//page` is the same page as `https://example.com/test/page`

I think it's best to resolve this problem where it occurs so we don't break any external routes.

Jade-GG:
@indykoning I've made it filter these slashes away in the fetch functions instead, so that the endpoint can never start with a `/` there 🙂

indykoning (approved):

PR #563, opened 2024-09-06
[2.x] Add new exception for visibility in the flat table

Jade-GG:
`visibility` does not come with its very own `visibility_value` column. As it seems there's already a list of exceptions for this scenario I've just added the visibility type to that.

### DIFF
--- a/resources/views/cart/queries/cart.graphql
+++ b/resources/views/cart/queries/cart.graphql
@@ -14,6 +14,7 @@ items {
             'type_id',
             'url_key',
             'url_suffix',
+            'visibility',
         ])->merge(config('rapidez.frontend.cart_attributes'))->unique() as $attribute)
             {{ $attribute }}
         @endforeach
--- a/src/Models/Scopes/Product/WithProductAttributesScope.php
+++ b/src/Models/Scopes/Product/WithProductAttributesScope.php
@@ -37,6 +37,7 @@ class WithProductAttributesScope implements Scope
                 if ($attribute->input == 'select' && ! in_array($attribute->source_model, [
                     'Magento\Tax\Model\TaxClass\Source\Product',
                     'Magento\Eav\Model\Entity\Attribute\Source\Boolean',
+                    'Magento\Catalog\Model\Product\Visibility',
                 ])) {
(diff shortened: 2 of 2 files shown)


royduin:
But visibility is selected by default, see: https://github.com/rapidez/core/blob/a801c682559aa06fc75227d1152b767d21bb7fb8/src/Models/Scopes/Product/WithProductAttributesScope.php#L27

Jade-GG:
> But visibility is selected by default, see:
> 
> https://github.com/rapidez/core/blob/a801c682559aa06fc75227d1152b767d21bb7fb8/src/Models/Scopes/Product/WithProductAttributesScope.php#L27

This is true, the issue comes from trying to add it as a cart attribute (after adding it to the gql schema), which then also adds it as a select automatically.

Given the fact that it does already get selected automatically here, maybe we should also add it in the default cart query? (which also requires adding it into the schema in compadre)

royduin:
And where do you need the `visibility` for?

Jade-GG:
> And where do you need the `visibility` for?

We use this in one project with a specific use case to determine how to link back to said product (whether or not it's a sub-product is based on whether or not it's visible). For now I can overwrite `cart.graphql` and manually put it in but surely that shouldn't be necessary.

royduin:
So what's the follow-up on this?

Jade-GG:
We could just add visibility as a default gql variable, if that's an ok alternative.

royduin:
And v3?

royduin:
So I saw: https://github.com/rapidez/magento2-compadre/pull/10, but this requires Compadre, we don't want that. It should stay optional.

Jade-GG:
> So I saw: [rapidez/magento2-compadre#10](https://github.com/rapidez/magento2-compadre/pull/10), but this requires Compadre, we don't want that. It should stay optional.

I do want to put a compadre version check around it to make sure, but I don't know what the version number for compadre would be 😅

Could also just not put this in the core and keep it as a custom thing in projects that need it, it's only got one pretty specific use case for now.

royduin:
Just override the query for now, if we need it again we can think this over. Thanks!

PR #564, opened 2024-09-09
[2.x] Cart summary refactor

Jade-GG:
The idea here is to use segments much like we currently have in the checkout (which, I assume, will no longer exist in Rapidez 3.x).

For now as a draft PR because I'm not quite sure yet how to set and structure the data properly. How it works right now:
- You have a `value_including_tax` and `value_excluding_tax` value that you can choose between, and a `display_tax` boolean that tells you what the magento setting is for that value.
- For some segments it doesn't make sense to have an "including tax" or "excluding tax" value, so there it's just `value` right now, but I don't know if I like that approach. I could just set both including and excluding tax values to be the same number, but that feels worse.


royduin:
Looking good to me, what do you thing @indykoning?

There is a followup on the `showTax` shop wide implementation?

royduin:
@indykoning?

indykoning (approved):

PR #566, opened 2024-09-10
Make Graphql and GraphqlMutation errorCallback the same

indykoning:
In all different contexts. checkResponseForExpiredCart would be sent different types of data.
I've pulled the arguments of both callback functions of the graphql and graphqlMutation component inline.
Also caught an error where the graphql component's callback component should return variables instead of data to be in-line with graphqlMutation


royduin:
Is it backwards compatible? Is this something that's overwritten in projects?

indykoning:
Only when calling this function directly is it not backwards compatible.
In projects (and most packages) the errorCallback had been used to execute this function. But this couldn't work in the old situation.
The errorCallback on the Graphql component has only been added since V2 and thus is not used for other things yet.

PR #567, opened 2024-09-10
Load Magento urls from website/default base_url if enabled

indykoning:
This allows custom magento backend urls
Especially useful for multistore multidomain setups.
Where you may want to obfuscate that different sites belong to the same Magento backend.

The config system is also updated to be functionally compatible with default Magento config functions.
Now requesting default config will give you only config from default
Requesting website config will give you config from the website, and falls back to default
Requesting store config will give you the config from the store, falling back to the website, and then default.


royduin:
Looking good! 3 things

1) Is it fully backwards compatible?
2) We need some docs on this
3) See internal RAP-710, maybe something to combine with this? Or first merge this and refactor later?

royduin:
@indykoning 😇

indykoning:
The config method is fully backwards compatible, all original functions still exist. 
`@config` and `Rapidez::config` will not be deprecated. The new config method is for more advanced usage like getting the urls from Website level and up, instead of store level and up.

docs: https://github.com/rapidez/docs/pull/51

Perhaps we could add a config setting to define which caching driver is used for these caches, wether it's the array driver, redis, default.

config/rapidez/system.php
```diff
@@ -10,6 +10,9 @@
     // Elasticsearch prefix.
     'es_prefix' => env('ELASTICSEARCH_PREFIX', 'rapidez'),
 
+    // Get Magento url from Database
+    'get_magento_url_from_db' => env('GET_MAGENTO_URL_FROM_DATABASE', false),
```
royduin, line 14: Maybe drop the "get"?
```suggestion
    'magento_url_from_db' => env('MAGENTO_URL_FROM_DATABASE', false),
```

And add it to the `.env.example`? + some more information in the comment / docs on this would be nice

composer.json
```diff
@@ -26,6 +26,7 @@
         "lcobucci/clock": "^2.0|^3.0",
         "lcobucci/jwt": "^4.0|^5.0",
         "rapidez/blade-directives": "~0.1",
+        "spatie/once": "*",
```
royduin, line 29: Keep this vs dropping Laravel 10 and use it from Laravel itself?

src/Models/OptionValue.php
```diff
@@ -14,17 +14,13 @@ public static function getCachedByOptionId(int $optionId): string
...
-                return html_entity_decode(self::where('option_id', $optionId)
-                    ->whereIn('store_id', [config('rapidez.store'), 0])
-                    ->orderByDesc('store_id')
-                    ->first('value')
-                    ->value ?? false);
-            });
-
-            config(['cache.app.' . $cacheKey => $optionValue]);
-        }
+        $optionValue = once(fn () => Cache::rememberForever($cacheKey, function () use ($optionId) {
```
royduin, line 17: Does this still makes sense as we're caching a lot of keys for every option id within Redis? Maybe it's faster to combine everything in one key and filter on that?

src/Models/Store.php
```diff
@@ -27,20 +27,17 @@ protected static function booting()
...
-                    'store.code',
-                    'store.website_id',
-                    'store.group_id',
-                    'store_group.root_category_id',
-                    'store_website.code AS website_code',
-                ])->get()->keyBy('store_id')->toArray();
-            });
-            config(['cache.app.stores' => $stores]);
-        }
+        $stores = once(fn () => Cache::rememberForever('stores', function () {
```
royduin, line 30: This feels a bit odd, maybe something like https://github.com/antriver/laravel-multi-cache can help out? Or maybe Laravel supports it already someway? `Cache::store(['array', 'redis'])->...`

indykoning: Since we are dependent on a PR being merged for it to work properly i've created a split PR https://github.com/rapidez/core/pull/583

src/Models/OptionValue.php
indykoning: You're right, i've moved it into a single cache

src/Models/Store.php
indykoning: This has been resolved by maintaining our own version. The split PR has been merged into this branch.

src/Models/Config.php
```diff
@@ -3,44 +3,123 @@
...
+}
 
 class Config extends Model
 {
     protected $table = 'core_config_data';
 
     protected $primaryKey = 'config_id';
 
+    protected $configCache = [];
+
```
royduin, line 25: We're not using this anymore right?

indykoning: That's right, it's only using the cache and local variables now.

config/rapidez/system.php
```diff
@@ -10,6 +10,9 @@
     // Elasticsearch prefix.
     'es_prefix' => env('ELASTICSEARCH_PREFIX', 'rapidez'),
 
+    // Get Magento url from Database
+    'magento_url_from_db' => env('GET_MAGENTO_URL_FROM_DATABASE', false),
```
royduin, line 14: Pros/cons of having this true/false by default?

royduin:
What about the Rapidez url? Should we set that from a Magento config? Could be useful for multisites. Currently multisites do work; as with the `url()` helper the request will be used to determinate the base, but that's not working within the console; `config('app.url')` != `url('/')`. For those situations the root will be forced with code like this:
```
Event::listen('rapidez:store-set', function ($store) {
    $baseUrl = config('rapidez.url.' . $store['code']);
    if ($baseUrl) {
        URL::forceRootUrl($baseUrl);
    }
});
```
Reference: https://github.com/rapidez/core/pull/398. Maybe we should have `URL::forceRootUrl()` somewhere in the core?

config/rapidez/system.php
indykoning: Pro for false:
- no breaking changes by default, only when you opt into this functionality will it behave differently 
- after upgrading you have the time to first get your database in order before enabling this feature.
Pro for true:
- You get a much better experience with url handling out of the box since you no longer need to define the urls in Rapidez itself with subfolders working

If your database has the correct urls by default, and you don't have them set using env variables or config.php.
But this is an assumption we'd have to make when defaulting it to true

indykoning:
> What about the Rapidez url? Should we set that from a Magento config? Could be useful for multisites. Currently multisites do work; as with the `url()` helper the request will be used to determinate the base, but that's not working within the console; `config('app.url')` != `url('/')`. For those situations the root will be forced with code like this:
> 
> ```
> Event::listen('rapidez:store-set', function ($store) {
>     $baseUrl = config('rapidez.url.' . $store['code']);
>     if ($baseUrl) {
>         URL::forceRootUrl($baseUrl);
>         config()->set('frontend.base_url', url('/'));
>     }
> });
> ```
> 
> Reference: #398. Maybe we should have `URL::forceRootUrl()` somewhere in the core?
> 
> Having a forced url could be useful for: [rapidez/blade-directives#18](https://github.com/rapidez/blade-directives/pull/18)

I've added this as well so as long as your store url is not the same as the website url then it will use forceRootUrl to set the url.

indykoning:
PS, since 2.x is no longer on master i've updated the branch. When we're ready i'll cherry-pick the relevant commits and create a 2.x PR

config/rapidez/system.php
royduin: Thanks! Just one thing left; documenting this option here: https://docs.rapidez.io/3.x/configuration.html#base-url

PR #568, opened 2024-09-16
Multiply fixed product tax by item quantity

Jade-GG:
The graphql endpoint only returns the fixed product tax amount for a single quantity regardless of the actual cart quantity.

Also note that---while you're able to set fixed product taxes with up to 4 decimals of precision---Magento will automatically round that to whatever currency you use before doing any calculations with it. As a result we actually do not lose any precision by doing it this way.


royduin:

indykoning (approved):

PR #573, opened 2024-09-18
PHP 8.3 support and Magento 2.4.7 tests

royduin:


royduin:
Dusk is still failing but I can't reproduce it locally. I also tried to add MariaDB support, but that takes more time so for later: RAP-868. We can keep these first steps in. @indykoning, follow-up: RAP-869.

PR #575, opened 2024-09-19
[2.x] Add resizedpath helper function

Jade-GG:
We currently use a replace with the media url, however that does not work on multisite at all, or in cases where the media url isn't exactly the same as what magento returns from graphql.

This solution catches every use case I know of so far. Placeholder image, product images, custom media, and a fallback to the product url.

### DIFF
--- a/resources/js/package.js
+++ b/resources/js/package.js
@@ -110,7 +110,11 @@ function init() {
                     this.scrollLock = bool
                 }
             },
-            resizedPath(imagePath, size, store) {
+            resizedPath(imagePath, size, store = null) {
+                if(!store) {
+                    store = window.config.store
+                }
+
                 let mediaPosition = imagePath.indexOf('/media/')
                 if (mediaPosition > 0) {
                     return `/storage/${store}/resizes/${size}/magento` + imagePath.substr(mediaPosition + 6)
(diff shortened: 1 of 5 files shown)


royduin:
Do you have examples when it's not working?

Jade-GG:
> Do you have examples when it's not working?

Any non-default store on a multistore site will return an image url that's not prefixed by the default media url, causing this replace to not work and all cart images to be broken on those sites.

BobWez98:
I can confirm this. With multisite the media url is not always included in the url magento returns in graphql. I can show you this localy.

royduin:
@indykoning 😇

indykoning (approved):

resources/js/package.js
```diff
@@ -110,6 +110,23 @@ function init() {
...
                 }
             },
+            resizedPath(imagePath, size, store = null) {
+                if (!store) {
+                    store = window.config.store
+                }
+
+                let mediaPosition = imagePath.indexOf('/media/')
+                if (mediaPosition > 0) {
+                    return `/storage/${store}/resizes/${size}/magento` + imagePath.substr(mediaPosition + 6)
```
royduin, line 120: This seems deprecated? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr, and it's unclear what we're stripping. What are those 6 characters?

resources/js/package.js
```diff
@@ -110,6 +110,23 @@ function init() {
...
+                }
+
+                let productPosition = imagePath.indexOf('/product/')
+                if (productPosition > 0) {
+                    return `/storage/${store}/resizes/${size}/magento/catalog` + imagePath.substr(productPosition)
```
royduin, line 125: Substr without a start?

BobWez98:
Follow up: https://github.com/rapidez/core/pull/602

PR #580, opened 2024-09-23
Colors changes

JimmyHoenderdaal:
Changes & adds some colors that we use regularly

### DIFF
--- a/resources/views/components/button/primary.blade.php
+++ b/resources/views/components/button/primary.blade.php
@@ -1,3 +1,3 @@
-<x-rapidez::button.base {{ $attributes->class('bg-primary text-primary-text') }}>
+<x-rapidez::button.base {{ $attributes->class('bg-primary text-white') }}>
     {{ $slot }}
 </x-rapidez::button.base>
--- a/tailwind.config.js
+++ b/tailwind.config.js
@@ -17,21 +17,19 @@ module.exports = {
     theme: {
         extend: {
             colors: {
-                primary: {
-                    DEFAULT: 'rgb(var(--primary) / <alpha-value>)', // Theme color
-                    text: 'rgb(var(--primary-text) / <alpha-value>)', // Text color that goes onto primary color
-                },
-                secondary: {
-                    DEFAULT: 'rgb(var(--secondary) / <alpha-value>)', // Conversion color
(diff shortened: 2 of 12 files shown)


tailwind.config.js
```diff
@@ -17,21 +17,19 @@ module.exports = {
...
-                inactive: 'rgb(var(--inactive) / <alpha-value>)', // Inactive text color
-                highlight: 'rgb(var(--highlight) / <alpha-value>)', // Background highlight color
-                border: 'rgb(var(--border) / <alpha-value>)', // Border color
+                enhanced: 'rgb(var(--enhanced) / <alpha-value>)', // Enhanced color
+                disabled: 'rgb(var(--disabled) / <alpha-value>)', // Disabled color
+                inactive: {
+                    DEFAULT: 'rgb(var(--inactive) / <alpha-value>)', // Inactive text color
+                    100: 'rgb(var(--inactive-100) / <alpha-value>)', // Background color
+                    110: 'rgb(var(--inactive-110) / <alpha-value>)', // Border color
+                },
```
royduin, line 29: Inactive implies it's used for inactive states but it's used for backgrounds, borders, etc

resources/views/components/button/primary.blade.php
```diff
@@ -1,3 +1,3 @@
-<x-rapidez::button.base {{ $attributes->class('bg-primary text-primary-text') }}>
+<x-rapidez::button.base {{ $attributes->class('bg-primary text-white') }}>
```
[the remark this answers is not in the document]
JimmyHoenderdaal: Yes, buttons are always custom in projects. there's not much to override either since its only classes you would add to a base.blade.php or index.blade.php

tailwind.config.js
JimmyHoenderdaal: It's because the inactive color is always some sort of gray, the color we use for backgrounds and borders is also always some sort of gray. Thats why we call them this way

royduin (changes requested):
The old situation feels more natural

JimmyHoenderdaal:
I don't really mind the border-border naming since those are used on very rare occasions since you already have the default border color set in the config.

For background colors, I like bg-base since it clearly represents the base color of components. Thus calling it base

PR #583, opened 2024-10-01
Implement multi cache

indykoning:


indykoning:
I will merge this into the load-magento-config-from-website branch since we aren't dependent on another PR anymore, so this is ready to be used.

PR #584, opened 2024-10-02
[2.x] Update frontend packages (deprecates CJS)

Jade-GG:
This is a minorly breaking change, as cjs is getting deprecated by vite 5 and removed in vite 6 (which is still in pre-release).

It seems the migration for the few cases I tried is super simple---just adding `"type": "module"` in `package.json` and updating the `postcss.config.js` file like I've done below. However, I cannot guarantee there aren't any projects that rely more on cjs that might break.


indykoning:
If we're making smaller breaking changes to the config js files anyways it might make sense to update https://github.com/rapidez/core/blob/8b7b172dddfd12effeea16187672e5b2f97b13a8/tailwind.config.js to this style as well.

It's something i've seen happen to facilitate https://vueuse.org/core/useBreakpoints/ in projects already anyways.

indykoning (approved):

Jade-GG:
> If we're making smaller breaking changes to the config js files anyways it might make sense to update https://github.com/rapidez/core/blob/8b7b172dddfd12effeea16187672e5b2f97b13a8/tailwind.config.js to this style as well.
> 
> It's something i've seen happen to facilitate https://vueuse.org/core/useBreakpoints/ in projects already anyways.

You're right, this actually needed to be changed to work and I seemingly missed it.

indykoning (approved):

Roene-JustBetter (approved):

royduin:
Note; you could update this in your project but with this PR's it's just updated for new projects.

PR #585, opened 2024-10-03
Remove unused templates, add checkout agreements

indykoning:
Because we want to leave the possibility open to use the checkout completely headless we're using graphql to get the agreements (with graphql caching on in Magento this query costs ~120ms)

However the slideover does not support being dynamically generated with unique id's in JS.
is this THE way? Or is there a better way to make this possible?


Roene-JustBetter (approved):

resources/views/components/slideover/index.blade.php
```diff
@@ -21,19 +21,47 @@
...
 
 <x-tag v-on:reset="toggleScroll(false)" :is="$tag">
-    <input id="{{ 'close-' . $id }}" class="hidden" type="reset">
+    <input
+        @if($idBind)
+            v-bind:for="{{ $closeId }}"
+        @else
+            for="{{ $closeId }}"
+        @endif
+        class="hidden" type="reset">
```
jordythevulder, line 36: is this the way to go?

JimmyHoenderdaal (approved):

Jade-GG:
This js-ability from the slideovers was originally removed [here](https://github.com/rapidez/core/commit/8d546259336aace018f041ca58293bfc323556d4), you might be better off stealing what we used to have there as it is a bit cleaner.

PR #586, opened 2024-10-04
[3.x] Improved install command

royduin:
With [Laravel Prompts](https://laravel.com/docs/11.x/prompts) we've some clean CLI prompts, so time for a improved install command with:

- Some ASCII art
- Magento Docker setup
- Validate
- Frontend build
- Packages
- Storage linking
- Indexing
- Config publishing

And some more information when a health check fails. Also the "old v2 behavior" is still present with the `--frontendonly` param.

We *could* fix some of the health check errors, but I think it's good to keep that a manual action so we're not touching the users Magento installation.

I also updated the v3 install docs: https://docs.rapidez.io/3.x/installation.html#install-rapidez

One thing I don't like about install commands is when I don't know what options I've, what the consequences are or what's it's going to do. I tried to address those things.

Thoughts? Something missing or any additional stuff useful to add?

### DIFF
--- a/src/Commands/InstallCommand.php
+++ b/src/Commands/InstallCommand.php
@@ -31,9 +31,9 @@ class InstallCommand extends Command
         $this
             ->intro()
             ->magento()
+            ->packages()
             ->validate()
             ->frontend()
-            ->packages()
             ->finalize();
     }
 
@@ -49,8 +49,8 @@ class InstallCommand extends Command
         $this->info('  Welcome to the Rapidez installer!');
         $this->info(str_repeat('=', 37));
         $this->newLine();
-        $this->warn('> You will get some questions, if you do not know what to choose;');
-        $this->warn('> Just enter through, the defaults are a good start.');
+        $this->warn('> You will be given some prompts to customize your Rapidez installation.');
+        $this->warn('> If you\'re unsure, stick to the given default options.');
         $this->newLine(2);
 
         return $this;
@@ -58,26 +58,30 @@ class InstallCommand extends Command
 
     protected function magento()
     {
-        if (! confirm('Do you have a Magento installation up-and-running?')) {
-            if (confirm(
-                label: 'Shall we setup a demo Magento installation for you in Docker?',
-                hint: 'This will run `docker-compose up -d` and takes a while, make sure Docker is running!'
-            )) {
-                $this->dockerInstall = true;
-                if (passthru('docker-compose up -d') === false) {
-                    $this->newLine();
-                    $this->error('Something went wrong, please check the errors and try again');
-                    exit;
-                }
-                if (passthru('docker exec rapidez_magento magerun2 indexer:reindex') === false) {
-                    $this->newLine();
-                    $this->error('Something went wrong, please check the errors and try again');
-                    exit;
-                }
-            } else {
-                $this->error('Please setup a Magento installation first!');
-                exit;
-            }
+        if (confirm('Do you have a Magento installation up-and-running?')) {
+            return $this;
+        }
+
+        if (!confirm(
+            label: 'Shall we setup a demo Magento installation for you in Docker?',
+            hint: 'This will run `docker compose up -d` and takes a while, make sure Docker is running!'
+        )) {
+            $this->error('Please setup a Magento installation first!');
+            exit;
+        }
+
+        $this->dockerInstall = true;
+
+        if (passthru('docker compose up -d') === false) {
+            $this->newLine();
+            $this->error('Something went wrong, please check the errors and try again');
+            exit;
+        }
+
+        if (passthru('docker exec rapidez_magento magerun2 indexer:reindex') === false) {
+            $this->newLine();
+            $this->error('Something went wrong, please check the errors and try again');
+            exit;
         }
 
         return $this;
@@ -222,7 +226,7 @@ class InstallCommand extends Command
             ->map(fn ($package) => 'rapidez/' . $package);
 
         if ($this->selectedPackages->isEmpty()) {
-            $this->warn('Ok, no whipped cream with stroopwaffle pieces on your ice cream for you!');
+            $this->warn('Ok, no whipped cream with stroopwafel pieces on your ice cream for you!');
             $this->newline(2);
 
             return $this;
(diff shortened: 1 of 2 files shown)


indykoning:
Perhaps we want to validate after installing the packages, since packages may come with their own requirements https://github.com/rapidez/paynl/pull/21/files#diff-372b727c0d9abc808461b82d701bcd8df3609a9396e050bd5f21ad6515d05f05
(We could even extend the health check in packages to check wether their required Magento plugin is installed in Magento like we currently do for Confira)

src/Commands/InstallCommand.php
```diff
@@ -3,22 +3,277 @@
...
+
+    protected function magento()
+    {
+        if (! confirm('Do you have a Magento installation up-and-running?')) {
+            if (confirm(
+                label: 'Shall we setup a demo Magento installation for you in Docker?',
+                hint: 'This will run `docker-compose up -d` and takes a while, make sure Docker is running!'
+            )) {
+                $this->dockerInstall = true;
+                if (passthru('docker-compose up -d') === false) {
```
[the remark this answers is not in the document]
BobWez98: Also, with my version of docker i think it depends on your system, i think it's  `docker compose up -d`  on linux.

BobWez98 (dismissed):

src/Commands/InstallCommand.php
```diff
@@ -3,22 +3,277 @@
...
+        $this->info(" |  _ \ __ _ _ __ (_) __| | ___ ____");
+        $this->info(" | |_) / _` | '_ \| |/ _` |/ _ \_  /");
+        $this->info(' |  _ < (_| | |_) | | (_| |  __// / ');
+        $this->info(" |_| \_\__,_| .__/|_|\__,_|\___/___|");
+        $this->info('            |_|                     ');
+        $this->info('  Welcome to the Rapidez installer!');
+        $this->info(str_repeat('=', 37));
+        $this->newLine();
+        $this->warn('> You will get some questions, if you do not know what to choose;');
+        $this->warn('> Just enter through, the defaults are a good start.');
```
Jade-GG, line 53: Would personally write this more like:

```
You will be given some prompts to customize your Rapidez installation.
If you're unsure, stick to the given default options.
```

src/Commands/InstallCommand.php
```diff
@@ -3,22 +3,277 @@
...
+            $packages[] = 'msi';
+        }
+
+        $this->selectedPackages = collect($packages)
+            ->flatten()
+            ->filter()
+            ->map(fn ($package) => 'rapidez/' . $package);
+
+        if ($this->selectedPackages->isEmpty()) {
+            $this->warn('Ok, no whipped cream with stroopwaffle pieces on your ice cream for you!');
```
Jade-GG, line 225: Even in English you still spell it `stroopwafel` ;)

src/Commands/InstallCommand.php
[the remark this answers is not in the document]
indykoning: > Also, with my version of docker i think it depends on your system, i think it's `docker compose up -d` on linux.

the `docker compose up -d` syntax is the more recent way, when it's installed as a docker plugin 
https://docs.docker.com/compose/install/linux/
if it's installed standalone it'll be docker-compose
https://docs.docker.com/compose/install/standalone/
> Note that Compose standalone uses the `-compose` syntax instead of the current standard syntax `compose`.
For example type `docker-compose up` when using Compose standalone, instead of `docker compose up`.

src/Listeners/Healthcheck/MagentoSettingsHealthcheck.php
```diff
@@ -27,7 +27,10 @@ public function handle()
 
         $configModel = config('rapidez.models.config');
         if (! $configModel::getCachedByPath('catalog/frontend/flat_catalog_product', 0)) {
-            $response['messages'][] = ['type' => 'error', 'value' => __('The product flat tables are disabled!')];
+            $response['messages'][] = ['type' => 'error', 'value' => __(
+                'The product flat tables are disabled!' . PHP_EOL .
+                'Please enabled them; see: https://docs.rapidez.io/3.x/installation.html#flat-tables'
```
indykoning, line 32:
```suggestion
                'Please enable them; see: https://docs.rapidez.io/3.x/installation.html#flat-tables'
```

src/Commands/InstallCommand.php
```diff
@@ -3,22 +3,277 @@
...
+        $this->warn('> You will get some questions, if you do not know what to choose;');
+        $this->warn('> Just enter through, the defaults are a good start.');
+        $this->newLine(2);
+
+        return $this;
+    }
+
+    protected function magento()
+    {
+        if (! confirm('Do you have a Magento installation up-and-running?')) {
```
Jade-GG, line 61: We're going 3 blocks deep here, invert these if statements?

```php
if (confirm('Do you have a Magento installation up-and-running?')) {
    return $this;
}

if (!confirm(
    label: 'Shall we setup a demo Magento installation for you in Docker?',
    hint: 'This will run `docker-compose up -d` and takes a while, make sure Docker is running!'
)) {
    $this->error('Please setup a Magento installation first!');
    exit;
}

[...]
```

Jade-GG (changes requested):

src/Commands/InstallCommand.php
indykoning, line 67: We should check wether all ports we want to use are available:
https://github.com/rapidez/rapidez/blob/55a324713fc860916e9b725cf8e19b8d9f930b1a/docker-compose.yml#L13
`TRAEFIK_PORT=80`

https://github.com/rapidez/rapidez/blob/55a324713fc860916e9b725cf8e19b8d9f930b1a/docker-compose.yml#L58
`DB_PORT=3306`

And if they are not, suggest or set a different port.

See: https://github.com/rapidez/rapidez/pull/82 for having domain resolution and the TRAEFIK_PORT variable available to set out of the box

royduin: Follow-up on this with internal reference: RAP-925

royduin:
Thanks for all the feedback @indykoning, @BobWez98 and @Jade-GG

PR #587, opened 2024-10-07
Added improved active filters, also made the filter functions globally available

BobWez98:
This adds a new frontend for the active filtering. Sadly we need a Vue component for this to map out the active filtering as there's some irregularity's in there like our Category filter, price and color filter.

I also made the Vue filters globally available so we can use them outside of templates, like the `url()` method already was.

This PR uses the colors added in #580


src/Http/Controllers/CategoryController.php
```diff
@@ -10,6 +10,7 @@ public function show(int $categoryId)
         $category = $categoryModel::findOrFail($categoryId);
 
         config(['frontend.category' => $category->only('entity_id')]);
+        config(['frontend.subcategories' => $category->subcategories->mapWithKeys(fn ($item, $key) => [$item->entity_id => $item->name])]);
```
royduin, line 13: I think pluck does the same: `->pluck('name', 'entity_id')`

resources/views/listing/partials/filter/selected.blade.php
```diff
@@ -1,20 +1,32 @@
-<selected-filters :inner-class="{ button: '!hidden last:!inline-flex' }">
-    <x-rapidez::button
-        class="w-full mb-3 text-sm"
-        slot-scope="{ clearValues, selectedValues, components }"
-        v-on:click="clearValues"
+<selected-filters :inner-class="{ button: '!inline-flex' }">
```
royduin, line 1: Is that prop / class still used now we're overwriting the slot?

config/rapidez/frontend.php
```diff
@@ -21,6 +22,9 @@
         // 'material',
     ],
 
+    // The attribute of the color swatch, used in the active filtering
+    'color_attribute' => 'color',
```
royduin, line 26: How does this work? From Magento you can have multiple swatch attributes with their own (color) values.

royduin (changes requested):

resources/js/components/Listing/Filters/CustomSelected.vue
```diff
@@ -0,0 +1,60 @@
...
+                    ) {
+                        result = this.selectedValues[filterKey].value.map((selected) => selected)
+                    }
+
+                    if (filterKey === 'price') {
+                        result = price(this.selectedValues[filterKey].value[0]) + ' - ' + price(this.selectedValues[filterKey].value[1])
+                    }
+
+                    
+                    if (Object.keys(swatches.value).includes(filterKey)) {
```
BobWez98, line 41: I removed the Color config @royduin . Every swatch can be used as a "active" filter now. Even though color filters are circles with the color as background. I think the label will suffice as "active filter" ( E.G. "Blue") as most webshops do this.

resources/views/listing/partials/filter/selected.blade.php
BobWez98: nope

config/rapidez/frontend.php
BobWez98: See my comment here; https://github.com/rapidez/core/pull/587/files#diff-e28427a88f3021679abe0aa8a08b6796959c85a443f2b127192669829c0a663aR41

resources/js/components/Listing/Filters/CustomSelected.vue
```diff
@@ -0,0 +1,59 @@
...
+                            ? this.selectedValues[filterKey].value.length
+                            : !!this.selectedValues[filterKey].value)
+                    )
+                })
+                .flatMap((filterKey) => {
+                    let result
+
+                    if (
+                        Array.isArray(this.selectedValues[filterKey].value) &&
+                        !['category', 'price', ...Object.keys(swatches.value)].includes(filterKey)
```
royduin, line 32: Can't we do something with early returns or a switch case so we don't have to specify this list here?

resources/js/vue-components.js
```diff
@@ -28,3 +28,4 @@ Vue.component('listing', () => import('./components/Listing/Listing.vue'))
 Vue.component('checkout', () => import('./components/Checkout/Checkout.vue'))
 Vue.component('checkout-success', () => import('./components/Checkout/CheckoutSuccess.vue'))
 Vue.component('popup', () => import('./components/Popup.vue'))
+Vue.component('custom-selected', () => import('./components/Listing/Filters/CustomSelected.vue'))
```
royduin, line 31: Maybe rename it to something better, selected-filters-values?

resources/js/components/Listing/Filters/SelectedFiltersValues.vue
```diff
@@ -0,0 +1,56 @@
...
+        },
+    },
+
+    render() {
+        return this.$scopedSlots.default(this)
+    },
+
+    computed: {
+        activeFilters() {
+            return Object.keys(this.selectedValues)
```
BobWez98, line 18: Tried to make it as short as possible now @royduin

royduin:
@indykoning, what do you think? 😇

resources/js/components/Listing/Filters/SelectedFiltersValues.vue
```diff
@@ -0,0 +1,56 @@
...
+                    return showFilter && (Array.isArray(value) ? value.length : !!value)
+                })
+                .reduce((result, filterKey) => {
+                    let { value } = this.selectedValues[filterKey]
+
+                    if (filterKey === 'category') {
+                        return result.concat({ code: filterKey, value: config.subcategories[value] })
+                    }
+
+                    if (filterKey === 'price') {
```
indykoning, line 30: Is this only possible with a price filterkey or is this something you can choose yourself, wether you want it to be a min/max filter?

indykoning (approved):
LGTM, just one question 🙂

BobWez98:
Currently Rapidez does not support other slider type filters than price, see; https://github.com/rapidez/core/blob/master/resources/views/listing/partials/filter/price.blade.php . I'll make an internal note for this to be developed.

royduin:
Do we need this in v2 or is just v3 fine?

BobWez98:
I think v3 is fine.

PR #588, opened 2024-10-11
[3.x] Fix .prettierrc.js

Jade-GG:


royduin:
2.x also needs this right?

royduin:
Nevermind; that's within https://github.com/rapidez/core/pull/589

PR #593, opened 2024-10-15
Changed colors - alternative default suggestions

jordythevulder:
Don't merge yet, since all of these colors still need to be implemented through the core. This is a suggestion for the color naming.

I've created an alternative for [Original PR](https://github.com/rapidez/core/pull/580/files).

**Some notes:**
- Color variables are colors that aren't binded to any specific element. They can be used everywhere for everything. That's why i didn't use the the "default theme colors" binded to a specific element like text for example `theme.textColor`.
- The only variable name that is binded to a specific element is the primary-text and secondary-text since it's especially for the primary and secondary colors
- I find the name "default" more appropriate because it feels more generic for setting base colors when starting a project.
- I've reused the tailwind color naming philosophy with the 100/200/300 iterations for the shades of the colors, but i also added a little personal twist. By adding a 10/20/30 iteration which is a visual feedback that those colors are meant for borders. This also helps developers when you are working in a team to prevent them using the wrong colors because they look a like.

**Color specific:**
I am a designer as well, and when i make a design based on the clients colors i do the following:
- default / i create a dark text color which is mostly black, but then i integrate a little bit of color of the primary color of the client into the dark text giving it a subtle saturation.
- default-100 / Then to break the whitespaces within the design i create a light background color which is the default text color but i've added 80~90% pure white. so it's based on the default text color
- default-110 / You always need a border for something. I use the `default-100` light background and add about 5~10% pure black to it to create the perfect subtle border color.
- Inactive / To create hierarchy within the designs, some elements must be visually more prominent and some need to be less prominent. To do this i've reused the default text color and added about 40~50% pure white to it. This makes looks like its muted or more inactive. To correctly follow the naming logic above it actually should be default-500 or default-inactive, But i find that makes it more confusing by using 500, and default-inactive could be shortened into just inactive.
- Enhanced / I design allot of e-commerce related shops. And in my experience with multiple external CRO marketing agencies we had feedback that by adding a new custom color for anything conversion related it boosts the sales. And it actually does. So that deserved it's own variable.

By doing so everything blends together really well visually. So by naming them all "default" it bundles them and that also helps because those colors are based on the same color. That's my philosophy :)

image


Jade-GG:
I'll repeat my comment from the previous one here:

Is `default` really the best name for this set of colors? It seems weird to me to have three different things (text, background, and border) under one thing named `default`, and then to have a default of said default specifically being the text color. What makes the text color be the default of defaults?

Why don't we just have a `text`, `background`, and `border` color? I get that they're all derivatives of the same color but at least grouping them like this doesn't make sense to me.

Alternatively, I would at least prefer something like:
```js
default: {
    text: [...],
    background: [...],
    border: [...],
},
```
which makes it more clear what each color actually is.

Roene-JustBetter:
> I'll repeat my comment from the previous one here:
> 
> Is `default` really the best name for this set of colors? It seems weird to me to have three different things (text, background, and border) under one thing named `default`, and then to have a default of said default specifically being the text color. What makes the text color so special that it's earned the title of "the default of defaults"?
> 
> Why don't we just have a `text`, `background`, and `border` color? I get that they're all derivatives of the same color but at least grouping them like this doesn't make sense to me.
> 
> Alternatively, I would at least prefer something like:
> 
> ```js
> default: {
>     text: [...],
>     background: [...],
>     border: [...],
> },
> ```
> 
> which makes it more clear what each color actually is.

If we are going to use default as name I think the alternatively above is more readable. 
So I would prefer:
```js
default: {
    text: [...],
    background: [...],
    border: [...],
},
```

jordythevulder:
As i've mentioned above, I don't think it's the way to go to bind colors to specific elements. This way if you can use that exact color for another element it will be used as `bg-default-border` or `bg-default-text` which makes no sense to me. And what makes the "the default of defaults" is because the text color is the color where all the other defaults are based on.

Jade-GG:
> because the text color is the color where all the other defaults are based on

Sounds pretty opinionated to me, and I thought we wanted to avoid that in the core. Maybe *you* will base the background color on the text color, but I'm sure some people would choose to start with the background/border and then make the text from that.

We could also make it `default-900` which then follows tailwind standards in this case, if you're really against making it an actually descriptive name.

royduin:
Closing this in favor of https://github.com/rapidez/core/pull/622

PR #597, opened 2024-10-16
Fix undefined message when an error occurs when placing order

indykoning:


royduin:
This is v3 only? As it's going to master now.

indykoning:
Checkout.vue will no longer exist in V3, nor will it use REST so it's only relevant for V2
i've changed the branches

indykoning:
I'm going to recreate this branch on V2

indykoning:
See: https://github.com/rapidez/core/pull/599

PR #598, opened 2024-10-17
[3.x] Add Dutch translations

Jade-GG:
Takes all translated strings that exist inside the core (aside from the ones on the `socials` component) and provides a dutch translation file for them.


JimmyHoenderdaal (dismissed):

lang/nl.json
royduin, line 1: Did you check all these translations to be present within the core? So this file includes all translations and doesn't include any translations not present within the core?

How can we verify / make sure that new translations get added and old translations get removed? I know https://github.com/barryvdh/laravel-translation-manager can search within templates for translations., not sure if there is a standalone version which can run as a Github action to check. Came across this one: https://github.com/LarsWiegers/laravel-translations-checker but look's like that package isn't searching within templates; just checking other language files.

src/RapidezServiceProvider.php
```diff
@@ -128,6 +128,7 @@ protected function bootPublishables(): self
 
             $this->publishes([
                 __DIR__ . '/../resources/lang' => resource_path('lang/vendor/rapidez'),
+                __DIR__ . '/../lang'           => resource_path('lang'),
```
royduin, line 131: When it's already published it will not be overwritten right?

src/RapidezServiceProvider.php
```diff
@@ -128,6 +128,7 @@ protected function bootPublishables(): self
 
             $this->publishes([
                 __DIR__ . '/../resources/lang' => resource_path('lang/vendor/rapidez'),
```
royduin, line 130: Should we move this? As language files where moved with Laravel 9, see: https://laravel.com/docs/9.x/upgrade#the-lang-directory and all examples here: https://laravel.com/docs/11.x/packages#language-files refer to the lang directory directly.

royduin (changes requested):

lang/nl.json
Jade-GG: > Did you check all these translations to be present within the core? So this file includes all translations and doesn't include any translations not present within the core?

Yes, and yes. This file is directly generated by going through all `@lang(...)` and `__(...)` mentions, as well as any variables that go into these (like input labels).

> How can we verify / make sure that new translations get added and old translations get removed?

I'm not sure, it's tricky specifically because of this last little bit where things get translated indirectly. Keeping an eye on pull requests is of course a thing, but it's not perfect.

src/RapidezServiceProvider.php
Jade-GG: Correct, although I think this is actually not the right folder (as your other comment mentions)

src/RapidezServiceProvider.php
Jade-GG: We can, for sure. The only issue is that it makes the publish command a bit more annoying if the json file and the php files are in the same folder, as the structure you should get after publishing is different. The way that the php translations work with the namespace is different to how the json files work (for some reason) so you get a discrepancy like this:

```
(what it looks like in the core)
/lang
    /en
        frontend.php
    /nl
        frontend.php
    nl.json
```

```
(in project after publishing, should look like this)
/lang
    /vendor
        /rapidez
            /en
                frontend.php
            /nl
                frontend.php
    nl.json
```

lang/nl.json
royduin: Could you organise a followup on this?

Jade-GG: I've made an internal issue RAP-950 for this :)

lang/vendor/rapidez/nl/frontend.php
```diff
@@ -0,0 +1,38 @@
...
+        'email'             => 'Een email adres is verplicht.',
+    ],
+
+    'asc'       => 'Oplopend',
+    'desc'      => 'Aflopend',
+    'relevance' => 'Relevantie',
+    'newest'    => 'Nieuwste',
+    'all'       => 'Alles',
+
+    // 'sorting' => [
```
JimmyHoenderdaal, line 32: maybe add
    'sorting' => [
        'price' => [
            'asc' => 'Prijs oplopend',
            'desc' => 'Prijs aflopend',
        ],
        'name' => [
            'asc' => 'Naam A-Z',
            'desc' => 'Naam Z-A',
        ],
    ],
    
    since they dont get translated this way

jordythevulder (dismissed):

indykoning (approved):
Just to make sure, if you add an nl.json to your project. Will those translations overwrite these?

Jade-GG:
> Just to make sure, if you add an nl.json to your project. Will those translations overwrite these?

Yep! I made sure to check that :)

PR #600, opened 2024-10-17
Automatic font preload links in the head

royduin:
The package just grabs all the woff2 fonts discovered by Vite and generates preload links for them:
```html
<link rel='preload' href='https://website.com/build/assets/font.woff2' as='font' type='font/woff2' crossorigin>
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,13 +18,14 @@
     "require": {
         "php": "^8.1|^8.2|^8.3",
         "blade-ui-kit/blade-heroicons": "^2.4",
-        "mailerlite/laravel-elasticsearch": "^11.1",
         "illuminate/database": "^10.0|^11.0",
         "illuminate/events": "^10.0|^11.0",
         "illuminate/queue": "^10.0|^11.0",
         "illuminate/support": "^10.0|^11.0",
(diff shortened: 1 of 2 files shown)


royduin:
Doesn't seem to play nice with Statamic's static caching when used in combination with https://github.com/rapidez/statamic

royduin:
So the issue was Nginx with:
```
upstream sent too big header while reading response header from upstream
```
But that's because in the specific project there are 28 fonts within the manifest and the size limit is not inline with the Nginx defaults: https://github.com/justbetter/laravel-http3earlyhints/blob/1fc4c118ab792f5bf5ab3ec81eb27625857a38ba/src/config.php#L13

When you've a multisite like: https://docs.rapidez.io/3.x/theming.html#multistore, all fonts are in the manifest. So the fonts of all your website are added as a preload.

Should we split and have a manifest per website? Or should we check the generated CSS file if the fonts are actually in the specific CSS file?

royduin:
Related:
- https://github.com/Log1x/laravel-webfonts/issues/38
- https://github.com/vitejs/vite/issues/18392

Jade-GG:
Closing, as by now there are other solutions that we're going to look into with RAP-979

PR #601, opened 2024-10-21
[3.x] Use slot scopes without destructuring

Jade-GG:
The destructuring made it harder to add functionality to vue components, as it meant you also had to overwrite the view containing the `slot-scope` just because we defined every individual thing unnecessarily.

This is of course an extremely breaking change, and should also be done in other packages. Just making sure we actually want this before I do that though 😉

### DIFF
--- a/resources/views/layouts/partials/header/autocomplete.blade.php
+++ b/resources/views/layouts/partials/header/autocomplete.blade.php
@@ -42,7 +42,7 @@
             >
                 <div
                     class="{{ config('rapidez.frontend.z-indexes.header-dropdowns') }} absolute -inset-x-10 top-full max-h-[600px] overflow-auto rounded-b-xl border bg-white p-2 md:p-5 shadow-xl md:inset-x-0 md:w-full md:-translate-y-px"
-                    v-if="dataSearchScope.downshiftProps.isOpen && (dataSearchScope.suggestions.length || autocompleteScope.resultsCount)"
+                    v-if="dataSearchScope.downshiftProps.isOpen && (dataSearchScope.data.length || autocompleteScope.resultsCount)"
                 >
                     <template v-for="(resultsData, resultsType) in autocompleteScope.results ?? {}" v-if="resultsData?.hits?.length">
                         @foreach (config('rapidez.frontend.autocomplete.additionals') as $key => $fields)
--- a/resources/views/layouts/partials/header/autocomplete.blade.php
+++ b/resources/views/layouts/partials/header/autocomplete.blade.php
@@ -18,7 +18,7 @@
     class="w-full"
     v-cloak
 >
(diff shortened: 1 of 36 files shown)


indykoning (dismissed):

royduin:
Do we really want this everywhere? I'm not sure if this is better. Thoughts @indykoning and @BobWez98?

Jade-GG:
> Do we really want this everywhere? I'm not sure if this is better. Thoughts @indykoning and @BobWez98?

What seems nice to me here is that it's much easier to keep variable scopes clear. When you see just a variable named `data` in a template you might not always know what it's from. Plus, you don't have to worry about variable name clashing in this type of setup :)

There's definitely some places where it's overkill though, I've done them all in this PR for consistency. If we want to break from the consistency to keep some templates cleaner, then I think we should probably define some sort of rule when to use destructuring and when not to use it.

BobWez98:
I think this is nice whenever there's a lot of variables. Like here; https://github.com/rapidez/core/pull/601/files#diff-761aeeeb279b8890378bf9eb7f9956bba2d671c1e71e2175de4b768e0ca28b7aL9 .
Whenever there's only one variable used i don't see why we should do this.

royduin:
With V3 (just merged in master) there are some changes (that's why there are so many conflicts no) and this is used on some places. It's also possible to rename data to something else like here: https://github.com/rapidez/checkout-theme/pull/118/files#diff-e51d8c1159c452b5ecbf3b7be1374b69bf4944493c994e48db5270f11e6b2dd6R25

Where do we like it and where don't we like it?

royduin:
Looking forward to a follow-up on this

resources/views/layouts/partials/header/autocomplete.blade.php
```diff
@@ -31,20 +31,20 @@ class="relative [&_*]:!m-0 [&_[isclearicon=]]:!mr-2 [&_.cancel-icon]:!fill-[#595
...
         >
             <div
                 slot="render"
-                slot-scope="{ downshiftProps: { isOpen }, data: suggestions }"
+                slot-scope="dataSearchScope"
             >
                 <div
                     class="{{ config('rapidez.frontend.z-indexes.header-dropdowns') }} absolute -inset-x-10 top-full max-h-[600px] overflow-auto rounded-b-xl border bg-white p-2 md:p-5 shadow-xl md:inset-x-0 md:w-full md:-translate-y-px"
-                    v-if="isOpen && (suggestions.length || resultsCount)"
+                    v-if="dataSearchScope.downshiftProps.isOpen && (dataSearchScope.suggestions.length || autocompleteScope.resultsCount)"
```
Roene-JustBetter, line 45:
```suggestion
                    v-if="dataSearchScope.downshiftProps.isOpen && (dataSearchScope.data.length || autocompleteScope.resultsCount)"
```

PR #602, opened 2024-10-21
[2.x] ResizedPath shorter helper function

BobWez98:
Refactor of #575 

I don't see why we need to guess the position of certain parts of the url like in #575 . I think we can just replace the part from magento. I also removed the (possible) url parameters magento likes to return to make sure the urls can be cached.

V3: #629

### DIFF
--- a/resources/js/package.js
+++ b/resources/js/package.js
@@ -110,7 +110,11 @@ function init() {
                     this.scrollLock = bool
                 }
             },
-            resizedPath(imagePath, size, store) {
+            resizedPath(imagePath, size, store = null) {
+                if(!store) {
+                    store = window.config.store
+                }
+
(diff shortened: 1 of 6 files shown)


resources/js/package.js
```diff
@@ -110,6 +110,16 @@ function init() {
                     this.scrollLock = bool
                 }
             },
+            resizedPath(imagePath, size, store = null) {
+                if (!store) {
+                    store = window.config.store
+                }
+
+                let url = new URL(imagePath)
+                url = (url.origin + url.pathname).replace(config.base_url + '/media', '')
```
indykoning, line 119: We can probably remove the url.origin here and skip needing to remove base_url

indykoning (approved):

PR #605, opened 2024-10-21
[3.x] Use default meta description from magento

Jade-GG:
Magento provides a config for this that we can listen to in case it's not set on the page.


indykoning (approved):

royduin:
We also need this listed in the docs: https://docs.rapidez.io/3.x/configuration.html#magento

Jade-GG:
https://github.com/rapidez/docs/pull/55

PR #609, opened 2024-10-22
[2.x] Reduce forced reflow when initialising

indykoning:
Before (the dotted line is the onload event):
image
After (as you can see the onload can be fired before the style changes):
image


royduin:
Do we also need this in v3?

indykoning:
> Do we also need this in v3?

We do, i will first attempt to merge V2 into V3 as it's still in development.
If that fails i will cherry-pick this feature

PR #610, opened 2024-10-22
[3.x] Use correct object structure in postcss config

Jade-GG:


indykoning (approved):

royduin:
What does it fix? Do you have a better description I can use for the commit message.

Jade-GG:
Yes, this is related to the vite update that changed everything to ESM modules, but this file was updated incorrectly. So for a description, something like:

Use correct object structure in postcss config

royduin:
This also needs be fixed in v2 right?

Jade-GG:
> This also needs be fixed in v2 right?

Ah you're right, I thought the branch split was from before this update. I'll make a PR.

PR #612, opened 2024-10-23
Remove unused templates, add checkout agreements

indykoning:
Because we want to leave the possibility open to use the checkout completely headless we're using graphql to get the agreements (with graphql caching on in Magento this query costs ~120ms)

However the slideover does not support being dynamically generated with unique id's in JS.
is this THE way? Or is there a better way to make this possible?

Reopened this way since #585 cannot be reopened


indykoning:
https://github.com/rapidez/core/pull/585#issuecomment-2431893343

Jade-GG:
We're looking at a nicer way to do this with internal issue RAP-919 😇

resources/views/components/slideover/partials/header.blade.php
```diff
@@ -13,7 +19,13 @@
                     {{ $title }}
                 </span>
             @endif
-            <label for="{{ $closeId }}" class="absolute right-0 top-1/2 -translate-y-1/2 cursor-pointer text-white">
+            <label
+                @if($idBind)
+                    v-bind:for="{{ $isInForm ? $closeId : $idBind }}"
+                @else
+                    for="{{ $isInForm ? $closeId : $id }}"
+                @endif
```
royduin, line 27: I don't like this repeated multiple times

royduin (commented):
Changing it to a draft for now, this also depends on a v3 version of https://github.com/rapidez/core/pull/619

resources/views/components/slideover/partials/header.blade.php
Jade-GG: I agree that these aren't great. This could be simplified by putting both values into the v-bind like I had done before with something like

```
v-bind:for="{{ $idBind ?: ('\'' . $id . '\'') }}
```

indykoning:
The new global slideover component has been implemented!

resources/views/checkout/steps/agreements.blade.php
```diff
@@ -1,38 +1,25 @@
...
-                dusk="agreements"
-                required
-            />
-            <label class="text text-sm cursor-pointer underline hover:no-underline" for="{{ $agreement->agreement_id }}_agreement">
-                {{ $agreement->checkbox_text }}
-            </label>
+<graphql query="{ checkoutAgreements { agreement_id name checkbox_text content is_html mode } }" cache="checkout.agreements" v-cloak>
+    <div slot-scope="{ data }" v-if="data">
+        <div v-for="agreement in data.checkoutAgreements" :key="agreement.agreement_id">
+            <global-slideover :title="agreement.name" :content="'<div class=&quot;p-3' + (!agreement.is_html ? ' whitespace-pre-line' : '') + '&quot;>' + agreement.content + '</div>'" v-slot="slideover">
```
royduin, line 4: We can't think of anything else for the content prop? Teleporting maybe?

royduin: Internal reference: RAP-1061

PR #613, opened 2024-10-24
[2.x] Make sure you actually log in in the checkout

Jade-GG:
Previously, you could just type in any random password in the password box and it would validate without the need to log in, allowing people to actually try placing an order on an existing account without logging in.

This fixes that bug. I'm not sure if this is also a thing in 3.x as that checkout works quite differently, so I haven't made a PR for that.

### DIFF
--- a/resources/views/checkout/steps/login.blade.php
+++ b/resources/views/checkout/steps/login.blade.php
@@ -12,17 +12,27 @@
                 v-on:input="loginInputChange"
                 required
             />
-            <x-rapidez::input
-                v-if="!emailAvailable"
-                :label="false"
-                class="mt-3"
-                name="password"
-                type="password"
-                placeholder="Password"
-                ref="password"
-                v-on:input="loginInputChange"
-                required
-            />
+            <div class="relative">
+                <x-rapidez::input
+                    v-if="!emailAvailable"
+                    :label="false"
+                    class="mt-3"
+                    name="password"
+                    type="password"
+                    placeholder="Password"
+                    ref="password"
+                    v-on:input="loginInputChange"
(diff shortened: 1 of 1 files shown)


resources/views/checkout/steps/login.blade.php
```diff
@@ -12,17 +12,27 @@
...
+                    ref="password"
+                    v-on:input="loginInputChange"
+                    required
+                />
+
+                <input
+                    type="checkbox"
+                    v-if="!emailAvailable"
+                    oninvalid="this.setCustomValidity('{{ __('Please log in') }}')"
+                    class="absolute h-full inset-0 opacity-0 pointer-events-none"
```
jordythevulder, line 32: is this the way to go? absolute inset-0 pointer events none? seems more like a quick fix rather than a proper one?

Jade-GG: We need some sort of extra input that the browser still renders, so that we can display the validation message. There isn't really a better method for that :)

jordythevulder (approved):

royduin:
I'd like to know if v3 also needs a fix for this

indykoning (approved):

Jade-GG:
> I'd like to know if v3 also needs a fix for this

From what I can see, the form will still validate incorrectly on v3: https://github.com/rapidez/core/blob/master/resources/views/checkout/steps/login.blade.php

So v3 will likely also need a fix unless I'm missing something wrapped around it that validates the login credentials.

royduin:
Could you arrange that?

royduin:
Any info?

PR #614, opened 2024-10-25
[3.x] Add store scope to product option

BobWez98:
In a project i got some results for other stores. This fixes that.

V2 #615


BobWez98:
We can use `$option->titles->firstForCurrentStore()` for this. See; https://github.com/rapidez/core/blob/master/src/Models/ProductOption.php#L33

PR #616, opened 2024-10-25
Use fallback route middlewares

royduin:
This controller invokes other controllers, but middlewares registered there in the constructor do not work. Not sure yet how Laravel handles this so with this we're checking middlewares of "child controllers" en register them again.

- Not sure it it's fully compatible with everything; is `uses` always a controller for example? Otherwise this will break
- The middlewares of "child controllers" get registered on all fallback routes, this should be scoped to the specific child


indykoning:
Unfortunately it can also be an anonymous function. We should check for that
https://docs.rapidez.io/3.x/package-development.html#fallback-routing

royduin:
2.x https://github.com/rapidez/core/pull/643
3.x https://github.com/rapidez/core/pull/617

PR #617, opened 2024-10-25
Also apply middlewares when running fallback routes

indykoning:
This PR supersedes: https://github.com/rapidez/core/pull/616

Instead of simply calling the given functions we send them through the routing pipeline as is done in Laravel itself
https://github.com/laravel/framework/blob/5527e72efe5d86cf83afe21944a47fe9a0bdc406/src/Illuminate/Routing/Router.php#L797

Which applies the middlewares, but only for the given router.
However this expects a route instead of a route action. Thus we turn it into a route.
and to maintain backwards compatibility, and exact compatibility with regular routing arguments we pass it through the createRoute function
https://github.com/laravel/framework/blob/5527e72efe5d86cf83afe21944a47fe9a0bdc406/src/Illuminate/Routing/Router.php#L554

Otherwise it would only accept `[Class, 'function']` arguments

### DIFF
--- a/src/Http/Controllers/FallbackController.php
+++ b/src/Http/Controllers/FallbackController.php
@@ -3,28 +3,42 @@
 namespace Rapidez\Core\Http\Controllers;
 
 use Exception;
+use Illuminate\Container\Container;
 use Illuminate\Database\Eloquent\ModelNotFoundException;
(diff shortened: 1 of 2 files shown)


royduin:
Could you add a bit of info?

PR #618, opened 2024-10-29
[2.x] Add wrapper hack for turbo

Jade-GG:
This fixes an annoying issue that's been inside Turbo since its inception: Turbo will always replace the *entire* `<body>` tag. As a result, this causes widgets that have been loaded into the body by scripts like GTM to be destroyed on navigation (see internal slack for current examples). Even worse the scripts will persist but their elements will not, so we get a bunch of JS errors.

Turbo doesn't support changing this root node to something else, even though this has been asked in issues since at least 2021. There are some events we can hook into with the renderer to change certain behaviors, but this is unwieldy and doesn't really work well for this use case.

This PR hacks into the PageRenderer class and replaces [the assignNewBody function](https://github.com/hotwired/turbo/blob/ea54ae5ad4b6b28cb62ccd62951352641ed08293/src/core/drive/page_renderer.js#L182) with one that does the exact same except with a `#turbo-wrapper` element (with a fallback to still use the body so it's not a big breaking change).

(Note that this PR goes to `2.x` for now, I will make a 3.x PR only if/when we're sure this is final)

### DIFF
--- a/resources/js/turbolinks.js
+++ b/resources/js/turbolinks.js
@@ -10,13 +10,13 @@ Turbo.setProgressBarDelay(5)
 // This allows things placed in the body by scripts like GTM to persist
 Object.assign(PageRenderer.prototype, {
     assignNewBody() {
-        const container = document.querySelector("#turbo-wrapper")
-        const newContainer = this.newElement.querySelector("#turbo-wrapper")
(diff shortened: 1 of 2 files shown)


royduin:
Closing this for now as it was for 1 specific use case.

PR #619, opened 2024-10-29
[2.x] Add global slideover

Jade-GG:
This adds a global slideover into the bottom of the body which can be re-used many times on a page without having to fill the DOM with many elements for every single slideover.

Usage example:

```blade
<x-rapidez::slideover.global :title="__('My slideover')">
    <x-slot:label class="rounded p-3 border">
        Click me
    </x-slot:label>
    <div class="font-semibold">
        Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
        Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.
        Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur.
        Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
    </div>
</x-rapidez::slideover.global>
```

This is built specifically with the limiting of DOM elements in mind, which ends up meaning that we put the slideover contents as escaped HTML directly into the a property of the vue component rather than actually having it on the page as a hidden element. This does mean we don't have a drop-in mobile-only version like with the standand slideover, but this can still be achieved if needed.

(this PR is built on 2.x for now to be able to test in current projects, but I will also make a 3.x PR when this is ready)

### DIFF
--- a/resources/js/components/GlobalSlideover.vue
+++ b/resources/js/components/GlobalSlideover.vue
@@ -1,0 +1,31 @@
+<script>
+export default {
+    props: {
+        title: String,
+        position: String,
+        contents: String,
+    },
+
+    render() {
+        return this.$scopedSlots.default(this)
+    },
+
+    methods: {
+        open() {
+            this.$root.globalSlideover.content = this.slideoverContents
+            this.$root.globalSlideover.title = this.title
+            this.$root.globalSlideover.position = this.position
+        },
+    },
+
+    computed: {
+        slideoverContents() {
+            if (this.contents) {
+                return this.contents
+            }
+
+            return this.$el.querySelector('.global-slideover-wrapper').innerHTML;
+        }
+    },
+}
+</script>
--- a/resources/js/components/GlobalSlideover.vue
+++ b/resources/js/components/GlobalSlideover.vue
@@ -24,8 +24,8 @@ export default {
                 return this.contents
             }
 
-            return this.$el.querySelector('.global-slideover-wrapper').innerHTML;
-        }
+            return this.$el.querySelector('.global-slideover-wrapper').innerHTML
+        },
     },
 }
 </script>
--- a/resources/js/components/GlobalSlideover.vue
+++ b/resources/js/components/GlobalSlideover.vue
@@ -12,7 +12,7 @@ export default {
 
     methods: {
         open() {
-            this.$root.$emit('global-slideover-open', {'content': this.content, 'title': this.title, 'position': this.position});
+            this.$root.$emit('global-slideover-open', { content: this.content, title: this.title, position: this.position })
         },
     },
 }
--- a/resources/js/components/GlobalSlideover.vue
+++ b/resources/js/components/GlobalSlideover.vue
@@ -3,7 +3,7 @@ export default {
     props: {
         title: String,
         position: String,
-        contents: String,
+        content: String,
     },
 
     render() {
@@ -12,19 +12,7 @@ export default {
 
     methods: {
         open() {
-            this.$root.globalSlideover.content = this.slideoverContents
-            this.$root.globalSlideover.title = this.title
-            this.$root.globalSlideover.position = this.position
-        },
-    },
-
-    computed: {
-        slideoverContents() {
-            if (this.contents) {
-                return this.contents
-            }
-
-            return this.$el.querySelector('.global-slideover-wrapper').innerHTML
+            this.$root.$emit('global-slideover-open', {'content': this.content, 'title': this.title, 'position': this.position});
         },
     },
 }
--- a/resources/views/components/slideover/global.blade.php
+++ b/resources/views/components/slideover/global.blade.php
@@ -1,19 +1,18 @@
 {{--
-    This is a wrapper for the global slideover. Use this as much as possible to reduce the amount of html elements in the DOM.
+This is a wrapper for the global slideover. Use this as much as possible to reduce the amount of html elements in the DOM.
 
-    Standard usage may look like this:
-    <x-rapidez::slideover.global :title="__('My slideover')">
-        <x-slot:label class="rounded p-3 border">
-            Click me
-        </x-slot:label>
+Standard usage may look like this:
+```
+<x-rapidez::slideover.global :title="__('My slideover')">
+    <x-slot:label class="rounded p-3 border">
+        Click me
+    </x-slot:label>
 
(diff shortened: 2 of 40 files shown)


Roene-JustBetter (approved):

jordythevulder (approved):

JimmyHoenderdaal (dismissed):

royduin:
@indykoning, what do you think?

resources/views/components/slideover/global.blade.php
```diff
@@ -0,0 +1,26 @@
...
+            Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur.
+            Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
+        </div>
+    </x-rapidez::slideover.global>
+--}}
+
+@props(['title', 'position' => 'left'])
+@slots(['label', 'contents'])
+
+<global-slideover title="{{ $title }}" position="{{ $position }}" contents="{{ $slot->toHtml() }}" v-slot="slideover">
```
indykoning, line 22: Just to be sure, this has been tested with a slot with classes and other attributes? And won't break out of the contents tag if that happens

Jade-GG: I've put a bunch of contents in without issue, the `{{ ... }}` escapes everything (as laravel calls the `e()` function).

indykoning (dismissed):

resources/js/components/GlobalSlideover.vue
```diff
@@ -0,0 +1,31 @@
...
+        },
+    },
+
+    computed: {
+        slideoverContents() {
+            if (this.contents) {
+                return this.contents
+            }
+
+            return this.$el.querySelector('.global-slideover-wrapper').innerHTML
```
indykoning, line 27: Am i correct in seeing that this element does not exist in this PR?
Also note that this is not reactive, so updates to this elements content will not update the slideover

Jade-GG: yes, this is something that's useful for when/if we implement a mobile-only version of this. In that case, we need the content to actually exist in the DOM, and thus we would have to have a queryable div.

However, as for right now, that's not implemented, so it might be better to just leave this out for now?

resources/views/components/slideover/global.blade.php
royduin, line 22: Can't we use a slot for the content? This feels/looks weird with HTML in it

resources/js/components/GlobalSlideover.vue
royduin: If we're not using it; remove it.

resources/views/layouts/partials/global-slideover.blade.php
```diff
@@ -0,0 +1,12 @@
+<template v-if="globalSlideover.content ?? false" v-cloak>
+    <x-rapidez::slideover
+        id="slideover-global"
+        v-bind:class="{ '-right-full peer-checked:right-0 !left-auto': (globalSlideover.position ?? 'left') === 'right' }"
+    >
+        <x-slot:title>
+            <div v-html="globalSlideover.title ?? ''"></div>
```
royduin, line 7: The variable does always exist right? Do we really need the `?? ''` fallback?

resources/views/components/slideover/global.blade.php
```diff
@@ -0,0 +1,26 @@
...
+            Click me
+        </x-slot:label>
+
+        <div class="font-semibold">
+            Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
+            Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.
+            Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur.
+            Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
+        </div>
+    </x-rapidez::slideover.global>
```
royduin, line 16:
````suggestion
This is a wrapper for the global slideover. Use this as much as possible to reduce the amount of html elements in the DOM.

Standard usage may look like this:
```
<x-rapidez::slideover.global :title="__('My slideover')">
    <x-slot:label class="rounded p-3 border">
        Click me
    </x-slot:label>

    <div class="font-semibold">
        Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
    </div>
</x-rapidez::slideover.global>
```
````

File: resources/views/components/slideover/global.blade.php (hunk 1 above)
Comment (Jade-GG) [reply resources/views/components/slideover/global.blade.php:21]: That's essentially what the `global-slideover-wrapper` is for. If we want to use it, we can do something like this:

```blade
@props(['title', 'position' => 'left'])
@slots(['label', 'contents'])

<global-slideover title="{{ $title }}" position="{{ $position }}" v-slot="slideover">
  <div {{ $attributes->class('contents') }}>
    <div class="hidden global-slideover-wrapper">
      {{ $slot }}
    </div>
    <label {{ $label->attributes->class('global-slideover-label cursor-pointer') }} for="slideover-global" v-on:click="slideover.open">
      {{ $label }}
    </label>
  </div>
</global-slideover>
```

However, in this case we have a pretty weird structure which isn't necessarily any better. I tried to get vue slots to work, but I wasn't able to. I could give it another go though as that would definitely be the cleanest.

Jade-GG: The issue with using Vue slots at least is that you can't really turn a `this.$scopedSlots.xxxxx()` into actual html, there is no helper function for it and it only returns something that Vue uses internally to generate the html.

So then what remains is the other option I posted above.

royduin:
@indykoning is going to check this soon. Do we need this in v2? Or is just v3 fine?

royduin:
Please double check this PR @Jade-GG and @indykoning, and resolve the conflict @Jade-GG 😇 also a 3.x PR can be made

Jade-GG:
See #672 😇

PR #620, opened 2024-10-29
[3.x] Implement PHPStan level 1

Jade-GG:


indykoning (dismissed):

royduin:
Can we skip the docblocks?

Jade-GG:
> Can we skip the docblocks?

We can't skip these without either getting rid of phpstan entirely, or having a ton of different ignores.

Jade-GG:
> Can we skip the docblocks?

Done :)

royduin:
Thanks!! 💯

PR #307, opened 2023-08-02
Improve frontend

JimmyHoenderdaal:

PR #622, opened 2024-10-30
Color names refactor

royduin:
Currently we've already 3 PR's about this:

- https://github.com/rapidez/core/pull/580
- https://github.com/rapidez/core/pull/592
- https://github.com/rapidez/core/pull/593

Hereby option 4 🚀 

The current colors:
```js
neutral: 'rgb(var(--neutral) / <alpha-value>)', // Default text color
inactive: 'rgb(var(--inactive) / <alpha-value>)', // Inactive text color
highlight: 'rgb(var(--highlight) / <alpha-value>)', // Background highlight color
border: 'rgb(var(--border) / <alpha-value>)', // Border color
```

The comments imply where they're used for, but `bg-neutral` can also be found within the source despite the `Default text color` comment. Besides `border-border` which feels weird, `inactive` implying the element is inactive but it's used to "dim" a color and `bg-highlight` which isn't actually highlighting something. Time for a refresh.

---

Tailwind isn't providing any "opinionated" or "project specific" naming; they just provide the color utilities which they recommend, which is great but if you'd like to quickly change the whole look and feel of Rapidez even within packages you've to publish all views and do some search and replaces. We recommend against that as mentioned within the docs: https://docs.rapidez.io/2.x/theming.html#views

> It's recommended to only add the views you've changed into your source control for upgradability. To keep track of what you've changed in a view it's a good idea to add the unchanged version to version control before you make any changes.

Otherwise you've to check all the templates changes every update. No problem if you're building everything from scratch but if you only like to change some parts, it's a better approach to just publish what you'd to change.

---

So next chapter; what about other frameworks? Well... Bootstrap does provide some, see the [colors](https://getbootstrap.com/docs/5.3/customize/color/), [color utilities](https://getbootstrap.com/docs/5.3/utilities/colors/) and [border colors](https://getbootstrap.com/docs/5.3/utilities/borders/#color) for example. [UIkit text colors](https://getuikit.com/docs/text#text-color) and [some others](https://github.com/troxler/awesome-css-frameworks#general-purpose). But still nothing really convincing... until: [Github's Primer](https://primer.style/)!

---

So Primer? Yeah, check out these:
- [Color overview](https://primer.style/foundations/color/overview)
- [Base scales](https://primer.style/foundations/color/base-scales)
- [Token names](https://primer.style/foundations/primitives/token-names)

The "base scales" are provided by Tailwind and cover the "unclear" color names currently used for the background, border and text. Just pick a color palette. We just have to translate those to something clear; the "[functional category](https://primer.style/foundations/color/overview#design-token-categories)" for "[neutral colors](https://primer.style/foundations/color/overview#neutral-colors)". They split that into:

- [Foreground](https://primer.style/foundations/color/overview#foreground)
- [Background](https://primer.style/foundations/color/overview#background)
- [Border](https://primer.style/foundations/color/overview#border)
- [Shadow](https://primer.style/foundations/color/overview#shadow)

And on those neutral colors there are "[color modifiers](https://primer.style/foundations/primitives/token-names#color-modifiers)":
- `default`
- `muted`
- `emphasis`

Where `emphasis` can also be found in other frameworks like Bootstrap and UIkit.

Next to those neutral colors there are "[semantic colors](https://primer.style/foundations/color/overview#semantic-colors)", we already have the colors from Tailwind there; the state colors are currently only used with notifications so those colors are defined within the [/config/rapidez/fronted.php](https://github.com/rapidez/core/blob/954aa63942da3130f0d6a2623eba013196fb982a/config/rapidez/frontend.php#L74). Maybe something for later when those are more widely used. Furthermore we've `primary` and `secondary`. We could introduce one more here but more on that later.

---

We could translate this all into the Tailwind config with a CSS variable and a default color from Tailwind's gray palette;

```js
foreground: {
    emphasis: color('--foreground-emphasis', colors.gray[800]),
    DEFAULT: color('--foreground', colors.gray[700]),
    muted: color('--foreground-muted', colors.gray[500]),
},

border: {
    emphasis: color('--border-emphasis', colors.gray[500]),
    DEFAULT: color('--border', colors.gray[200]),
    muted: color('--border-muted', colors.gray[100]),
},

background: {
    emphasis: color('--background-emphasis', colors.gray[200]),
    DEFAULT: color('--background', colors.gray[100]),
    muted: color('--background-muted', colors.gray[50]),
},
```

Not sure yet if we need them all but it gives some standard, leaving the shadow out for now. With this you can use these colors on anything;
- `text-foreground`
- `text-foreground-emphasis`
- `bg-foreground`
etc

Where the last one feels odd. But how do we end up in situations where we're doing stuff like that? These variables are here to give you some system as a base. If you need another color somewhere like `bg-foreground`, don't use it as it doesn't make sense and when somebody is changing it this will lead to unexpected color changes. Use it where it's for.

What we could do is some sort of inverse / dark mode here so these colors invert. But are there currently situations where we need that?

---

Next topic; automatic colors:
```js
textColor: theme => theme('colors.foreground'),
borderColor: theme => ({
    default: theme('colors.border'),
    ...theme('colors.border'),
}),
backgroundColor: theme => theme('colors.background'),
```

With this part the defined colors will be used by default. So these will be generated and will have the default configured color:
- `text`
- `bg`
- `border`

Only `border` is a weird one as with Tailwind `border` only applies a width. The defaults colors are defined on `*` but there are situations where that's overwritten and you'd like to set the default color again; like with Tailwind Forms where the [border color is set on inputs](https://github.com/tailwindlabs/tailwindcss-forms/blob/781de1d732cfab7d11eaef31de5e0ade9084c7a9/src/index.js#L50). In that case you can use `border-default`.

---

Still there? Buttons... how should the default button look? Use the default background? `bg-inactive` => `bg`? Next to that you'll get primary, secondary, outline and maybe an enhanced like here: https://github.com/rapidez/blade-components/pull/1? Or should that be named remarkable or highlighted? That "color" is there for conversion elements like add to cart buttons, checkout, etc. Maybe add that with a default to the primary color? As the "enhanced" color isn't always the "primary". But currently we've a `secondary` color which isn't used anywhere? Maybe just use `secondary` for those?

---

Last thing; I don't like the RGB colors in the CSS variables. So I added a simple function in the Tailwind config:
```js
function color(variable, fallback) {
    return 'color-mix(in srgb, var('+variable+', '+fallback+') calc(100% * <alpha-value>), transparent)'
}
```
With this we can just use `color('--primary', '#2fbc85')` in the config and use HEX in the variables with still the possibility to [use opacity](https://tailwindcss.com/docs/text-color#changing-the-opacity). As backup we can also reference the Tailwind palette: `color('--primary-text', colors.white)`.


royduin:
It doesn't make much sense to have the colors in CSS from the core, so I moved them to `rapidez/rapidez`, see: https://github.com/rapidez/rapidez/pull/85. Already described this in the new v3 docs: https://docs.rapidez.io/3.x/theming.html#css

Jade-GG:
I'm having trouble seeing how this refactor would work in practice. Could you work out a small example on how, for example, an "enhanced" button inside of a little container with a background would look in practice?

To what degree would this refactor be a search-and-replace job with names in the frontend? From what I can tell we're splitting some colors up into different names even though they're the same color, so I can imagine it's a little more involved than that?

royduin:
If we like to go this way; the next step is to walk through the whole source and choose the right colors where the old color names are used.

royduin:
So the alternative suggested is:

- `neutral`
- `neutral-100`
- `neutral-110`
- `inactive`

Which translates into something like this:

```js
neutral: {
    110: colors.gray[800],
    DEFAULT: colors.gray[700],
    100: colors.gray[500],
},
inactive: colors.gray[300],
```

Where `neutral` is used as text color, so on the body you'll need: `text-neutral`. But we can also translate this to defaults with:
```js
textColor: theme => theme('colors.neutral'),
borderColor: theme => ({
    default: theme('colors.neutral.110'),
    ...theme('colors.neutral'),
}),
backgroundColor: theme => ({
    default: theme('colors.neutral.100'),
    ...theme('colors.neutral'),
}),
```

So you don't have to use "neutral" explicit; just use `bg`, `text` and `border` for the defaults. With this the example given above would become:
```blade
{{-- Or just "text" as it's the default with the config --}}
<body class="text-neutral">
    <div class="flex">
        <div class="w-2/3 bg border rounded p-3">
            <p>Lorem ipsum</p>
            
            {{-- or shorter with text-110, but 110 feels odd --}}
            <ul class="text-neutral-110">
                <li>USP</li>
                <li>USP</li>
                <li class="text-primary">USP</li>
            </ul>

            <button class="bg-secondary text-secondary-text">
                Call to action
            </button>

            {{-- or shorter with bg-110, but 110 feels odd --}}
            <div class="bg-neutral-110 border rounded p-3">
                Lorem ipsum
            </div>

            {{--
            So the default border is 110, so the "muted" version is the "default"? 
            And there is no muted, but we've inactive for that?
            And we could replace "border-neutral-110" with "border-default"
            --}}
            <input type="text" class="border border-neutral-110 focus:border-neutral disabled:border-inactive">
        </div>
        {{--
        There lighter version of the background is inactive?
        And there is no "emphasis" / "darker" border?
        --}}
        <div class="w-1/3 bg-inactive border border-?">
            <ul>
                <li><a href="#">Filter</a></li>
                {{-- Or we could write these shorter with text-... --}}
                <li><a href="#" class="text-neutral-100">Filter disabled</a></li>
                <li><a href="#" class="text-neutral-110">Filter active</a></li>
            </ul>
        </div>
    </div>
</body>
```

indykoning (dismissed):

Roene-JustBetter (approved):
✅ . Checked it locally: 
- Homepage
- Product overview
- Product detail
- Checkout

resources/views/product/overview.blade.php
```diff
@@ -52,7 +52,7 @@
         </div>
     </div>
     @if (App::providerIsLoaded('Rapidez\Reviews\ReviewsServiceProvider'))
-        <div class="my-5 py-8 bg-highlight min-h-[515px]">
+        <div class="my-5 py-8 bg-emphasis min-h-[515px]">
```
royduin, line 55:
```suggestion
        <div class="my-5 py-8 bg min-h-[515px]">
```

Roene-JustBetter (approved):

resources/views/components/radio.blade.php
```diff
@@ -1,6 +1,6 @@
 <label class="flex items-center">
-    <input type="radio" {{ $attributes->merge(['class' => 'focus:ring-neutral h-4 w-4 text-neutral border']) }}>
-    <div class="ml-2 text-gray-700">
+    <input type="radio" {{ $attributes->merge(['class' => 'focus:ring-border-default size-4 text border']) }}>
```
jordythevulder, line 2: ring-border-default?

resources/views/listing/partials/filter/boolean.blade.php
```diff
@@ -19,7 +19,10 @@ class="relative pb-4"
                         v-bind:checked="value[item.key]"
                         v-on:change="handleChange(item.key)"
                     >
-                        <div class="font-sans font-medium text-sm items-center flex text-inactive" :class="{'text-neutral': value[item.key] == true}">
+                        <div
+                            class="font-sans font-medium text-sm items-center flex"
+                            :class="value[item.key] ? 'text' : 'text-muted'"
```
jordythevulder, line 24: i see border-default but here i still see "text" do we use "-default" everywhere or do we use "border", "bg", "text"?

resources/views/product/overview.blade.php
```diff
@@ -25,13 +25,13 @@
...
                         <div class="border-t pt-5 text-lg font-bold">@lang('Description')</div>
-                        <div class="prose text-inactive" itemprop="description">
+                        <div class="prose text-muted" itemprop="description">
                             {!! $product->description !!}
                         </div>
                     </div>
                     <div>
                         <div class="mb-2 border-t pt-5 text-lg font-bold">@lang('Specifications')</div>
-                        <dl class="flex flex-col text-inactive [&>dd]:rounded [&>dd]:p-2 odd:[&>dd]:bg-highlight odd:[&>dd]:font-semibold odd:[&>dd]:text-neutral even:[&>dd]:pl-4">
+                        <dl class="flex flex-col text-muted [&>dd]:rounded [&>dd]:p-2 odd:[&>dd]:bg odd:[&>dd]:font-semibold odd:[&>dd]:text even:[&>dd]:pl-4">
```
jordythevulder, line 34: this can be *: since everything is a <dd

resources/views/components/radio.blade.php
royduin: This will be removed and https://github.com/rapidez/blade-components will be used, the files: https://github.com/rapidez/blade-components/tree/master/resources/views/components/input/radio, please check those; if anything needs to change it needs to change there.

resources/views/listing/partials/filter/boolean.blade.php
royduin: Just `border`, `bg` and `text` where ever you can. The `border-default` is only used to overwrite a border color (like with Tailwind Forms as mentioned in the PR description)

resources/views/product/overview.blade.php
royduin: A bit out of scope for this PR, but every improvement is welcome. Who want to PR this?

jordythevulder (approved):

resources/views/product/overview.blade.php
royduin: Seems like that happend here: https://github.com/rapidez/core/pull/307/files#diff-99969e2a5bbf40165a88c6e09d3d7644a5f2c99598dfca5fbe042d178542a5cf, could you fix that @JimmyHoenderdaal? It should stay a dl with dt's and dd's. Please double check the styling! Wrapping the dt+dd within a div is allowed as also used within the cart/checkout sidebar.

JimmyHoenderdaal: Internal ref: RAP-1041

JimmyHoenderdaal (approved):

royduin:
Shadow, outline and ring colors (#691)

royduin:
A follow up on this for shadow, outline and ring colors: https://github.com/rapidez/core/pull/691

PR #624, opened 2024-10-31
Added gallery thumbnail count

jordythevulder:
This will add a customizable thumbnail quantity.

image


royduin:
Can't we extract some of this somewhere else? Maybe a bit to much logic within the template.

resources/views/product/partials/images.blade.php
```diff
@@ -1,75 +1,110 @@
...
+                        <x-heroicon-o-chevron-left class="size-8 text-inactive" />
+                    </button>
+                    <button class="z-10 top-1/2 right-3 -translate-y-1/2 absolute" v-if="active != images.length-1" v-on:click="change(active+1)" aria-label="@lang('Next')">
+                        <x-heroicon-o-chevron-right class="size-8 text-inactive" />
+                    </button>
+                </div>
+
+                <x-rapidez::no-image v-else class="h-96 rounded" />
+
+                <div v-if="images.length > 1" class="flex mt-3 gap-2">
```
Roene-JustBetter, line 48: Can we split the thumbnails to an own file?

resources/views/product/partials/images.blade.php
```diff
@@ -1,75 +1,110 @@
...
+                        :src="config.media_url + '/catalog/product' + images[active]"
                         alt="{{ $product->name }}"
-                        class="object-contain w-full m-auto max-h-[80px]"
+                        class="object-contain max-h-full mx-auto block"
                         loading="lazy"
-                        width="80"
-                        height="80"
                     />
+                    <div class="z-10 pointer-events-none fixed top-3 right-3">
+                        <x-heroicon-o-x-mark class="h-6 w-6" />
```
Roene-JustBetter, line 98: `size-6`

Roene-JustBetter (changes requested):

Roene-JustBetter (dismissed):

JimmyHoenderdaal (dismissed):

resources/views/product/partials/gallery/thumbnails.blade.php
```diff
@@ -0,0 +1,39 @@
...
+        <span
+            v-if="(imageId + 1) < images.length"
+            v-on:click="toggleZoom()"
+            class="absolute inset-0 items-center justify-center bg-neutral/20 hidden"
+            :class="{
+                'xl:!flex': imageId === {{ config('rapidez.frontend.product_gallery_thumbnails.xl') }},
+                'lg:max-xl:!flex': imageId === {{ config('rapidez.frontend.product_gallery_thumbnails.lg') }},
+                'md:max-lg:!flex': imageId === {{ config('rapidez.frontend.product_gallery_thumbnails.md') }},
+                'sm:max-md:!flex': imageId === {{ config('rapidez.frontend.product_gallery_thumbnails.sm') }},
+                'max-sm:!flex': imageId === {{ config('rapidez.frontend.product_gallery_thumbnails.xs') }}
```
Jade-GG, line 31: Do these need to be `important`s? I'd assume that breakpoints always take precedence over the `hidden` above so these shouldn't be necessary right?

resources/views/product/partials/gallery/thumbnails.blade.php
```diff
@@ -0,0 +1,39 @@
+<div v-if="images.length > 1" class="flex mt-3 gap-2">
+    <button
+        v-for="(image, imageId) in images.slice(0, {{ max(config('rapidez.frontend.product_gallery_thumbnails.xs'), config('rapidez.frontend.product_gallery_thumbnails.sm'), config('rapidez.frontend.product_gallery_thumbnails.md'), config('rapidez.frontend.product_gallery_thumbnails.lg'), config('rapidez.frontend.product_gallery_thumbnails.xl')) + 1 }})"
```
royduin, line 3: `max(config('rapidez.frontend.product_gallery_thumbnails'))` should be enough, but... I don't like the config for such a small part separated from the template. Maybe just define them on top of here?
```blade
@php($breakpoints = ['xl' => 7, 'lg' => 5, 'md' => 4, 'sm' => 3, 'xs' => 4])
```
And use that here; `max($breakpoints)` and through the template

Roene-JustBetter: But if we _only_ want to change the breakpoints and everything else stays the same, then do we have to override this file in our project just to change breakpoints instead of changing it in the config file?

resources/views/product/partials/gallery/thumbnails.blade.php
jordythevulder: You're completely right! Removed them. Did need them before adding media queries. thanks

3,
royduin, line 1: Wow, what happend?

config/rapidez/frontend.php
```diff
@@ -78,17 +78,17 @@
         ],
     ],
 
+    // Add to cart settings to automaticly select configurable- or product options (true/false)
```
Jade-GG, line 81:
```suggestion
    // Add to cart settings to automatically select configurable- or product options (true/false)
```

PR #626, opened 2024-11-01
[3.x] Make sure you actually log in in the checkout

Jade-GG:
See: https://github.com/rapidez/core/pull/613


royduin:
It seems like the tests are failing since this PR: https://github.com/rapidez/core/actions/workflows/dusk.yml?query=branch%3Amaster could you check that 😇

PR #628, opened 2024-11-04
[3.x] Disable checkout button and add notice if not in stock

BobWez98:
Whenever a product in cart is out of stock, there will be a notice with the product that notifies the user of the product being out of stock. This functionality also disables the "proceed to checkout" button.

V2: #627 

Internally there's a story to have some extra frontend added for this.


resources/js/package.js
```diff
@@ -92,6 +92,15 @@ function init() {
...
                 }
             },
+
+            canOrderCartItem(item) {
+                if ('stock_item' in item.product && 'in_stock' in item.product.stock_item && item.product.stock_item.in_stock !== null) {
+                    return item.product.stock_item.in_stock
+                }
+
+                return true
+            },
```
indykoning, line 102: The Cart Store may be a more logical place for this.
We're already adding additional data to items here https://github.com/rapidez/core/blob/c2a0bd78042822d5ac94e832acf958a0dda17092/resources/js/stores/useCart.js#L203

Maybe we could do it in a way to polyfill the currently broken `is_available` so once we drop support for Magento versions that do have it in it's broken state we can remove this code without breaking changes to the frontend

PR #644, opened 2024-11-14
Add support for HTTP/3 Early Hints

indykoning:
Early hints can improve performance and lighthouse scores by returning Preload and preconnect link headers BEFORE the server has it's response ready. So JS, CSS, and fonts could be loaded before your HTML has finished downloading.

lowering the time to first paint, reducing layout shifts caused by fonts, and improving the speed at which the JS can start executing.

This will automatically install the early hints package, and configure it to work with Rapidez and all it's fallback routes.

docs: https://github.com/rapidez/docs/pull/63

(Do we want this in V2 as well?)

### DIFF
--- a/config/rapidez/routing.php
+++ b/config/rapidez/routing.php
@@ -23,4 +23,8 @@ return [
         // This does not cache the response, it caches the controller used for that page.
         'cache_duration' => 3600,
     ],
+
+    'earlyhints' => [
+        'enabled' => env('EARLY_HINTS_ENABLED', true),
+    ]
 ];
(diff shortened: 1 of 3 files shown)


config/rapidez/routing.php
```diff
@@ -23,4 +23,8 @@
         // This does not cache the response, it caches the controller used for that page.
         'cache_duration' => 3600,
     ],
+
+    'earlyhints' => [
+        'enabled' => env('EARLY_HINTS_ENABLED', true),
+    ]
```
royduin, line 29:
```suggestion

    // See: https://docs.rapidez.io/3.x/configuration.html#early-hints
    'earlyhints' => [
        'enabled' => env('EARLY_HINTS_ENABLED', true),
    ]
```

royduin:
Just v3 is fine, thanks! 🚀

PR #608, opened 2024-10-22
[3.x] Split up quantity select into 2 variants

BobWez98:
V3 PR for #595

PR #603, opened 2024-10-21
Styling quantity select / enhanced quantity select

Roene-JustBetter:
Visual fixes for quantity select(s)

_Quantity select enhanced:_
Scherm­afbeelding 2024-10-21 om 11 39 50
Scherm­afbeelding 2024-10-21 om 11 49 37

Quantity select:
Scherm­afbeelding 2024-10-21 om 12 03 09
Scherm­afbeelding 2024-10-21 om 11 49 20

PR #595, opened 2024-10-16
[2.x] Split up quantity select into 2 variants

BobWez98:
Once this is merged i'll add the PR for v3.

This PR splits up the quantity select in 2 variants; a default and a enhanced version with plus and minus buttons. They use the same props so they can be switched seamlessly. 
The use on the product page: 
```
<x-rapidez::quantity :minSaleQty="$product->min_sale_qty" model="addToCart.qty" :qtyIncrements="$product->qty_increments" />
```
Or
```
<x-rapidez::quantity.enhanced :minSaleQty="$product->min_sale_qty" model="addToCart.qty" :qtyIncrements="$product->qty_increments" />
```

This component can also be used in the cart, we replace the form [here](https://github.com/rapidez/core/blob/master/resources/views/cart/item/quantity.blade.php#L8) by:
```
<div class="flex items-center gap-1">
        <x-rapidez::quantity @change="mutate" index="item.id" model="variables.quantity" minSaleQty="(item.product.stock_item?.min_sale_qty ?? 1)" qtyIncrements="(item.product.stock_item?.qty_increments ?? 1)"/>
</div>
```
The only difference is that the default quantity needs an extra `@change` eventlistener.
or 
```
<div class="flex items-center gap-1">
        <x-rapidez::quantity.enhanced index="item.id" model="variables.quantity" minSaleQty="(item.product.stock_item?.min_sale_qty ?? 1)" qtyIncrements="(item.product.stock_item?.qty_increments ?? 1)"/>
</div>
```

Here we use the index to determine which input to update. In the EnhancedQuantitySelect.vue.

PR #645, opened 2024-11-15
Quantity refactor

royduin:
Used https://github.com/rapidez/core/pull/608 as base which is the v3 version of https://github.com/rapidez/core/pull/595. Only v2 is styled here: https://github.com/rapidez/core/pull/603 but with this that needs to applied again.

With this we've just one quantity component which is a better choice so we don't have to maintain 2 versions.

Maybe make this a v3 feature and drop the v2 PR's?

Thoughts? The cleanest would be to just have an input and only use `stepUp()` and `stepDown()` on the buttons but that's changing the DOM directly so Vue doesn't see it, if we can find another way we could get rid of the Vue component.

I also saw the `use_config_` column aren't used which could result in incorrect values, no issue within the cart as it's coming from Magento directly through GraphQL. Maybe something?

Not sure yet if Dusk will run fine as currently the `dusk="qty"` doesn't have an unique id.


resources/views/components/quantity.blade.php
```diff
@@ -0,0 +1,34 @@
...
+        >
+            <div class="bg-neutral h-0.5 w-3 hover:bg-white"></div>
+        </x-rapidez::button.outline>
+        <input
+            name="qty"
+            type="number"
+            dusk="qty"
+            class="w-14 px-1 text-center"
+            aria-label="@lang('Quantity')"
+            {{ $attributes }}
```
royduin, line 21: As an alternative we could do something like this to avoid the usage of `$attributes` twice:
```
@attributes([
    ':min' => 'qtySelect.min',
    ':max' => 'qtySelect.max',
    ':step' => 'qtySelect.step',
    'v-model.number' => $attributes->get('v-model.number'),
    'v-on:change' => $attributes->get('v-on:change'),
])
```
But this feels pretty repeating and needs to be updated manually when something else is added.

BobWez98 (dismissed):

JimmyHoenderdaal:
Need the new color changes inside here

JimmyHoenderdaal (changes requested):
This needs to be updated with the new color changes

royduin:
On hold until it's styled by @Roene-JustBetter

Roene-JustBetter:
> On hold until it's styled by @Roene-JustBetter

Styling have been changed.
Scherm­afbeelding 2024-12-12 om 13 03 01
Scherm­afbeelding 2024-12-12 om 13 13 56

Roene-JustBetter (approved):

royduin:
Thanks @Roene-JustBetter! 🚀 tomorrow I'll check the Dusk tests 🤞🏻

PR #646, opened 2024-11-15
On initial page load run init before turbo:load has been fired

indykoning:
This will make Rapidez initialize before marketing scripts, browser extensions and other scripts run.
Improving startup times of Rapidez.

Before:
image

After:
image

Currently as draft as we should figure out the best method/event to run this on in order not to impact LCP

### DIFF
--- a/resources/js/package.js
+++ b/resources/js/package.js
@@ -15,11 +15,10 @@ import useMask from './stores/useMask'
 import { swatches, clear as clearSwatches } from './stores/useSwatches'
 import { clear as clearAttributes } from './stores/useAttributes.js'
 import './vue'
-import { computed } from 'vue'
 import './fetch'
 import './filters'
 import './mixins'
-import './turbolinks'
+(() => import('./turbolinks'))()
 import './cookies'
 import './callbacks'
 import './vue-components'
@@ -116,7 +115,7 @@ function init() {
                             Turbo.visit(window.url('/search?q=' + encodeURIComponent(value)))
                         }
                     },
-                    
+
                     setSearchParams(url) {
                         window.history.pushState(window.history.state, '', new URL(url))
                     },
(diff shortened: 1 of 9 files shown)


indykoning:
After the last commit the LCP is no longer impacted.
We no longer fire before ALL browser extensions, but we do fire earlier.
After:
image

Also i've placed more in timeouts and animation frames, so they schedule their own tasks instead of putting it all on a task that's already taking too long

resources/js/package.js
```diff
@@ -59,96 +91,80 @@ function init() {
...
+            setTimeout(() => {
+                const event = new CustomEvent('vue:loaded', { detail: { vue: window.app } })
+                document.dispatchEvent(event)
+            })
+        }
+    )
 }
+
 document.addEventListener('turbo:load', init)
+setTimeout(init)
```
indykoning, line 170: This specific order of setTimeout and [requestAnimationFrame](https://github.com/rapidez/core/pull/646/files#diff-251b99eac1abebedfe294a2334032b29b5f843d0b5d0834c32c055635f79c935R94) ensures LCP is not impacted.
And ensures small task time for the `turbo:load`/evaluate script, and `vue:loaded` events by scheduling them all in their own task.

Using a setTimeout for both, a requestAnimationFrame for both, or executing this immediately and booting Vue in a requestAnimationFrame all has impact on the LCP

royduin:
So this is still work in progress? Is it tested within a real project?

royduin:
I'll change it to a draft, please mark it when it's ready to merge

indykoning:
It seems the tests were caused by the order that event listeners were being applied in.
Vue could not prevent the event listener for turbolinks causing a navigation before the vue events were fired.

Loading turbo lazily fixes this issue.

PR #651, opened 2024-11-18
Frontend colors suggestion change

Roene-JustBetter:
**Changed colors**
- highlight = neutral-100
- border = neutral-110

**Added colors**
- Enhanced color 
- Enhanced text color
_The enhanced color will most likely be used for conversions, such as "Add to Cart" buttons._

Updated tailwind.config with colors:
Scherm­afbeelding 2024-11-18 om 14 05 19

If we are going to use the colors above we should also change it in here: 
https://github.com/rapidez/reviews & https://github.com/rapidez/account

_Note_
Buttons & inputs will be replaced with the buttons from the [blade-components](https://github.com/rapidez/blade-components). There is a draft here: https://github.com/rapidez/core/pull/632

### DIFF
--- a/resources/views/checkout/steps/agreements.blade.php
+++ b/resources/views/checkout/steps/agreements.blade.php
@@ -21,7 +21,7 @@
     </x-rapidez::slideover>
 
     @if ($agreement->mode == 'AUTO')
-        <label class="text-neutral cursor-pointer underline hover:no-underline" for="{{ $agreement->agreement_id }}_agreement">
+        <label class="cursor-pointer underline hover:no-underline" for="{{ $agreement->agreement_id }}_agreement">
             {{ $agreement->checkbox_text }}
         </label>
     @else
@@ -32,7 +32,7 @@
                 dusk="agreements"
                 required
             >
-                <label class="text-neutral cursor-pointer underline hover:no-underline" for="{{ $agreement->agreement_id }}_agreement">
+                <label class="cursor-pointer underline hover:no-underline" for="{{ $agreement->agreement_id }}_agreement">
                     {{ $agreement->checkbox_text }}
                 </label>
             </x-rapidez::checkbox>
--- a/resources/views/checkout/steps/agreements.blade.php
+++ b/resources/views/checkout/steps/agreements.blade.php
@@ -21,7 +21,7 @@
     </x-rapidez::slideover>
 
     @if ($agreement->mode == 'AUTO')
-        <label class="text-gray-700 cursor-pointer underline hover:no-underline" for="{{ $agreement->agreement_id }}_agreement">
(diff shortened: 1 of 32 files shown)


resources/views/checkout/steps/agreements.blade.php
```diff
@@ -21,7 +21,7 @@
     </x-rapidez::slideover>
 
     @if ($agreement->mode == 'AUTO')
-        <label class="text-gray-700 cursor-pointer underline hover:no-underline" for="{{ $agreement->agreement_id }}_agreement">
+        <label class="text-neutral cursor-pointer underline hover:no-underline" for="{{ $agreement->agreement_id }}_agreement">
```
jordythevulder, line 24: text-neutral shouldn't be necessary since the body text is already set on the <body tag

resources/views/components/select.blade.php
```diff
@@ -10,7 +10,7 @@
         'id' => $name ?? null,
         'name' => $name ?? null,
         'dusk' => $attributes->get('v-bind:dusk') ? null : $name ?? null,
-        'class' => 'w-full py-2 pl-3 pr-8 border-gray-200 rounded focus:ring-neutral focus:border-neutral sm:text-sm'
+        'class' => 'w-full py-2 pl-3 pr-8 border border-neutral-110 rounded outline-0 ring-0 focus:ring-transparent focus:border-neutral sm:text-sm'
```
jordythevulder, line 13: border-neutral-110 shouldn't be necessary since it's set by default on the "border" class

tailwind.config.js
```diff
@@ -22,20 +22,26 @@ export default {
...
+                    DEFAULT: 'rgb(var(--neutral) / <alpha-value>)',
+                    100: 'rgb(var(--neutral-100) / <alpha-value>)',
+                    110: 'rgb(var(--neutral-110) / <alpha-value>)',
+                },
+                enhanced: {
+                    DEFAULT: 'rgb(var(--enhanced) / <alpha-value>)',
+                    text: 'rgb(var(--enhanced-text) / <alpha-value>)',
+                },
+                inactive: 'rgb(var(--inactive) / <alpha-value>)',
             },
```
jordythevulder, line 42: can enhanced be placed under secondary? this groups the default / default text and groups the neutral with the inactive. Feels a bit more complete

JimmyHoenderdaal (dismissed):

resources/views/checkout/steps/agreements.blade.php
Roene-JustBetter: Correct, I have removed this.

jordythevulder (approved):

JimmyHoenderdaal (approved):

PR #654, opened 2024-11-19
[2.x] Add support for HTTP/3 Early Hints

indykoning:
See: https://github.com/rapidez/core/pull/644

This change would require dropping L10 support as we have with V3
(due to the new middleware syntax)


royduin:
Just v3 is fine, thanks! 🚀

PR #661, opened 2024-11-27
Init after turbolinks load

indykoning:
Unfortunately #646 introduced a bug breaking the autocomplete.
Oddly enough this should already be broken as documented by:
https://v2.vuejs.org/v2/guide/components-registration#:~:text=global%20registration%20must%20take%20place%20before%20the%20root%20Vue%20instance%20is%20created

However when turbolink is loaded first this does not seem to be an issue.

As a workaround we init Vue after turbolinks has been loaded so this functionality does not break.

This reduces the impact of that PR, however doesn't break any functionality:

Before:
image
This is what it looked like without browser extensions:
image

After:
image
This is what it looks like without browser extensions:
image

### DIFF
--- a/resources/views/layouts/app.blade.php
+++ b/resources/views/layouts/app.blade.php
@@ -20,7 +20,10 @@
     <meta name="robots" content="@yield('robots', Rapidez::config('design/search_engine_robots/default_robots', 'INDEX,FOLLOW'))"/>
     <link rel="canonical" href="@yield('canonical', url()->current())" />
 
-    @vite(['resources/css/app.css', 'resources/js/app.js', 'vendor/rapidez/core/resources/js/turbolinks.js'])
+    @vite(['resources/css/app.css', 'resources/js/app.js'])
+    @if ($file = vite_filename_path('turbolinks.js'))
+        @vite([$file])
+    @endif
 
     @stack('head')
     @config('design/head/includes')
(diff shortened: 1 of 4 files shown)


resources/views/layouts/app.blade.php
```diff
@@ -20,7 +20,7 @@
     <meta name="robots" content="@yield('robots', Rapidez::config('design/search_engine_robots/default_robots', 'INDEX,FOLLOW'))"/>
     <link rel="canonical" href="@yield('canonical', url()->current())" />
 
-    @vite(['resources/css/app.css', 'resources/js/app.js'])
+    @vite(['resources/css/app.css', 'resources/js/app.js', 'vendor/rapidez/core/resources/js/turbolinks.js'])
```
royduin, line 23: Can't we just eager load it so we don't have to specify it here?

indykoning: I had done it this way in the previous PR because not doing so caused issues.
However i cannot reproduce that issue anymore so my further work has probably fixed that.

I'll update it and just eager load it

royduin:
Dusk is failing 😇

resources/views/layouts/app.blade.php
indykoning: Unfortunately i have run into the issue again, the order of event listeners are incorrect so Vue cannot prevent the turbo navigation

PR #675, opened 2024-12-16
[3] Fix layout shift on thumbnails, quantity

JimmyHoenderdaal:
Removed absolute php image so we don't get any layout shifts anymore
Did some more tweaking for the quantity and reviews

For the image thumbnails we couldn't merge the PHP and Vue code because of product options


resources/views/components/quantity.blade.php
```diff
@@ -6,6 +6,7 @@
 <quantity-select v-slot="qtySelect" {{ $attributes }}>
     <div class="flex items-center justify-center border rounded bg-white h-12 self-start">
         <button
+            disabled
```
Roene-JustBetter, line 9: Why have you added this?

resources/views/components/quantity.blade.php
```diff
@@ -17,6 +18,7 @@ class="shrink-0 pl-2.5 text disabled:cursor-not-allowed disabled:opacity-50"
             name="qty"
             type="number"
             dusk="qty"
+            value="1"
```
Roene-JustBetter, line 21: I think we don't need a value here.

Roene-JustBetter (changes requested):

resources/views/components/quantity.blade.php
JimmyHoenderdaal: because the left button is disabled after full pageload, we just set it to disabled before JS loads so you don't have any changes

After JS is loaded it uses that value instead

resources/views/components/quantity.blade.php
JimmyHoenderdaal: We don't but this is so we don't have any changes after JS has loaded, before it was just empty before JS loaded
its just a small thing that prevents any visual changes after JS has loaded

Roene-JustBetter (dismissed):

resources/views/product/partials/gallery/slider.blade.php
```diff
@@ -1,23 +1,33 @@
...
     <a
-        :href="config.media_url + '/catalog/product' + images[active]"
-        class="flex items-center justify-center border rounded p-5 h-[440px] cursor-zoom-in"
         v-on:click.prevent="toggleZoom"
+        v-bind:href="config.media_url + '/catalog/product' + images[active]"
+        class="flex h-[440px] cursor-zoom-in items-center justify-center rounded border p-5"
     >
         <img
-            :src="'/storage/{{ config('rapidez.store') }}/resizes/400/magento/catalog/product' + images[active] + '.webp'"
+            src="{{ route('resized-image', [
```
Jade-GG, line 8: The dusk tests are failing because this route name does not exist. Can you check that?

resources/views/components/reviews/stars.blade.php
royduin, line 1: This should go in https://github.com/rapidez/reviews, not sure if the readme is up-to-date?

JimmyHoenderdaal: Yes, I have made this PR for reviews https://github.com/rapidez/reviews/pull/28

Roene-JustBetter (dismissed):

jordythevulder (dismissed):

Roene-JustBetter (dismissed):

jordythevulder (approved):

royduin (changes requested):
It should be possible to have 1 version without the duplication.

Jade-GG:
> It should be possible to have 1 version without the duplication.

To a certain degree, we could almost certainly write some really ugly unreadable frontend that doesn't have duplication. It's made a lot more complex because product options can vary the amount of thumbnails shown. You would have to somehow merge a dynamic `v-for` with the php `@foreach`, which can't really be done cleanly.

The way to do this would be to always fill the whole row of thumbnails regardless of how many there actually are on the base product, basically replacing the `v-for` with a set of `v-if`s. Would look something like this (untested code):

```blade
@for ($imageId = 0; $imageId < max($breakpoints); $imageId++)
    <button
        v-if="{{ $imageId }} < images.length"
        v-on:click="change({{ $imageId }})"
        @class([
            $baseClasses,
            'outline outline-1 border-primary' => $imageId == 0,
            'xl:hidden' => $imageId > $breakpoints['xl'],
            'lg:max-xl:hidden' => $imageId > $breakpoints['lg'],
            'md:max-lg:hidden' => $imageId > $breakpoints['md'],
            'sm:max-md:hidden' => $imageId > $breakpoints['sm'],
            'max-sm:hidden' => $imageId > $breakpoints['xs'],
        ])
        v-bind:class="{ 'outline outline-1 border-primary': active == {{ $imageId }} }"
    >
        <img
            src="/storage/{{ config('rapidez.store') }}/resizes/80x80/magento/catalog/product/{{ $product->images[$imageId] }}.webp"
            v-bind:src="'/storage/{{ config('rapidez.store') }}/resizes/80x80/magento/catalog/product/' + images[{{ $imageId }}] + '.webp'"
            alt="{{ $product->name }}"
            class="block max-h-full w-auto object-contain"
            width="80"
            height="80"
        />
        @if ($imageId < count($product->images) - 1)
            <span @class([
                'absolute inset-0 hidden items-center justify-center bg-black/20',
                'xl:flex' => $imageId === $breakpoints['xl'],
                'lg:max-xl:flex' => $imageId === $breakpoints['lg'],
                'md:max-lg:flex' => $imageId === $breakpoints['md'],
                'sm:max-md:flex' => $imageId === $breakpoints['sm'],
                'max-sm:flex' => $imageId === $breakpoints['xs'],
            ])>
                <span
                    class="size-9 flex items-center justify-center rounded-full bg-white text-sm font-bold text shadow-lg"
                    v-text="'+' + images.length - {{ $imageId }} - 1"
                >
                    +{{ count($product->images) - $imageId - 1 }}
                </span>
            </span>
        @endif
    </button>
@endfor
```

resources/views/product/partials/gallery/thumbnails.blade.php
Roene-JustBetter, line 1: I would prefer on version instead of 2 versions if that is possible.

Roene-JustBetter (changes requested):

royduin:
Closing in favor of https://github.com/rapidez/core/pull/710

PR #679, opened 2024-12-20
[3] Add blade reviews to templates

JimmyHoenderdaal:
this should be merged after: https://github.com/rapidez/reviews/pull/29


resources/views/product/overview.blade.php
```diff
@@ -20,7 +20,7 @@
                     @includeWhen($product->type_id !== 'grouped', 'rapidez::product.partials.addtocart')
                     @if (App::providerIsLoaded('Rapidez\Reviews\ReviewsServiceProvider'))
                         @if ($product->reviews_score)
-                            <stars :score="{{ $product->reviews_score }}" :count="{{ $product->reviews_count }}"></stars>
+                            <x-rapidez-reviews::stars score="{{ $product->reviews_score }}" count="{{ $product->reviews_count }}" />
```
royduin, line 23:
```suggestion
                            <x-rapidez-reviews::stars :score="$product->reviews_score" :count="$product->reviews_count" />
```

This also works right?

royduin:
- There is a conflict
- As this requires the latest version of `rapidez/reviews` we should make sure Composer knows. But.. it's not a requirement of the core; it's an optional package. I think defining a conflict is the best option here: https://getcomposer.org/doc/04-schema.md#conflict

PR #680, opened 2024-12-20
Remove slideover form core and use blade-components slideover

JimmyHoenderdaal:
These files can be removed as we use them from blade-components
and added the overflow-clip as the readme states
this should be merged first for it to fit together properly
https://github.com/rapidez/blade-components/pull/13
(otherwise overflow-clip and the z-indexes won't be correct)


Roene-JustBetter (approved):

jordythevulder (approved):

royduin:
It seems like this is already merged, only the slideover needs to be removed. Please double check

PR #683, opened 2024-12-23
[3.x] Fix autocomplete & some other reactivesearch errors

Jade-GG:
This PR most importantly fixes the Autocomplete which breaks when used on a fresh page load.

I don't know if this is the solution we really want here... it works, but it's not very nice. It's the same thing that's used in the [listing](https://github.com/rapidez/core/blob/master/resources/views/components/listing.blade.php#L3-L9), except it doesn't push into the head because for some reason that didn't work.

Other fixes in this PR include a small error caused by going to a category page without subcategories, and checking if turbo actually exists before trying to load the page size from it.

### DIFF
--- a/resources/js/components/Listing/Filters/CategoryFilter.vue
+++ b/resources/js/components/Listing/Filters/CategoryFilter.vue
@@ -95,7 +95,7 @@ export default {
                     }
 
                     // No child categories, get siblings instead.
-                    let parentCategoryId = categoryPaths[0].key.at(-2)
+                    let parentCategoryId = categoryPaths[0]?.key.at(-2)
                     return allCategoryPaths.filter(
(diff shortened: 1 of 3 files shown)


royduin:
Previously this worked and the autocomplete file was loaded when clicking in the input. With this it's not lazy loading anymore. What changed?

Jade-GG:
> Previously this worked and the autocomplete file was loaded when clicking in the input. With this it's not lazy loading anymore. What changed?

I'm not sure, I tried a lot of things but nothing really worked 🙁 bit of a weird frustrating bug.

royduin:
@indykoning will look into it with RAP-1109

Jade-GG:
Fixed in different PR.

PR #685, opened 2024-12-30
[3.x] Add getCustomAttributeValue function

Jade-GG:
Right now, if you want to find the value of a custom_attribute value, you need to do something like this:

```js
data.customer.custom_attributes?.find(attribute => attribute.code == 'my_attribute')?.value
```

This is annoying and ugly, especially when found in blade templates and especially when combined into longer statements.

With this function, we can simplify this and make it much more readable:

```js
getCustomAttributeValue(data.customer, 'my_attribute')
```

It also handles `selected_options` and transforms that into an object with the option values as keys. Before this PR, that would have to be written something like this:

```js
data.customer.custom_attributes?.find(attribute => attribute.code == 'my_attribute')?.selected_options?.find(option => option.value == 'my_option')
```

With this PR, it can be written like this instead:

```js
getCustomAttributeValue(data.customer, 'my_attribute')?.my_option
```

### DIFF
--- a/resources/js/package.js
+++ b/resources/js/package.js
@@ -139,6 +139,24 @@ function init() {
 
                     return `/storage/${store}/resizes/${size}/magento${url}`
                 },
+
+                getCustomAttributeValue(object, code) {
+                    let attribute = object?.custom_attributes?.find((attribute) => attribute.code == code) ?? null
+
+                    if (!attribute) {
+                        return null
+                    }
+
+                    if ('value' in attribute) {
+                        return attribute.value
+                    }
+
+                    if ('selected_options' in attribute) {
+                        return Object.fromEntries(attribute.selected_options.map((option) => [option.value, option.label]))
+                    }
+
+                    return null
+                },
             },
             computed: {
                 // Wrap the local storage in getter and setter functions so you do not have to interact using .value


royduin:
We don't actually do that within the core? It only appears in projects?

Jade-GG:
> We don't actually do that within the core? It only appears in projects?

Correct, it turns out the only place we actually use it across the rapidez github repositories would be [here](https://github.com/rapidez/amasty-custom-form/blob/429db37a0eb70740cca5f4facd407e16409b7ae7/resources/js/callbacks.js#L32).

royduin:
Let's wait for @indykoning his approval

resources/js/package.js
```diff
@@ -139,6 +139,24 @@ function init() {
...
+                    if (!attribute) {
+                        return null
+                    }
+
+                    if ('value' in attribute) {
+                        return attribute.value
+                    }
+
+                    if ('selected_options' in attribute) {
+                        return Object.fromEntries(attribute.selected_options.map((option) => [option.value, option.label]))
```
indykoning, line 155: With the current implementation we would either get an array or direct value depending on the attribute.

How would this be used and would it be more preferable to return an array in both cases so the implementation of the function can always be the same?

Instead of having to check wether we received an array.

Jade-GG: I don't necessarily see this as a problem, because I figure when you're programming you would know what type of variable you're retrieving data from. You'd have the same issue when you're doing it the old way (except worse).

I definitely don't think it would be preferable to have this function always return one consistent type because the two possibilities are just fundamentally different data. If you wanted something for that, I think the best way would be to have two different functions.

indykoning: This is something we currently do for attribute_values on the cart

https://github.com/rapidez/core/blob/941cf91a332db3f87bf87f31a2238b701ce01f9d/resources/js/stores/useCart.js#L239

in order to make usage in template

https://github.com/rapidez/core/blob/e11744a4ed9105b367ca33dc38aa340f8518dbc2/resources/views/cart/item.blade.php#L36

more consistent

Jade-GG: I wouldn't say that's the same situation here though, as we've got an Object instead of an Array. However, something we could do is always return an object from `getCustomAttributes(data.customer)` and then you can use it like:

```js
getCustomAttributes(data.customer).my_attribute
getCustomAttributes(data.customer).my_options.option
```

maybe that's better? It kind ends up being the same thing, where `my_attribute` is a singular value but `my_options` is an Object.

We could also turn `my_attribute` here into an Object with a `value` so you always have to do `.value` after it... I'm not sure if that's any nicer.

royduin:
Closed in favor or RAP-1196

PR #694, opened 2025-01-08
Fix the disabled options to work if there's only one super attribute

BobWez98:
When a product has only one super attribute, this happens here: https://github.com/rapidez/core/blame/2.x/resources/js/components/Product/AddToCart.vue#L353 

When there's only one super attribute, the cross reference will return right away and because of this the option will never be added to the disabled options array: https://github.com/rapidez/core/blame/2.x/resources/js/components/Product/AddToCart.vue#L370 

3.x: #695

### DIFF
--- a/resources/js/components/Product/AddToCart.vue
+++ b/resources/js/components/Product/AddToCart.vue
@@ -338,6 +338,10 @@ export default {
                     }
 
                     if (!option.in_stock) {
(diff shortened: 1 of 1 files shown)


royduin:
Looking good to me, but let's wait for @indykoning as he created this part a long time ago.

PR #696, opened 2025-01-09
Add overlay color

JimmyHoenderdaal:
I've added the overlay color to the color variables so we can use this instead of bg-black/40.

Still figuring out whether we should include the opacity in the variable or add it when we use the variable (bg-overlay vs. bg-overlay/40).

Adding it when we use it could lead to inconsistencies, but having it in the config might mean less control in some cases.


Roene-JustBetter (changes requested):
I think it should work like this: `bg-overlay/40`.

resources/views/layouts/partials/header/autocomplete.blade.php
```diff
@@ -22,7 +22,7 @@ class="relative w-full"
 >
     <x-rapidez::reactive-base slot-scope="autocompleteScope">
         <div
-            class="z-header-autocomplete-overlay pointer-events-none fixed inset-0 cursor-pointer bg-black/40 opacity-0 transition duration-500"
+            class="z-header-autocomplete-overlay pointer-events-none fixed inset-0 cursor-pointer bg-overlay opacity-0 transition duration-500"
             :class="autocompleteScope.overlay ? 'pointer-events-auto opacity-100 prevent-scroll' : 'opacity-0 pointer-events-none '"
```
jordythevulder, line 26: I also think we need an <x-overlay component, so every overlay is the same. What do you think?

JimmyHoenderdaal: That would be a good idea but should this exist inside the rapidez core or blade-components you think?

Roene-JustBetter (approved):

resources/views/layouts/partials/header/autocomplete.blade.php
jordythevulder: i think the core would do

Roene-JustBetter: Little bit out of scope for this pull request?

royduin:
Don't we want the opacity within the color definition? I could imagine the opacity could be different when changing the color?

resources/views/layouts/partials/header/autocomplete.blade.php
JimmyHoenderdaal: I agree with Roene, Let's do this at a later moment, and then inside the blade-components package

JimmyHoenderdaal:
> Don't we want the opacity within the color definition? I could imagine the opacity could be different when changing the color?

That was my original thought too, I'll revert the feedback I got on this

Roene-JustBetter (approved):

royduin:
Let's wait for https://github.com/rapidez/blade-components/pull/14

Roene-JustBetter (dismissed):

royduin:
We should require this release: https://github.com/rapidez/blade-components/releases/tag/1.5.0 and the colors needs to be added to the docs.

royduin:
Conflict 😇

PR #698, opened 2025-01-09
[3.x] Fix issues with add-to-cart on listing when product has super_attributes

Jade-GG:
Within the context of the product listing:
- The add to cart button would *always* redirect to the product page even if you have options selected, making the options shown in the listing completely useless. With this PR, it will actually add the product to your cart if you have selected an option for each of the super_attributes.

- Selecting an option that should change the product image didn't actually change the product image, even though those images were all getting indexed into ES. This PR makes the product image listen to the selected options appropriately.

- The add to cart button was always hidden if you didn't elect to show swatches in the product listing, whereas that option is completely unrelated to whether or not the button should be shown. With the previous point I've had to move a few things around, along with which I've changed this behavior to be more in line with what you would expect.


resources/js/components/Product/AddToCart.vue
```diff
@@ -262,6 +258,25 @@ export default {
     },
 
     computed: {
+        currentThumbnail: function () {
+            return this.simpleProduct?.thumbnail || this.simpleProduct?.images?.[0] || this.product?.thumbnail
+        },
+
+        shouldRedirectToProduct: function () {
```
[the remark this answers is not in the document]
Jade-GG: In principle it shouldn't really matter for this check. If you have a product that doesn't have all options filled, it doesn't matter whether the options are shown or not.

resources/js/components/Product/AddToCart.vue
```diff
@@ -262,6 +258,25 @@ export default {
     },
 
     computed: {
+        currentThumbnail: function () {
+            return this.simpleProduct?.thumbnail || this.simpleProduct?.images?.[0] || this.product?.thumbnail
+        },
```
royduin, line 263: This matches how this is handled on the PDP? Possible/useful to merge/extract so it's in one place?

Jade-GG: No, the PDP actually uses the `Images.vue` component where the `product-super-attribute-change` event. It's different also because on the PDP you have the whole gallery where here you only have the singular image.

royduin:
Conflict 😇

PR #700, opened 2025-01-13
Prevent body scroll when input checked

jordythevulder:
Within the Rapidez Blade components, we use the peer-checked utility for slideovers and popups. When you open a slideover or popup, you get a full-screen view of the component. However, you can still scroll in the background of the components. To prevent this, you can add class="prevent-scroll" to the <input element. When the input is checked, the <body gets overflow: hidden, removing the scroll ability.


Roene-JustBetter (dismissed):

royduin:
This requires the latest version of `rapidez/blade-components` right? That should be bumped in the `composer.json` to https://github.com/rapidez/blade-components/releases/tag/1.4.0

PR #702, opened 2025-01-13
[3.x] fix overwriting stack with 404s

BobWez98:
Rapidez tries the routes that are in the fallbackcontroller, when they fail and throw a 404, Laravel then wants to render a 404 page overwriting the stack. This can cause for example; not loading in the resizer script added by the startPush in rapidez/statamic: https://github.com/rapidez/statamic/blob/master/src/RapidezStatamicServiceProvider.php#L246

Alternative of: https://github.com/rapidez/statamic/pull/109

2.x: #701


royduin:
Closing in favor of https://github.com/rapidez/statamic/pull/109

PR #703, opened 2025-01-14
Implement teleport in slideover

indykoning:
This PR changes the global slideover to use the Teleport component so plain html can be passed without trouble

### DIFF
--- a/resources/views/checkout/steps/agreements.blade.php
+++ b/resources/views/checkout/steps/agreements.blade.php
@@ -1,23 +1,31 @@
 <graphql query="{ checkoutAgreements { agreement_id name checkbox_text content is_html mode } }" cache="checkout.agreements" v-cloak>
     <div slot-scope="{ data }" v-if="data">
         <div v-for="agreement in data.checkoutAgreements" :key="agreement.agreement_id">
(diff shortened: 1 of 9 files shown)


Roene-JustBetter (dismissed):

resources/views/checkout/steps/agreements.blade.php
```diff
@@ -1,23 +1,31 @@
...
-                </template>
-                <div v-else>
-                    <x-rapidez::input.checkbox
-                        name="agreement_ids[]"
-                        v-bind:value="agreement.agreement_id"
-                        dusk="agreements"
-                        required
-                    >
+            <global-slideover :title="agreement.name"
+                v-slot="slideover">
```
royduin, line 5: Maybe keep this on one line?

royduin:
You're happy with the result? Should we have some docs on this? What's the simplest example how to use it?

indykoning:
Yes, i'm quite content with the result. With no workarounds being necessary to properly render the html!
A simple example is already provided at the top of the global slideover component: https://github.com/rapidez/core/blob/master/resources/views/components/slideover/global.blade.php

We could add docs, though we do not currently do that for other components

Jade-GG (approved):

PR #704, opened 2025-01-15
[2.x] Don't force html suffix

BobWez98:
When the `catalog/seo/product_url_suffix` or `catalog/seo/category_url_suffix` config setting is set to `null` this causes the  fallback check to return the default because `null` is a `false` value. https://github.com/rapidez/core/blob/4d362bce8cf5e8fd140b58b0315e6175240d353e/src/Models/Config.php#L62 

3.x: #705

### DIFF
--- a/src/Models/Product.php
+++ b/src/Models/Product.php
@@ -198,7 +198,7 @@ class Product extends Model
     {
         $configModel = config('rapidez.models.config');
 
-        return '/' . ($this->url_key ? $this->url_key . $configModel::getCachedByPath('catalog/seo/product_url_suffix', '.html') : 'catalog/product/view/id/' . $this->entity_id);
+        return '/' . ($this->url_key ? $this->url_key . $configModel::getValue('catalog/seo/product_url_suffix') : 'catalog/product/view/id/' . $this->entity_id);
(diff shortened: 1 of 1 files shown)


src/Models/Product.php
```diff
@@ -198,7 +198,7 @@ public function getUrlAttribute(): string
     {
         $configModel = config('rapidez.models.config');
 
-        return '/' . ($this->url_key ? $this->url_key . $configModel::getCachedByPath('catalog/seo/product_url_suffix', '.html') : 'catalog/product/view/id/' . $this->entity_id);
+        return '/' . ($this->url_key ? $this->url_key . $configModel::getValue('catalog/seo/product_url_suffix') : 'catalog/product/view/id/' . $this->entity_id);
```
royduin, line 201: `.html` is the default within Magento: https://github.com/magento/magento2/blob/6340fa15d608c91ee85b75e4f99cf239a0c8e857/app/code/Magento/Catalog/etc/config.xml#L44, our fallback should be inline with that.

royduin:
Closing in favor of https://github.com/rapidez/core/pull/736

PR #710, opened 2025-01-22
[3.x] Fix layout shift on thumbnails & quantity

Jade-GG:
This is an updated version of #675 without any duplication. I took the untested code I posted there, tested and fixed it, and made some adjustments to make the code slightly nicer.

Some notes on this:
- We cannot use the `resizedPath` helper here currently, as the URL constructor doesn't like paths that are not full URLs. This means we cannot pass something like `/catalog/product/m/p/mp07-blue_main_1.jpg.webp` (which is what we have here) into the function, it will throw an error. Should we change/fix this?
- I used `@unless` to make the `v-cloak`+`v-if` combination as clear as I could, making sure both checks have the same conditional. At the end of the day, we're making a small code quality sacrifice here to allow us to fully eliminate any content jumps.
- I minimized DOM usage a little bit for the `+3` overlay by only adding it on actual breakpoint positions (this is the `in_array($imageId + 1, $breakpoints)` check). This does, however, complicate the code further. I figured it's worth moving the overlay to a separate partial, but I would love to hear opinions on this.
- The breakpoints were weird. I've made them more in line to the page spacing available, however it's still not perfect and the thumbnails don't always line up with the big image. Ideally, someone with actual frontend skills should have a look at how the breakpoints behave and make it cleaner. (For example, on a window width between 1280px and 1536px, having more than 6 thumbnails makes the thumbnails themselves shrink)


resources/views/product/partials/images.blade.php
```diff
@@ -4,28 +4,15 @@
...
-        </div>
-    @endif
 
-    <images v-cloak>
+    <images>
         <div class="flex-1" slot-scope="{ images, active, zoomed, toggleZoom, change }">
             <div class="sticky top-5 bg-white">
                 @include('rapidez::product.partials.gallery.slider')
-                @include('rapidez::product.partials.gallery.thumbnails')
+                @includeWhen(count($product->images) > 1, 'rapidez::product.partials.gallery.thumbnails')
```
royduin, line 12: It's possible the configurable product does have 1 image, but a variant does have multiple. In that case the variant images won't show right?

resources/views/product/partials/gallery/thumbnails/showMore.blade.php
royduin, line 1: We're not using camelCase with Blade templates, `show-more` would be better.

resources/views/product/partials/gallery/thumbnails.blade.php
```diff
@@ -1,40 +1,49 @@
...
+
+    The `$breakpoints` variable determines how many images are shown on each individual breakpoint.
+--}}
+
+@php($breakpoints = ['xl' => 7, 'lg' => 5, 'md' => 7, 'sm' => 5, 'xs' => 3])
+
+<div class="mt-3 flex gap-2">
+    @for ($imageId = 0; $imageId < max($breakpoints); $imageId++)
+        <button
+            @unless ($imageId < count($product->images)) v-cloak @endunless
```
royduin, line 16: Maybe [`@attributes`](https://github.com/rapidez/blade-directives#attributes) could make this cleaner?

resources/views/product/partials/images.blade.php
Jade-GG: This is true, good point.

resources/views/product/partials/gallery/thumbnails.blade.php
Jade-GG:
```blade
@unless ($imageId < count($product->images)) v-cloak @endunless
v-if="{{ $imageId }} < images.length"
```

vs

```blade
@attributes(['v-cloak' => !($imageId < count($product->images))])
v-if="{{ $imageId }} < images.length"
```

vs

```blade
@attributes([
    'v-cloak' => !($imageId < count($product->images)),
    'v-if' => $imageId . ' < images.length',
])
```

To me the first one (original) is still the most readable. I definitely see the value in using something like `@attributes` for this but to me it feels like the code gets less readable by making it a bit more abstract.

We could maybe make a specific directive just for this sort of functionality, where you want the v-cloak and v-if combined, don't know if that's really worthwhile though.

royduin: Option 3 feels the cleanest to me, maybe ask some more?

With option 1:
- We don't use `@unless` that much
- The `v-cloak` is somewhere on that line; even with syntax highlighting you don't spot it directly

With option 3:
- You see directly all attributes
- The condition is readable

Jade-GG: I've asked the team and it looks like we have a majority vote for the 3rd option, so I'll swap it for that.

JimmyHoenderdaal (dismissed):

Roene-JustBetter (dismissed):

resources/views/product/partials/gallery/thumbnails.blade.php
```diff
@@ -1,40 +1,51 @@
...
+                @endif
+                v-bind:src="'/storage/{{ config('rapidez.store') }}/resizes/80x80/magento/catalog/product/' + images[{{ $imageId }}] + '.webp'"
+                alt="{{ $product->name }}"
+                class="block max-h-full w-auto object-contain"
+                width="80"
+                height="80"
+            />
+
+            {{-- Only include the showMore if it's actually possible to be visible --}}
+            @includeWhen(in_array($imageId + 1, $breakpoints), 'rapidez::product.partials.gallery.thumbnails.show-more')
```
indykoning, line 48: I think we might run into the same case here, where the configurable product has fewer images than the simple products do

Jade-GG: Not here, because `$imageId` is just from the loop which is based on `max($breakpoints)`, and that number is completely unrelated to how many images we actually have :)

resources/views/product/partials/gallery/slider.blade.php
```diff
@@ -1,23 +1,28 @@
...
-        <x-heroicon-o-chevron-right class="size-8 text-inactive" />
-    </button>
+
+    @if (count($product->images ?? []) > 1)
+        <button class="z-10 top-1/2 left-3 -translate-y-1/2 absolute" v-cloak v-if="active" v-on:click="change(active-1)" aria-label="@lang('Prev')">
+            <x-heroicon-o-chevron-left class="size-8 text-inactive" />
+        </button>
+        <button class="z-10 top-1/2 right-3 -translate-y-1/2 absolute" v-if="active != images.length-1" v-on:click="change(active+1)" aria-label="@lang('Next')">
+            <x-heroicon-o-chevron-right class="size-8 text-inactive" />
+        </button>
```
royduin, line 24:
```suggestion
        <button v-if="active" v-on:click="change(active-1)" class="z-10 top-1/2 left-3 -translate-y-1/2 absolute" aria-label="@lang('Prev')" v-cloak>
            <x-heroicon-o-chevron-left class="size-8 text-inactive" />
        </button>
        <button v-if="active != images.length-1" v-on:click="change(active+1)" class="z-10 top-1/2 right-3 -translate-y-1/2 absolute" aria-label="@lang('Next')">
            <x-heroicon-o-chevron-right class="size-8 text-inactive" />
        </button>
```

resources/views/product/partials/gallery/slider.blade.php
```diff
@@ -1,23 +1,28 @@
...
+    @endif
 </div>
 
-<x-rapidez::no-image v-else class="h-96 rounded" />
+<x-rapidez::no-image v-else v-cloak class="h-96 rounded" />
```
royduin, line 28:
```suggestion
<x-rapidez::no-image v-else class="h-96 rounded" v-cloak />
```

resources/views/product/partials/gallery/thumbnails.blade.php
```diff
@@ -1,40 +1,51 @@
...
-                'max-sm:flex': imageId === {{ $breakpoints['xs'] }}
+{{--
+    In this component we avoid layout shifts on the page by pre-populating the possible amount of thumbnail slots.
+    On page load all of the images come from PHP, using `v-cloak` to hide things that shouldn't be shown, but need
+    a placeholder element.
+    After Vue gets loaded in, it will take over and replace the `v-cloak`s with the `v-if`s, which means that
+    the two conditionals should be identical to avoid layout shifts.
+
+    The `$breakpoints` variable determines how many images are shown on each individual breakpoint.
+--}}
```
royduin, line 9:
```suggestion
{{--
This components initializes with data from PHP to avoid layout shifts.
When Vue is loaded it takes over and all `v-cloak`'s show up.
With `$breakpoints` you can control the amount of images.
--}}
```

resources/views/product/partials/gallery/thumbnails.blade.php
```diff
@@ -1,40 +1,51 @@
...
             }"
+            v-on:click="change({{ $imageId }})"
         >
-            <span class="size-9 flex items-center justify-center rounded-full shadow-lg shadow-default bg-white text-sm font-bold text">
-                +@{{ images.length - (imageId + 1) }}
-            </span>
-        </span>
-    </button>
+            <img
+                {{-- Note: always put the base `src` before a `v-bind:src` or it will not work --}}
```
royduin, line 36:
```suggestion
                {{-- src should always be above v-bind:src --}}
```

royduin (changes requested):

PR #711, opened 2025-01-22
Add endpoint to request a checkout with the given quote hash and customer token

indykoning:
This PR offers a secure way to share a customer token and cart mask to Rapidez in order to support a standalone checkout.

the data will be sent as POST data and then stored in the client's cookie. 
So this data will never be visible plaintext.

Next action will be creating a new Magento plugin which calls this with the data and redirecting

### DIFF
--- a/src/Http/Controllers/SignedCheckoutController.php
+++ b/src/Http/Controllers/SignedCheckoutController.php
@@ -1,0 +1,36 @@
+<?php
+
+namespace Rapidez\Core\Http\Controllers;
+
+use Illuminate\Http\Request;
+use Illuminate\Support\Facades\Cache;
+use Rapidez\Core\Models\Quote;
+
+class SignedCheckoutController
+{
+    public function __invoke(Request $request)
+    {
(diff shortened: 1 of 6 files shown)


src/Http/Controllers/SignedCheckoutController.php
```diff
@@ -0,0 +1,31 @@
...
+
+class SignedCheckoutController
+{
+    public function __invoke(Request $request)
+    {
+        if (! $request->hasValidSignature() || ! Cache::has('checkout-' . $request->get('key'))) {
+            return redirect(config('rapidez.magento_url'), 301);
+        }
+
+        $data = Cache::get('checkout-' . $request->get('key'));
```
royduin, line 17: So we store something when creating the signed route and when navigating to it we'll get it. What if in the meantime the cache will be flushed?

royduin: Can't we store it somewhere persistent?

indykoning: I've added docs on using the standalone checkout, and configuring more persistent storage: https://github.com/rapidez/docs/pull/74

PR #716, opened 2025-01-24
[Instantsearch] Update selected refinements

Jade-GG:
By matching the items to the actual filter data we can:
- Check if it's a boolean and show the proper label
- Get the actual filter name so we don't have to show the filter code

I've also updated the "clear filters" to match the old one, and used v-show instead of v-if to not need the empty div in the v-else.

afbeelding

### DIFF
--- a/resources/views/listing/partials/filter/selected.blade.php
+++ b/resources/views/listing/partials/filter/selected.blade.php
@@ -2,7 +2,7 @@
     <template v-slot="{ canRefine, refine, createURL }">
         <div v-show="canRefine" class="flex flex-wrap items-baseline justify-between gap-2 w-full pb-2">
             <div class="font-semibold text-base">
-                @lang('You have filtered for:')
+                @lang('Selected filters:')
             </div>
             <a
                 v-bind:href="createURL()"
--- a/resources/views/listing/partials/filter/selected.blade.php
+++ b/resources/views/listing/partials/filter/selected.blade.php
@@ -35,10 +35,9 @@
                         class="flex items-center gap-1 p-1 text-xs rounded-lg bg transition hover:opacity-80"
                     >
                         @{{ item.filter.name }}:
-                        <template v-if="item.filter.input === 'boolean'">
-                            <template v-if="refinement.label == '1'">@lang('Yes')</template>
-                            <template v-if="refinement.label == '0'">@lang('No')</template>
-                        </template>
+                        <template v-if="false"></template>
+                        @include('rapidez::listing.partials.filter.selected.boolean')
+                        @include('rapidez::listing.partials.filter.selected.swatch')
                         <template v-else>
                             @{{ refinement.label }}
                         </template>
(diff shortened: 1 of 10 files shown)


resources/views/listing/partials/filter/selected.blade.php
```diff
@@ -28,61 +30,26 @@ class="flex flex-wrap gap-2 relative"
...
-                            Having the label here is useful when filtering booleans,
-                            but the item label is currently the "code".
-                            --}}
-                            @{{ item.label }}:
-                            {{-- When the label is a boolean a yes/no would be nice --}}
+                        @{{ item.filter.name }}:
+                        <template v-if="item.filter.input === 'boolean'">
+                            <template v-if="refinement.label == '1'">@lang('Yes')</template>
+                            <template v-if="refinement.label == '0'">@lang('No')</template>
+                        </template>
```
indykoning, line 41: This will probably also happen for the swatches (i think showing the actual swatch if possible is nicer)
So maybe it's a good idea to split this up like we have done with the filters themselves: https://github.com/rapidez/core/blob/77cc7f9c116449af8dfa39b699e490264f124295/resources/views/listing/filters.blade.php#L10

resources/views/listing/partials/filter/selected.blade.php
```diff
@@ -1,22 +1,24 @@
...
-            v-on:click.prevent="refine"
-        >
-            @lang('Reset filters')
-        </a>
-
-        {{-- TODO: Without this the default renders... --}}
-        <div v-else></div>
+        <div v-show="canRefine" class="flex flex-wrap items-baseline justify-between gap-2 w-full pb-2">
+            <div class="font-semibold text-base">
+                @lang('You have filtered for:')
```
indykoning, line 5: I can't put my finger on it, i feel like this reads a bit weird. 
```suggestion
                @lang('Selected filters:')
```
or
```suggestion
                @lang('Current filters:')
```
may read a bit nicer

Jade-GG: Looks like we overwrite this in every single project to be without this label, and just show "reset filters".

But, if I were to choose, I'd pick "Selected filters:" here.

resources/views/listing/partials/filter/selected.blade.php
Jade-GG: Could be done with something like this:
```blade
<template v-if="false">
@include('rapidez::listing.partials.filter.selected.boolean')
@include('rapidez::listing.partials.filter.selected.swatch')
<template v-else>
    @{{ refinement.label }}
</template>
```

We do then want the `v-else` here so we would need the `<template v-if="false">`.

Definitely cleaner if we end up having more exceptions. Would this be the nicest approach though?

indykoning (approved):

resources/views/listing/partials/filter/selected.blade.php
```diff
@@ -1,22 +1,24 @@
...
-            v-on:click.prevent="refine"
-        >
-            @lang('Reset filters')
-        </a>
-
-        {{-- TODO: Without this the default renders... --}}
-        <div v-else></div>
+        <div v-show="canRefine" class="flex flex-wrap items-baseline justify-between gap-2 w-full pb-2">
+            <div class="font-semibold text-base">
+                @lang('Selected filters:')
```
royduin, line 5:
```suggestion
                @lang('Selected filters'):
```

PR #717, opened 2025-01-29
[Instantsearch] Add key to item

Jade-GG:
Selected options stayed selected after filtering and getting a different item in its place. Adding a key here fixes that.

### DIFF
--- a/resources/views/listing/partials/item.blade.php
+++ b/resources/views/listing/partials/item.blade.php
@@ -1,4 +1,4 @@
-<div class="size-full">
+<div class="size-full" v-bind:key="item.entity_id">
     <add-to-cart v-bind:product="item" v-slot="addToCart" v-cloak>
         <div class="group relative flex flex-1 flex-col rounded border bg-white p-5" :key="item.entity_id">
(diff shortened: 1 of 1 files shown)


indykoning (approved):

resources/views/listing/partials/item.blade.php
```diff
@@ -1,4 +1,4 @@
-<div class="size-full">
+<div class="size-full" v-bind:key="item.entity_id">
     <add-to-cart v-bind:product="item" v-slot="addToCart" v-cloak>
         <div class="group relative flex flex-1 flex-col rounded border bg-white p-5" :key="item.entity_id">
```
royduin, line 3: This key should be removed I think?

Jade-GG: You're right, fixed.

royduin:
Is this also a problem on master?

Jade-GG:
Just checked, looks like it is. Go to [this category](https://test.rapidez.io/men/tops-men/hoodies-and-sweatshirts-men.html), select a color option on the first one, and then filter for "wool".

PR #718, opened 2025-01-29
[Instantsearch] Swatch filter

Jade-GG:
I've kept the `useSwatches` stuff here as there isn't really any other way to do this cleanly.

We ran into this before elsewhere recently and looked more deeply into what options we have and there really isn't much. Basically two options:
- We keep it as is, getting the swatch data with an API call and using it here. This is clean, but it requires extra functionality for something that seems like it should be quite simple.
- We could add the swatch data into the label itself and then filter it out on the frontend, which would allow us to keep all the data within ES. This would work, but it's not very clean and quite hacky.

### DIFF
--- a/resources/views/listing/partials/filter/selected/swatch.blade.php
+++ b/resources/views/listing/partials/filter/selected/swatch.blade.php
@@ -1,5 +1,5 @@
 {{-- TODO: Display swatches correctly here --}}
 
-<template v-else-if="item.filter.input === 'swatch'">
-    @{{ refinement.label }}
+<template v-else-if="item.filter.super">
+    @{{ refinement.swatch.label }}
 </template>
--- a/resources/views/listing/partials/filter/selected/swatch.blade.php
+++ b/resources/views/listing/partials/filter/selected/swatch.blade.php
@@ -1,0 +1,5 @@
+{{-- TODO: Display swatches correctly here --}}
+
+<template v-else-if="item.filter.input === 'swatch'">
+    @{{ refinement.label }}
+</template>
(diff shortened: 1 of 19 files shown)


resources/views/listing/partials/filter/selected/swatch.blade.php
```diff
@@ -1,5 +1,5 @@
 {{-- TODO: Display swatches correctly here --}}
 
-<template v-else-if="item.filter.input === 'swatch'">
-    @{{ refinement.label }}
+<template v-else-if="item.filter.super">
+    @{{ refinement.swatch.label }}
```
indykoning, line 4: Currently it is possible for swatch.label to be null due to https://github.com/rapidez/core/pull/718/files#diff-862a032d6992ba9a24c4278dd9a2ff425bfadcef76f9f721673d86967b5175f8R245

What is the best way to solve this? Use `refinement.label` as a fallback?

resources/views/listing/partials/filter/swatch.blade.php
```diff
@@ -1,37 +1,43 @@
...
+    <template v-slot="{ items, refine }">
+        <div v-show="items.length" class="relative pb-4">
+            <x-rapidez::filter.heading>
+                <ul class="flex flex-wrap gap-2 items-center my-1">
+                    <li v-for="item in withSwatches(items, filter)">
+                        <label
+                            v-if="filter.visual_swatch"
+                            class="size-6 ring-1 ring-emphasis/10 ring-inset cursor-pointer flex items-center justify-center hover:opacity-75 rounded-full transition"
+                            v-bind:class="{ 'outline outline-2 outline-emphasis outline-offset-1': item.isRefined }"
+                            v-bind:style="{ background: item.swatch.swatch }"
```
indykoning, line 16: Currently it is possible for swatch.swatch to be null due to https://github.com/rapidez/core/pull/718/files#diff-862a032d6992ba9a24c4278dd9a2ff425bfadcef76f9f721673d86967b5175f8R245

We might want to add a check for this

indykoning (changes requested):

PR #722, opened 2025-01-31
Added running state to graphql component

indykoning:
This PR adds a boolean value as was already available for graphql mutation components to determine wether the component is currently running a query.

This can be used to show a loader or placeholder.

I've also added a computed property to the graphql-mutation for consistency sake

### DIFF
--- a/resources/js/components/GraphqlMutation.vue
+++ b/resources/js/components/GraphqlMutation.vue
@@ -1,5 +1,4 @@
 <script>
-import { computed } from 'vue';
 import { GraphQLError, combiningGraphQL, magentoGraphQL } from '../fetch'
 
 export default {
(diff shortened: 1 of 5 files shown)


resources/js/components/GraphqlMutation.vue
```diff
@@ -199,5 +199,10 @@ export default {
             })
         },
     },
+    computed: {
+        running() {
+            return this.mutating
+        },
```
royduin, line 205: Maybe better to turn this around; change it to `running` everywhere for consistency and keep a `mutating` computed to keep it backwards compatible? So when we like to drop this "backwards compatible fix" we can just remove the computed.

indykoning: Agreed, i've updated the PR

PR #723, opened 2025-02-06
Remove custom shadow colors

Roene-JustBetter:
Removed the shadow from the minicart as it looks a bit strange together with the border. 

**Before:** 
Scherm­afbeelding 2025-02-06 om 10 39 43

**After:** 
Scherm­afbeelding 2025-02-06 om 10 35 49

### DIFF
--- a/resources/views/checkout/partials/progressbar.blade.php
+++ b/resources/views/checkout/partials/progressbar.blade.php
@@ -6,11 +6,12 @@
                 <div class="absolute flex w-full h-0.5 top-5 left-1/2 {{ $currentStepKey > $checkoutStepKey ? 'bg-emphasis' : 'bg' }}"></div>
             @endif
             <a href="{{ route('checkout', $checkoutStep) }}" @class([
-                'relative flex size-10 mx-auto justify-center rounded-full font-bold items-center border',
+                'relative flex size-10 mx-auto justify-center rounded-full font-bold items-center',
                 'pointer-events-none cursor-default' => $currentStepKey < $checkoutStepKey,
-                'bg-emphasis' => $checkoutStepKey <= $currentStepKey,
-                'bg-white' => $checkoutStepKey > $currentStepKey,
-                'bg-emphasis shadow-md shadow-emphasis pointer-events-none cursor-default' => $checkoutStepKey === $currentStepKey
+                'bg-secondary' => $checkoutStepKey < $currentStepKey,
+                'text-white' => $checkoutStepKey <= $currentStepKey,
+                'bg-white border' => $checkoutStepKey > $currentStepKey,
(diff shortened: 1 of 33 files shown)


royduin:
Maybe double check all uses of shadow, how was this before? A bit of shadow *could* be nice. Check with @jordythevulder?

Roene-JustBetter:
> Maybe double check all uses of shadow, how was this before? A bit of shadow _could_ be nice. Check with @jordythevulder?

I will check it with @jordythevulder the issue with the shadow is that the colors we used are to dark. I have updated the colors to lighter variants in the color palette. 

New shadow color:
Scherm­afbeelding 2025-02-10 om 10 02 42

Old shadow color: 
Scherm­afbeelding 2025-02-10 om 10 08 47

royduin:
- The colors also need to be changed at https://github.com/rapidez/blade-components
- Do we also need a shadow on the menu dropdown?
- Please double check with @jordythevulder on all places where it's used or where it could/should be used

resources/views/layouts/partials/header.blade.php
```diff
@@ -1,4 +1,4 @@
-<header class="relative mb-5 max-md:pb-4 border-b z-header">
+<header class="relative mb-5 max-md:pb-4 border-b shadow bg/50 z-header">
```
royduin, line 1: Where is `bg/50` coming from?

tailwind.config.js
```diff
@@ -82,6 +76,8 @@ export default {
 
                 'popup': '130',
                 'popup-actions': '10',
+
+                'cookie': '140',
```
royduin, line 80: This should go in a separated PR

resources/views/layouts/partials/header.blade.php
Roene-JustBetter: Removed it doesn't need it here.

resources/views/checkout/partials/progressbar.blade.php
Roene-JustBetter, line 1: Scherm­afbeelding 2025-03-05 om 10 47 39
Checked this with @jordythevulder

tailwind.config.js
Roene-JustBetter: https://github.com/rapidez/core/pull/761 Removed it here and created another Pull request.

royduin:
We're removing the shadow colors as they're not changed that often. The problem with this implementation was that there wasn't a opacity defined, resulting in ugly big shadows. 

We do have a Tailwind play "playground" here: https://play.tailwindcss.com/lyreLA2s54?file=config.

If you do need to change the shadow; we recommend to change the whole box shadows, not the shadow colors as this isn't a separated thing in CSS. If you just need to change some shadow colors; use the default Tailwind utilities.

PR #726, opened 2025-02-07
[2.x] Get option labels with attribute options in the cart

BobWez98:
This PR is a solution to not having option labels in the cart attributes. Sadly the `customAttributeMetadata` query does not support getting the option labels, however `customAttributeMetadataV2` does. We also had to change the mapping a bit.

3.x: #755

### DIFF
--- a/resources/js/stores/useAttributes.js
+++ b/resources/js/stores/useAttributes.js
@@ -1,5 +1,6 @@
 // TODO: In this file there is a lot of duplication compared to useSwatches. Can we improve that?
 import { computedAsync, useLocalStorage } from '@vueuse/core'
+import { computed } from 'vue'
 
 export const attributesStorage = useLocalStorage('attributes', {})
 let isRefreshing = false
@@ -37,8 +38,8 @@ export const attributes = computedAsync(
     { lazy: true, shallow: false },
 )
 
-window.attributeLabel = (attributeCode) => {
+window.attributeLabel = computed((attributeCode) => {
     return Object.values(attributes.value)?.find((attribute) => attribute.code === attributeCode)?.name
-}
+})
 
 export default () => attributes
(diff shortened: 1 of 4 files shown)


resources/views/cart/item.blade.php
```diff
@@ -27,7 +27,7 @@ class="mx-auto"
                     </div>
                     <div v-for="option in config.cart_attributes">
                         <template v-if="item.product.attribute_values?.[option] && typeof item.product.attribute_values[option] === 'object'">
-                            @{{ option }}: <span v-html="item.product.attribute_values[option]?.join(', ')"></span>
+                            @{{ item.product.attribute_values[option].label }}: <span v-html="item.product.attribute_values[option].options?.join(', ')"></span>
```
indykoning, line 30: Might placing the label here into a separate variable simplify the code?
We're memoizing the result in fetchAttributeValues, so we could put it into a computedAsync and get the label straight in the frontend here. Instead of from the product

BobWez98:
Updated PR

resources/js/stores/useAttributes.js
```diff
@@ -37,4 +37,8 @@ export const attributes = computedAsync(
     { lazy: true, shallow: false },
 )
 
+window.attributeLabel = (attributeCode) => {
+    return Object.values(attributes.value)?.find((attribute) => attribute.code === attributeCode)?.name
```
indykoning, line 41: I think we might want to return a computed value of this expression.
attributes is a lazy computed async. It will only be fetched when it's used and returns a default value in the meantime.

calling `attributeLabel` might give us undefined back while the request is still going and it will not be able to update.
if we wrap this in a computed function it has the chance to rerender with the updated attributes

resources/js/stores/useAttributes.js
```diff
@@ -37,4 +38,8 @@ export const attributes = computedAsync(
     { lazy: true, shallow: false },
 )
 
+window.attributeLabel = computed((attributeCode) => {
```
indykoning, line 41: If this works it's fine, i'd expect the return value would need to be a computed value instead of the function itself for it to work properly

indykoning (approved):

PR #727, opened 2025-02-10
[2.x] Make sure you can't place two orders at the same time

Jade-GG:
If you click the "Place order" button twice in a very short timespan (e.g. your mouse is broken and double clicks the button) we have seen a very rare edge case where this seemingly actually causes an order to be placed twice. I was not able to reproduce this locally, however I do believe that's the actual root cause.

Originally I made [a patch](https://github.com/rapidez/checkout-theme/pull/165) that caused the button to disable instantly on click. However, it seems that didn't actually fix the issue and caused a race condition to rarely happen that disabled the button without submitting.

Instead, here's an alternative solution to only fix the double orders thing.

### DIFF
--- a/resources/js/components/Checkout/Checkout.vue
+++ b/resources/js/components/Checkout/Checkout.vue
@@ -10,6 +10,7 @@ export default {
 
     data() {
         return {
+            placingOrder: false,
             backEvent: false,
             steps: [],
         }
@@ -211,6 +212,12 @@ export default {
                 return false
             }
 
+            if (this.placingOrder) {
+                return false
+            }
+
+            this.placingOrder = true
+
             try {
                 this.$root.$emit('before-checkout-payment-saved', {
                     order: {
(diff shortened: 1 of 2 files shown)


royduin:
- This isn't actually used now?
- Do we also need this on other buttons? Like the add to cart button?

Jade-GG:
> * This isn't actually used now?

Do you mean it's not used in Rapidez v3? Because yeah, the graphql component is used for the new checkout. We could add a check like this on the graphql components though, depending on whether this is also a problem in v3 (I did not test this). If you want I can make that update.

> * Do we also need this on other buttons? Like the add to cart button?

Possibly, although I wouldn't say we'd *need* it. Stuff like the add to cart is less problematic if it gets clicked twice because you can always just fix the quantity in your cart later.

Jade-GG:
> * Do we still need a disabled/loading state? Some visual feedback would be nice.

The disabled state will be set a few ticks later by the default `loading` state of Rapidez (which automatically gets set by any `fetch.js` function) so it is there 🙂 

> * Could you also check this in v3?

Sure, I'll put it on my to-do list for this sprint.

> * If it happens in the checkout most likely it will also be the case with the cart

Possibly, that'll depend on the graphql component (and the impact would probably be limited). This is what I'll be checking for the v3 version, so if I fix it there I will also fix that here.

Jade-GG:
https://github.com/rapidez/core/pull/740

royduin (approved):
Looking good to me, let's wait for @indykoning his review.

indykoning (approved):

PR #728, opened 2025-02-10
Remove no-cache from cache-control

indykoning:
By default Symfony (and thus by extension Laravel) returns no-cache, private as the cache-control headers.

https://github.com/symfony/symfony/blob/6f4f04b98f6134996a57d21d1851d558a34dc45c/src/Symfony/Component/HttpFoundation/ResponseHeaderBag.php#L251

This causes Turbo to skip loading from cache: 

https://github.com/hotwired/turbo/blob/ae14a60e7a5784deeb54cadeb47778693855c57f/src/core/drive/page_snapshot.js#L66

Using the Etag and/or last_modified date (which we have available for categories, pages and products) allows pages to be cached but always needing to be revalidated:
https://github.com/symfony/symfony/blob/6f4f04b98f6134996a57d21d1851d558a34dc45c/src/Symfony/Component/HttpFoundation/ResponseHeaderBag.php#L247

This will allow turbo to use the prefetched page instead of having to fetch the page again before fully showing.

As you can see we first have a request to prefetch the page (this is when i hover the link)
Then after the click it immediately shows the page (as seen by the render function hovered)
~~After the page has fully loaded the revalidation request will trigger.~~ That was incorrect, this was only caused by my mouse still being hovered over the link.
image

Full caching would be even better however this could result in stale data, or data from another user being shown in projects where this is rendered in PHP.
This is better left up to the developer and their project by implementing a middleware, like https://laravel.com/docs/11.x/responses#cache-control-middleware

I believe this PR is able to be backported to 2.x, if we want this i'll create a PR for that as well.

### DIFF
--- a/src/Http/Controllers/CategoryController.php
+++ b/src/Http/Controllers/CategoryController.php
@@ -17,7 +17,6 @@ class CategoryController extends Controller
 
         $response = response()->view('rapidez::category.overview', compact('category'));
         $response->setCache(['etag' => md5($response->getContent() ?? ''), 'last_modified' => $category->updated_at]);
-        $response->isNotModified(request());
 
         return $response;
     }
--- a/src/Http/Controllers/CategoryController.php
+++ b/src/Http/Controllers/CategoryController.php
@@ -16,8 +16,9 @@ class CategoryController extends Controller
         session(['latest_category_path' => $category->path]);
 
         $response = response()->view('rapidez::category.overview', compact('category'));
-        $response->setCache(['etag' => md5($response->getContent() ?? ''), 'last_modified' => $category->updated_at]);
 
-        return $response;
+        return $response
+            ->setEtag(md5($response->getContent() ?? ''))
+            ->setLastModified($category->updated_at);
     }
 }
--- a/src/Http/Controllers/CategoryController.php
+++ b/src/Http/Controllers/CategoryController.php
@@ -2,9 +2,7 @@
 
 namespace Rapidez\Core\Http\Controllers;
 
-use Illuminate\Routing\Controller;
-
-class CategoryController extends Controller
+class CategoryController
 {
     public function show(int $categoryId)
     {
--- a/src/Http/Controllers/CategoryController.php
+++ b/src/Http/Controllers/CategoryController.php
@@ -2,7 +2,9 @@
 
 namespace Rapidez\Core\Http\Controllers;
 
-class CategoryController
+use Illuminate\Routing\Controller;
+
+class CategoryController extends Controller
 {
     public function show(int $categoryId)
     {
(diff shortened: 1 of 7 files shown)


src/Models/Category.php
```diff
@@ -12,6 +12,11 @@ class Category extends Model
 {
     use HasAlternatesThroughRewrites;
 
+    protected $casts = [
+        self::UPDATED_AT => 'datetime',
```
indykoning, line 16: note: i'm using self instead of static as static cannot be used since it's created at runtime and $casts is compiled earlier.

src/Http/Controllers/CategoryController.php
```diff
@@ -13,6 +15,9 @@ public function show(int $categoryId)
         config(['frontend.subcategories' => $category->subcategories->pluck('name', 'entity_id')]);
         session(['latest_category_path' => $category->path]);
 
-        return view('rapidez::category.overview', compact('category'));
+        $response = response()->view('rapidez::category.overview', compact('category'));
+        $response->setCache(['etag' => md5($response->getContent() ?? ''), 'last_modified' => $category->updated_at]);
+
+        return $response;
```
royduin, line 21: Can we write this cleaner like this?
```suggestion
        return response()
            ->view('rapidez::category.overview', compact('category'));
            ->setCache([
                'etag' => md5($response->getContent() ?? ''),
                'last_modified' => $category->updated_at
            ]);
```

Or maybe even better:

```suggestion
        return response()
            ->view('rapidez::category.overview', compact('category'));
            ->setEtag(md5($response->getContent()))
            ->setLastModified($category->updated_at);
```

royduin:
Any drawbacks of this? What if we're [displaying content from Statamic on a product/category page](https://docs.rapidez.io/3.x/packages/statamic.html#displaying-content)? The last modified date could be something else. We could make this extendable with for example Eventy, but maybe it's easier to have an accessor on the models for the updated dates? And what about global changes?

indykoning:
I do agree on allowing an accessor or eventy filters, other than that we don't need to worry much:
This method only gets used if Etags aren't supported:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Last-Modified

And it's mostly used to remove the explicit `no-cache` value and provide info to reduce crawler spam (if they listen to it)

src/Http/Controllers/CategoryController.php
indykoning: I can call the setEtag and setLastModified directly instead of using setCache.
Unfortunately due to the `setEtag` requiring the response content i cannot chain it onto the response()->viev()

royduin: Ah I see, maybe this:

```suggestion
        $response = response()->view('rapidez::category.overview', compact('category'));

        return $response
            ->setEtag(md5($response->getContent()))
            ->setLastModified($category->updated_at);
```

royduin: But it's pretty repeated, maybe we can extract it somehow?

PR #706, opened 2025-01-15
InstantSearch migration

royduin:
[ReactiveSearch](https://www.reactivesearch.io/) is within Rapidez from day 1, but the current status is unclear and there [isn't much activity on Github](https://github.com/appbaseio/reactivesearch/graphs/contributors). We choose this as Magento requires [Elasticsearch](https://github.com/elastic/elasticsearch) / [OpenSearch](https://github.com/opensearch-project/OpenSearch), so no additional requirements.

In the past we've worked with [InstantSearch](https://github.com/algolia/instantsearch) from [Algolia](https://www.algolia.com) which supports more search engines which is a good alternative and it integrates with multiple search engines:

- [Meilisearch](https://www.meilisearch.com/) ([plugin](https://github.com/meilisearch/meilisearch-js-plugins))
- [Typesense](https://typesense.org/) ([adapter](https://github.com/typesense/typesense-instantsearch-adapter))
- [Elasticsearch](https://www.elastic.co/elasticsearch) ([searchkit](https://www.searchkit.co/))

With this migration; we could migrate to [Laravel Scout](https://laravel.com/docs/11.x/scout) so we can get rid of our custom indexer, use a first party Laravel package, get batch imports out-of-the-box and more advantages. Laravel Scout also supports multiple search engines; Algolia, Meilisearch and Typesense out-of-the-box. There are some custom drivers for Elasticsearch:

- [matchish/laravel-scout-elasticsearch](https://github.com/matchish/laravel-scout-elasticsearch)
- [Jeroen-G/Explorer](https://github.com/Jeroen-G/Explorer)
- [babenkoivan/elastic-scout-driver](https://github.com/babenkoivan/elastic-scout-driver)

This current proof of concept is using Meilisearch, but I'm not yet convinced as locally (on a M1) I can't get it faster than 300ms where Elasticsearch responds within 50ms. Not sure yet how easy it is to switch between engines; maybe it will be configurable with Elasticsearch as default; so you can use the same infrastructure.

To be continued 🚀 

Oh and yes; we're also going to migrate to Vue 3; but after this within a separated PR.

---

**Update: Switched to Elasticsearch using Searchkit**. Performance is back to what we're used to and within the core (but also within some projects) custom query DSL is used; so sticking to Elasticsearch is preferred.

PR #734, opened 2025-02-14
[Instantsearch] Scout indexing products & categories

Jade-GG:
Requires [this patch](https://github.com/matchish/laravel-scout-elasticsearch/pull/298) for the products to index properly. Without it, only the first 500 products get indexed.

This PR introduces a generic version of the `Searchable` trait that we can put onto any Eloquent model to make it indexable by Scout, as well as some generic functions that can be hooked into to filter/query the data to your liking. This also replaces the indexer commands with one generic `IndexCommand`.

A few notes:

- The base Rapidez models will automatically get detected as a Scout model---any others need to be added into the `indexer.php` config file under `extra_models`. Is this the best way to go?
- I've kept function names as is, however it might be useful to change them. For example, `makeAllSearchableUsing()` might be better if we call a new function inside it called something like `searchableQuery()`
- We don't have some sort of initialization function before indexing, so I've put the `maxPositions` and `categories` variables for the Product model into Cache with the `array` driver. There may be a cleaner way to do this.
- I didn't touch any of the actual code that was already inside of the IndexProductsCommand for now to make sure we keep feature parity, but the `withCategories` function can probably be refactored?
- The `Category` model had a weird custom primary key name set which broke Scout. I couldn't find anything that this was actually used for, so I restored it back to what it should be.
- I've put the index prefix inside of the ConfigComposer's `getConfig()` function. I'm not sure if this is the best way to do this.


src/Commands/IndexCommand.php
```diff
@@ -0,0 +1,51 @@
...
+
+        IndexBeforeEvent::dispatch($this);
+
+        foreach ($stores as $store) {
+            Rapidez::setStore($store);
+
+            $this->line('Store: ' . $store['name']);
+
+            foreach ($types as $type => $model) {
+                config()->set('rapidez.index', $type);
```
indykoning, line 41: Is it possible to infer this from the model itself? It would save us from setting this value which thanks to your changes doesn't seem needed anymore.

Perhaps a function that reads from an empty model variable/constant, which if it's not set [Str::snake()](https://laravel.com/docs/master/strings#method-snake-case) on the class name

royduin: That's also an option; currently we do this differently on multiple places, for example:

- The category index is defined here: https://github.com/rapidez/core/pull/735/files#diff-694b39ded25e81aa9cfa9de2b404533c0f924a4919b0f68b38aebfb0741ff68fR1 which is using this config: https://github.com/rapidez/core/pull/735/files#diff-fe38fec1955243f9e20c1cdc08b64167096d11ee224c3c86fc8e4f4671750a51R64
- For the products we've this: https://github.com/rapidez/core/pull/706/files#diff-b30cee12bf5e6816f22522e4ab112a4dc162789811fdcd04c7df321dcb438dbbR132 which can be used in the frontend with `config.index`

resources/views/components/listing.blade.php
```diff
@@ -14,7 +14,7 @@
             label: window.config.translations.newest,
             field: 'created_at',
             order: 'desc',
-            value: config.index+'_created_at_desc',
+            value: config.index_prefix + '_products_' + config.store + '_created_at_desc',
```
royduin, line 17:
```suggestion
            // TODO: Extract this somewhere?
            value: config.index_prefix + '_products_' + config.store + '_created_at_desc',
```

src/Models/Product.php
```diff
@@ -253,4 +256,103 @@ public static function exist($productId): bool
...
+
+        $data['positions'] = $product->categoryProducts
+            ->pluck('position', 'category_id')
+            // Turn all positions positive
+            ->mapWithKeys(fn ($position, $category_id) => [$category_id => $maxPositions[$category_id] - $position]);
+
+        return $data;
+    }
+
+    public function withCategories(array $data): array
```
royduin, line 310:
```suggestion
    // TODO: With this and all methods above this model
    // is getting pretty big... can we move this?
    public function withCategories(array $data): array
```

src/Models/Traits/Searchable.php
```diff
@@ -0,0 +1,60 @@
...
+
+    public function searchableAs(): string
+    {
+        if (! config('rapidez.index')) {
+            throw new Exception('Do not use Scout directly. Please use `php artisan rapidez:index`.');
+        }
+
+        return implode('_', array_values([
+            config('scout.prefix'),
+            config('rapidez.index'),
```
royduin, line 46:
```suggestion
            // TODO: Maybe use the model name here?
            config('rapidez.index'),
```

PR #736, opened 2025-02-20
[2.x] Only apply default value when config option is not found

Jade-GG:
When retrieving a config option you may end up getting a value of `NULL`. The current implementation uses null-coalescing after getting the value, which causes that `NULL` to be replaced by the default value. However, if the config key does actually exist then this is not the correct behavior.

For example, if you don't want a `.html` suffix on your product pages, you set the config value to `NULL`. This then gets converted back into `.html` when indexing products, causing the wrong urls to be indexed.

This PR fixes that by implementing a `default` option into the `getValue` function and using that instead of the null-coalescing operator.

### DIFF
--- a/src/Models/Config.php
+++ b/src/Models/Config.php
@@ -21,6 +21,10 @@ class Config extends Model
 
     protected $primaryKey = 'config_id';
 
+    const CREATED_AT = null;
+
+    protected $guarded = [];
+
     protected static function booting()
     {
         static::addGlobalScope('scope-fallback', function (Builder $builder) {
--- a/src/Models/Config.php
+++ b/src/Models/Config.php
@@ -98,6 +98,9 @@ class Config extends Model
             // Catch the case it is intentionally set to null
             if (Arr::has($configCache, $cacheKey)) {
                 $result = Arr::get($configCache, $cacheKey);
+                if ($result === false) {
+                    $result = $options['default'] ?? null;
+                }
 
                 return (bool) $options['decrypt'] && is_string($result) ? static::decrypt($result) : $result;
             }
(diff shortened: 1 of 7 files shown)


royduin (approved):
Looking good to me, let's wait for @indykoning his review. Maybe a nice follow-up on this: RAP-1193

src/Models/Config.php
```diff
@@ -101,16 +105,17 @@ public static function getValue(
...
             ->where(fn ($query) => $query
                 ->when($scope === ConfigScopes::SCOPE_STORE, fn ($query) => $query->whereStore($scopeId))
                 ->when($scope !== ConfigScopes::SCOPE_DEFAULT, fn ($query) => $query->orWhere(fn ($query) => $query->whereWebsite($websiteId)))
                 ->orWhere(fn ($query) => $query->whereDefault())
             )
-            ->first('value')
-            ?->value;
+            ->first('value');
+
+        $result = $resultObject ? $resultObject->value : ($options['default'] ?? null);
```
indykoning, line 118: I think this could cause trouble with the cache function, in case the default value should differ.
Maybe force `$options['cache']` to false if the result did not come from the DB

Jade-GG: I don't know if I necessarily like abusing a boolean here to differentiate between the cases, as it feels a bit arbitrary and risky when it comes to possible future refactorings, but I do get the issue.

I'm trying to think what would be a better option though, in a lot of cases I can imagine we actually just want to eventually refactor the defaults to already exist within Rapidez so you don't have to specify them yourself 🤔

indykoning: A separate local variable is fine too. We just cannot cache the passed default value as it is not in-line with the actual data in the database. 
Caching it would mean we remove the choice of a differing default value.

Jade-GG: For now I've just accepted that caching `false` is the best solution, because we do still want to actually cache the config path even if the actual value was not found.

PR #737, opened 2025-02-20
[3.x] Only apply default value when config option is not found

Jade-GG:
See #736


royduin (dismissed):
Looking good to me, let's wait for @indykoning his review. Maybe a nice follow-up on this: RAP-1193

PR #738, opened 2025-02-20
[3.x] Region select

Jade-GG:
In Magento, for some countries it may be required to add a region (this is customizable in the configuration). Because it is possible---and we ran into a case where it's necessary---I have implemented a basic version here.

Because of the way we structured input components, it seemed cleanest to me to use a `has` selector to hide the region select when there are no regions. This is, however, only supported in browsers from 2023 and later. Older browsers will see an empty "region" block, which is not breaking. Also requires a tailwindcss version bump (from ^3.3.3 to ^3.4).

This will also have to be added into checkout-theme (and confira).


resources/views/components/input/select/region.blade.php
```diff
@@ -0,0 +1,21 @@
+@props(['country' => ''])
```
royduin, line 1: I think this is a required prop?

```suggestion
@props(['country'])
```

Having an empty `v-if` doesn't make any sense.

resources/views/components/input/select/region.blade.php
```diff
@@ -0,0 +1,21 @@
...
+    query="query country($id: String) { country(id: $id) { available_regions { code id name } } }"
+    v-bind:variables="{ id: {{ $country }} }"
+    v-bind:cache="'regions_' + {{ $country }}"
+    v-bind:key="{{ $country }}"
+>
+    <div class="contents" slot-scope="{ data }">
+        <template v-if="data && data.country.available_regions">
+            <x-rapidez::input.select
+                {{ $attributes }}
+                v-bind:required="'{{ Rapidez::config('general/region/state_required', '') }}'.split(',').includes({{ $country }})"
```
royduin, line 13: No need for defining a default when there isn't one?

```suggestion
                v-bind:required="'{{ Rapidez::config('general/region/state_required') }}'.split(',').includes({{ $country }})"
```

Jade-GG: I did this here just to make sure you don't get a stringified `null` in there, although I just checked and `null` becomes an empty string anyway, including when passed through the `e()` function.

resources/views/checkout/partials/address.blade.php
```diff
@@ -145,14 +145,30 @@
...
+        <div class="col-span-12 sm:col-span-6 has-[.exists]:block hidden">
+            <label>
+                <x-rapidez::label>@lang('Region')</x-rapidez::label>
+                <x-rapidez::input.select.region
+                    class="exists"
+                    name="{{ $type }}_region"
+                    dusk="{{ $type }}_region"
+                    country="variables.country_code"
+                    v-model="variables.region_id"
                     required
```
indykoning, line 172: Perhaps we'll want to listen to 
image
for the required attribute
`general/region/state_required`
and
`general/region/display_all`

resources/views/components/input/select/region.blade.php
```diff
@@ -10,6 +10,7 @@
         <template v-if="data && data.country.available_regions">
             <x-rapidez::input.select
                 {{ $attributes }}
+                v-if="'{{ Rapidez::config('general/region/state_required') }}'.split(',').includes({{ $country }}) || {{ Rapidez::config('general/region/display_all', '0') }}"
```
indykoning, line 13:
```suggestion
                v-if="{{ Rapidez::config('general/region/display_all', '0') }} || '{{ Rapidez::config('general/region/state_required') }}'.split(',').includes({{ $country }})"
```
Putting the simpler check first may allow us to skip the more complex second check

indykoning (dismissed):

royduin:
We also need this within:

- rapidez/account
- rapidez/checkout-theme
- rapidez/confira

Is there a follow-up? Should we merge this one first?

Jade-GG:
> We also need this within:
> 
>     * rapidez/account
> 
>     * rapidez/checkout-theme
> 
>     * rapidez/confira
> 
> 
> Is there a follow-up? Should we merge this one first?

Correct, I will follow up on these after this one has been approved. Looks like it has been approved now, so expect these pull requests soon™

royduin (dismissed):

resources/views/components/input/select/region.blade.php
```diff
@@ -0,0 +1,22 @@
...
+    v-bind:key="{{ $country }}"
+>
+    <div class="contents" slot-scope="{ data }">
+        <template v-if="data && data.country.available_regions">
+            <x-rapidez::input.select
+                {{ $attributes }}
+                v-if="{{ Rapidez::config('general/region/display_all', '0') }} || '{{ Rapidez::config('general/region/state_required') }}'.split(',').includes({{ $country }})"
+                v-bind:required="'{{ Rapidez::config('general/region/state_required') }}'.split(',').includes({{ $country }})"
+            >
+                <option v-for="region in data.country.available_regions" :value="region.id">
```
Roene-JustBetter, line 16: In the country select we sort alphabetically for a good UX. Should we do the same inside this select? @royduin @indykoning @Jade-GG

resources/views/components/input/select/country.blade.php
```diff
@@ -1,7 +1,7 @@
 <graphql query="{ countries { two_letter_abbreviation full_name_locale } }" cache="countries">
     <div v-if="data" slot-scope="{ data }">
         <x-rapidez::input.select {{ $attributes }}>
-            <option v-for="country in Object.values(data.countries).sort((a, b) => a.full_name_locale.localeCompare(b.full_name_locale))" :value="country.two_letter_abbreviation.toUpperCase()">
+            <option v-for="country in Object.values(data.countries).filter(country => country.full_name_locale).sort((a, b) => a.full_name_locale.localeCompare(b.full_name_locale))" :value="country.two_letter_abbreviation.toUpperCase()">
```
royduin, line 4: We're using `toSorted()` at the region select; `Object.values().sort()` here, cleaner to do the same everywhere?

yarn.lock
royduin, line 1: Why do we need the frontend dependencies to be updated?

Jade-GG: It's necessary for the `has` selector: https://tailwindcss.com/blog/tailwindcss-v3-4#new-has-variant

resources/views/components/input/select/country.blade.php
Jade-GG: Should be `toSorted()` everywhere, `sort()` changes the array in place and can cause weird reactivity issues. Not here, of course, because it's behind an `Object.values()`, but it's good to keep it consistent this way.

Roene-JustBetter (approved):

resources/views/components/input/select/country.blade.php
royduin: But I found multiple polyfills for `toSorted` in projects, sure this is the smartest option to use? We don't have that in the core?

Jade-GG: We could add a polyfill. However, note that if we're migrating to Tailwind v4, any browsers that don't support `toSorted` also don't support Tailwind v4 styling.

indykoning (approved):

PR #740, opened 2025-02-20
[3.x] Make sure you can't do a mutation twice simultaneously

Jade-GG:
If you have a broken mouse that double clicks (or you just click twice very fast for some reason) you can bypass the standard disabled state and mutate twice. Magento generally does not like this---in most cases it will just mean one of the queries is eaten, but sometimes this can break things (e.g. create 2 exact same orders from 1 cart).

### DIFF
--- a/resources/js/components/GraphqlMutation.vue
+++ b/resources/js/components/GraphqlMutation.vue
@@ -106,6 +106,10 @@ export default {
 
     methods: {
         async mutate() {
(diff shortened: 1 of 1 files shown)


royduin (approved):
Looking good to me, let's wait for @indykoning his review.

indykoning (approved):
Looks good!

PR #741, opened 2025-02-21
Feature/tailwind v4 update

Roene-JustBetter:
WIP - Tailwind update to version 4.

### DIFF
--- a/resources/css/app.css
+++ b/resources/css/app.css
@@ -13,10 +13,10 @@
 
 @theme {
     /* Our variables used by different utilities */
-    --primary: #2FBC85;
+    --primary: #2fbc85;
     --primary-text: var(--color-white);
 
-    --secondary: #202F60;
+    --secondary: #202f60;
     --secondary-text: var(--color-white);
 
     --conversion: var(--color-green-500);
@@ -34,7 +34,7 @@
     --background: var(--color-slate-100);
     --background-muted: var(--color-slate-50);
 
-      /* Utility colors linked to our variables */
+    /* Utility colors linked to our variables */
     --color-primary: var(--primary);
     --color-primary-text: var(--primary-text);
 
@@ -114,7 +114,7 @@
     }
 
     button:not(:disabled),
-    [role="button"]:not(:disabled) {
+    [role='button']:not(:disabled) {
         cursor: pointer;
     }
 }


tailwind.config.js
```diff
@@ -1,12 +1,3 @@
...
-import forms from '@tailwindcss/forms'
-import typography from '@tailwindcss/typography'
-import scrollbarHide from 'tailwind-scrollbar-hide'
-
-function color(variable, fallback) {
-    return 'color-mix(in srgb, var(' + variable + ', ' + fallback + ') calc(100% * <alpha-value>), transparent)'
-}
-
 export default {
     content: [
```
royduin, line 2: Did you try this? https://tailwindcss.com/docs/detecting-classes-in-source-files#explicitly-registering-sources, Laravel is also using this: https://github.com/laravel/laravel/blob/12.x/resources/css/app.css

postcss.config.js
royduin, line 1: Did you try this? https://tailwindcss.com/blog/tailwindcss-v4#first-party-vite-plugin, after that all postcss imports within the `package.json` could be removed.

resources/css/app.css
```diff
@@ -1,18 +1,138 @@
...
+
+@theme {
+    --color-primary: color-mix(in srgb, var(--primary, #2fbc85) calc(100% * 1), transparent);
+    --color-primary-text: color-mix(in srgb, var(--primary-text, #fff) calc(100% * 1), transparent);
+    --color-secondary: color-mix(in srgb, var(--secondary, #202f60) calc(100% * 1), transparent);
+    --color-secondary-text: color-mix(in srgb, var(--secondary-text, #fff) calc(100% * 1), transparent);
+    --color-conversion: color-mix(in srgb, var(--conversion, #22c55e) calc(100% * 1), transparent);
+    --color-conversion-text: color-mix(in srgb, var(--conversion-text, #fff) calc(100% * 1), transparent);
+    --color-foreground-emphasis: color-mix(in srgb, var(--foreground-emphasis, #0f172a) calc(100% * 1), transparent);
+    --color-foreground: color-mix(in srgb, var(--foreground, #1e293b) calc(100% * 1), transparent);
```
royduin, line 21: As per: https://tailwindcss.com/docs/colors#referencing-in-css and previously within the `tailwind.config.js` we can reference existing colors;

```
--color-foreground: color-mix(in srgb, var(--foreground, var(--color-slate-800)) calc(100% * 1), transparent);
```

Demo: https://play.tailwindcss.com/QB164GxvSS

postcss.config.js
Roene-JustBetter: Yes fixed it.

tailwind.config.js
Roene-JustBetter: I tried using the `@source` in the app.css inside the `Rapidez/core` But I didn't managed to load styling from other vendor packages like rapidez/menu or rapidez/account.

resources/css/app.css
```diff
@@ -1,18 +1,138 @@
...
+
+@utility text {
+    color: var(--text-color);
+}
+@utility text-emphasis {
+    color: var(--text-color-emphasis);
+}
+@utility text-muted {
+    color: var(--text-color-muted);
+}
```
royduin, line 100: With the scoped colors we don't need this, only `text` doesn't work by default so that one should stay. But... it doesn't support the opacity modifier (eg `text/50`), see: https://github.com/tailwindlabs/tailwindcss/issues/16824

For the time being, the "full version" does support it: `text-default/50`

tailwind.config.js
Roene-JustBetter: I have have checked the docs about using `@source`. 
https://tailwindcss.com/docs/detecting-classes-in-source-files#which-files-are-scanned

So we have to add: https://tailwindcss.com/docs/detecting-classes-in-source-files#explicitly-registering-sources./
So I tried this:
 ```
@source "../../vendor/rapidez/**/*.blade.php";
@source "../../vendor/rapidez/**/*.css";
@source "../../vendor/rapidez/**/*.vue";

@source '../../resources/views/**/*.blade.php';
@source '../../resources/css/**/*.css';
@source '../../resources/js/**/*.vue';

@source '../../../../config/rapidez/frontend.php';
@source '../../vendor/rapidez/core/config/rapidez/frontend.php';
@source '../../../../config/rapidez/menu.php';
@source '../../../../vendor/rapidez/menu/config/rapidez/menu.php';
@source '../../../../../vendor/laravel/framework/src/Illuminate/Pagination/resources/views/tailwind.blade.php';
```

But this work _partly_
This works correct and generated the classes from the core:
```
@source '../../resources/views/**/*.blade.php';
@source '../../resources/css/**/*.css';
@source '../../resources/js/**/*.vue';
```
Scherm­afbeelding 2025-03-06 om 13 20 11

The problem is that styling from vendor packages like account doesn't work:
```
@source "../../vendor/rapidez/**/*.blade.php";
@source "../../vendor/rapidez/**/*.css";
@source "../../vendor/rapidez/**/*.vue";
```
Scherm­afbeelding 2025-03-06 om 13 21 41

The issue is that we are using symlinks and that is not working correctly. I double checked this with @indykoning and we found this: https://github.com/tailwindlabs/tailwindcss/issues/16765

Roene-JustBetter: Looks like the `@source` is still nog working good. I have tried this inside [Rapidez](https://github.com/rapidez/rapidez/blob/master/resources/css/app.css) (so not inside the Core):

This works: 
`@source "../../vendor/rapidez/account/resources/views/account/**/*.blade.php";` 

This doesn't work: 
`@source "../../vendor/rapidez/*/resources/views/account/**/*.blade.php";`. 

The problem is with the `/*/`

indykoning:
We're being blocked by the following 2 issues:
 - Tailwind V4 does not support custom classes in `@apply` https://github.com/tailwindlabs/tailwindcss/discussions/16429
 - Tailwind V4 does not seem to resolve aliasses in `@source` or any other custom directives (like `@reference` which could've fixed issue 1) https://github.com/tailwindlabs/tailwindcss/issues/15159

PR #742, opened 2025-02-21
Fix z-index slideover and place it from the left instead of right

Roene-JustBetter:
Before:
Scherm­afbeelding 2025-02-21 om 15 00 46

After:
Scherm­afbeelding 2025-02-21 om 15 08 51

Removed the position right as the slideover opens default on the left side of the screen. 
Needed to add a `<div>` around for the z-index so it will go over evert item as you can see in the screenshot.

### DIFF
--- a/resources/views/listing/filters.blade.php
+++ b/resources/views/listing/filters.blade.php
@@ -1,22 +1,24 @@
 @php($id = uniqid('filters-'))
-<x-rapidez::slideover.mobile :$id :title="__('Filters')" position="right">
-    <div class="w-full p-2 max-lg:bg-white max-lg:p-5">
-        {{-- On mobile the filters aren't immedately visible so we should defer loading --}}
-        <lazy>
-            @include('rapidez::listing.partials.filter.selected')
-            @include('rapidez::listing.partials.filter.category')
-            <template v-for="filter in filters">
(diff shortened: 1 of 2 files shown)


resources/views/listing/filters.blade.php
```diff
@@ -1,22 +1,24 @@
...
-                @include('rapidez::listing.partials.filter.boolean')
-                @include('rapidez::listing.partials.filter.select')
-            </template>
-        </lazy>
-        <x-rapidez::button.primary for="{{ $id }}" class="w-full text-sm lg:hidden">
-            @lang('Show results')
-        </x-rapidez::button.primary>
-    </div>
-</x-rapidez::slideover.mobile>
+<div class="max-lg:relative max-lg:z-slideover">
```
royduin, line 2: Isn't this a z-index that should go in the slideover instead of wrapping it in it? As every slideover should get this z-index, not only this one with filters?

royduin: Merged and released: https://github.com/rapidez/blade-components/releases/tag/1.6.1, this should require that version

Roene-JustBetter: Added `"rapidez/blade-components": "^1.6.1"` - Checked it and z-index works correct now.

PR #752, opened 2025-03-03
[2.x] Allow for arbitrary options in additionals

Jade-GG:
see #751


indykoning:
Both PRs are pointing at master, i'm assuming this one should go to the 2.x branch

PR #753, opened 2025-03-03
[3.x] Use correct show_tax value

Jade-GG:
For this value the following options are possible:

- `1`: Excluding tax
- `2`: Including tax
- `3`: Including and excluding tax

With the cast to bool this would always return true. Instead we should check for value `2`.


indykoning:
What about value 3? That'd also instruct to show tax right?

Jade-GG:
> What about value 3? That'd also instruct to show tax right?

I'm not sure because this one instructs to show both. It's easier on the frontend to handle this by having `show_tax` be false and then adding the tax on after. However, I do not know how magento handles this option because we don't normally use/support this option.

indykoning:
I do not know how it is currently set up in the frontend so this may be disregarded.

Ignoring current implementation it would make sense to set `show_tax` to true when it's 2 or 3.
Then in the future we could add a `show_excluding_tax` and setting it to true when it's 1 or 3 option

indykoning (approved):

PR #758, opened 2025-03-04
[Instantsearch] Category filter max depth & root path

Jade-GG:
This adds the dynamic max depth on the category filter and the root path on category pages. I've also added the filter heading to make it collapsible.

### DIFF
--- a/resources/js/components/Listing/Listing.vue
+++ b/resources/js/components/Listing/Listing.vue
@@ -88,14 +88,14 @@ export default {
         // TODO: Maybe move this completely to PHP?
         // Any drawbacks? A window.config that
         // becomes to big? Is that an issue?
-        filters: function () {
+        filters() {
             return Object.values(this.attributes)
                 .filter((attribute) => attribute.filter)
                 .map((filter) => ({ ...filter, code: this.filterPrefix(filter) + filter.code, base_code: filter.code }))
                 .sort((a, b) => a.position - b.position)
         },
 
-        facets: function () {
+        facets() {
             return [
                 ...this.filters.map((filter) => ({
                     attribute: filter.code,
@@ -116,11 +116,11 @@ export default {
             return Array.from({ length: config.max_category_level ?? 3 }).map((_, index) => 'category_lvl' + (index + 1))
         },
 
-        sortings: function () {
+        sortings() {
             return Object.values(this.attributes).filter((attribute) => attribute.sorting)
         },
 
-        hitsPerPage: function () {
+        hitsPerPage() {
             return this.$root.config.grid_per_page_values
                 .map(function (pages, index) {
                     return {
@@ -132,7 +132,7 @@ export default {
                 .concat({ label: this.$root.config.translations.all, value: 10000 })
         },
 
-        sortOptions: function () {
+        sortOptions() {
             return [
                 {
                     label: window.config.translations.relevance,
@@ -182,7 +182,7 @@ export default {
     },
 
     watch: {
-        attributes: function (value) {
+        attributes(value) {
             this.loaded = Object.keys(value).length > 0
         },
     },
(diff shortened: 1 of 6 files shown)


resources/js/components/Listing/Listing.vue
```diff
@@ -102,15 +102,20 @@ export default {
...
+                    attribute: attribute,
+                    field: attribute + '.keyword',
+                    type: 'string',
+                })),
             ]
             // TODO: Double check this and how it's used.
             // .concat(this.additionalFilters)
         },
 
+        categoryAttributes() {
```
royduin, line 115: I think we should keep the syntax inline, use this one everywhere; or the syntax used there. Pros/cons?

src/Http/ViewComposers/ConfigComposer.php
```diff
@@ -57,6 +58,7 @@ public function getConfig(): array
             'show_tax'                     => (bool) Rapidez::config('tax/display/type', 1),
             'grid_per_page'                => Rapidez::config('catalog/frontend/grid_per_page', 12),
             'grid_per_page_values'         => explode(',', Rapidez::config('catalog/frontend/grid_per_page_values', '12,24,36')),
+            'max_category_level'           => Category::withoutGlobalScopes()->selectRaw('MAX(level) as maxlevel')->first()->maxlevel,
```
royduin, line 61: Should we cache this? It's just a simple quick query, but Redis is always faster.

Jade-GG: I thought this whole config composer got cached, but it looks like it does not? In that case, yeah it should be cached. But shouldn't we cache the whole thing?

royduin: All the individual `Rapidez::config()`'s are cached bundled with one cache key. This config could be dynamic; for example the checkout GraphQL queries are just useful for checkout pages. So just caching the output with `Cache::rememberForever()` is fine with me.

resources/js/components/Listing/Listing.vue
royduin: Ah sorry, the name is fine. It's the syntax; 
```
categoryAttributes() {
```
vs
```
categoryAttributes: function () {
```
That's currently mixed.

Jade-GG: Ah, well we do use the former in most places that I know of.

Looking at the documentation, the Vue3 documentation has the `categoryAttributes() {` format by default whereas (apparently) the Vue2 documentation has the `categoryAttributes: function() {` format. Personally I prefer the former just because it looks cleaner to me. Aside from that I don't think there are any other reasons to use one over the other.

royduin: Let's use the ES6 shorthand syntax everywhere 🚀 eg: `categoryAttributes() {`

royduin (approved):
Looking good to me, let's wait for @indykoning his approval

src/Http/ViewComposers/ConfigComposer.php
```diff
@@ -57,6 +58,7 @@ public function getConfig(): array
             'show_tax'                     => (bool) Rapidez::config('tax/display/type', 1),
             'grid_per_page'                => Rapidez::config('catalog/frontend/grid_per_page', 12),
             'grid_per_page_values'         => explode(',', Rapidez::config('catalog/frontend/grid_per_page_values', '12,24,36')),
+            'max_category_level'           => Cache::rememberForever('max_category_level', fn () => Category::withoutGlobalScopes()->selectRaw('MAX(level) as maxlevel')->first()->maxlevel),
```
indykoning, line 61: https://laravel.com/docs/master/eloquent#retrieving-aggregates
```suggestion
            'max_category_level'           => Cache::rememberForever('max_category_level', fn () => Category::withoutGlobalScopes()->max('level')),
```

PR #759, opened 2025-03-04
[3.x] Skip login checkout step when logged in (PHP solution)

BobWez98:
I have made 2 versions of this issue. This is a PHP solution when the Customer can be retrieved we skip the login step.

JS version:

### DIFF
--- a/src/Http/Controllers/CheckoutController.php
+++ b/src/Http/Controllers/CheckoutController.php
@@ -15,7 +15,7 @@ class CheckoutController
             $step = $checkoutSteps[0];
         }
 
-        if (auth('magento-customer')->user()) {
+        if ($step === 'login' && auth('magento-customer')->user()) {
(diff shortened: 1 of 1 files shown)


royduin (approved):
This one is looking the simplest / best to me.

Jade-GG (approved):
Agree, this is cleaner to me.

indykoning (approved):

royduin:
Did you also check the onestep checkout option?

BobWez98:
Yes, the `$step` will never be `login` then so this won't have any effect.

royduin:
But it does include the login step: https://github.com/rapidez/core/blob/a554a07fe6ed9eb558cdf243aa7625d291c8e477/resources/views/checkout/pages/onestep.blade.php#L20, not sure what that will show and if it still makes sense.

PR #760, opened 2025-03-04
[3.x] Skip login checkout step when logged in (PHP Solution)

BobWez98:
I have made 2 versions of this issue. This is a JS solution where we check if a user is logged in and redirect them to the next step.

PHP version: #759


resources/views/checkout/steps/login.blade.php
```diff
@@ -1,4 +1,4 @@
-<checkout-login v-slot="checkoutLogin">
+<checkout-login v-slot="checkoutLogin" next-url="{{ route('checkout', ['step' => 'credentials']) }}">
```
indykoning, line 1: This responsibility should fall within the [login page](https://github.com/rapidez/core/blob/master/resources/views/checkout/pages/login.blade.php)

Doing this in the step causes it to happen on onepage checkouts or other combinations (like done in the checkout theme) as well

indykoning (changes requested):

royduin:
Closing in favor of https://github.com/rapidez/core/pull/759

PR #762, opened 2025-03-05
[Instantsearch] More routing params

Jade-GG:
Adds the per page & sorting parameters. Unfortunately I was not yet able to make the sorting work nicely on refresh despite much trying.

What I have now is a terrible fix that should not be in the final version, but it does work flawlessly.


royduin:
Follow-up: https://github.com/rapidez/core/pull/799

PR #764, opened 2025-03-06
[Instantsearch] Index name from models

Jade-GG:
Adds a default index name in the exact same way that Laravel does for model tables.

Also exactly like the table thing, this allows you to override it in two ways:

```php
public string $indexName = 'customname';
```

```php
public function getIndexName()
{
    return 'customname';
}
```

Also now using `searchableAs` to determine the index name in the blade file instead of getting it from the config. We might be able to remove the prefix from the config now? It's not used anywhere right now.

### DIFF
--- a/resources/views/components/listing.blade.php
+++ b/resources/views/components/listing.blade.php
@@ -9,14 +9,6 @@
 <div class="min-h-screen">
     <listing
         {{ $attributes }}
-        :additional-sorting="[{
-            label: window.config.translations.newest,
-            field: 'created_at',
-            order: 'desc',
-            {{-- TODO: Extract this somewhere? --}}
-            value: config.index_prefix + '_{{ config('rapidez.models.product')::indexName() }}_' + config.store + '_created_at_desc',
-            key: '_created_at_desc'
-        }]"
         index-model="{{ config('rapidez.models.product')::indexName() }}"
         v-cloak
     >
--- a/resources/views/components/listing.blade.php
+++ b/resources/views/components/listing.blade.php
@@ -1,3 +1,5 @@
+@props(['rootPath' => null])
+
 @pushOnce('head', 'es_url-preconnect')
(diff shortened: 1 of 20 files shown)


resources/views/components/listing.blade.php
```diff
@@ -9,8 +9,7 @@
 <div class="min-h-screen">
     <listing
         {{ $attributes }}
-        {{-- TODO: Extract this somewhere? --}}
-        :index="config.index_prefix + '_products_' + config.store"
+        index="{{ (new (config('rapidez.models.product')))->searchableAs() }}"
```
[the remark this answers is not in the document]
Jade-GG: This is something I discussed with @royduin and we made a story to look at this later (RAP-1226).

In short, within a Laravel model any function can be called statically, except because of how PHP works it won't actually do that for public functions so you can only actually statically call protected and private functions.  
I had originally made a second protected function so that it could be called statically, but for simplicity's sake we decided to keep it like this for now and look at it later.

resources/views/layouts/partials/header/autocomplete.blade.php
```diff
@@ -17,7 +17,7 @@
         <ais-instant-search
             class="contents"
             :search-client="searchClient"
-            :index-name="config.index_prefix + '_products_' + config.store"
+            index-name="{{ (new (config('rapidez.models.product')))->searchableAs() }}"
```
royduin, line 20: This is now repeated multiple times; can't we just push this in the config and use that? `config.index.products`, `config.index.categories`, etc. as mentioned in the ConfigComposer? And please double check that file, search for `index`

royduin (commented):
Could you create a follow-up on this @Jade-GG?

resources/views/layouts/partials/header/autocomplete.blade.php
Jade-GG: Fixed in #769

PR #765, opened 2025-03-06
Remove need to preload lazy loaded components

indykoning:
Unfortunately it seems impossible to register components globally after the Vue instance is created:
https://v2.vuejs.org/v2/guide/components-registration#Global-Registration:~:text=global%20registration%20must%20take%20place%20before%20the%20root%20Vue%20instance%20is%20created

While local registration is possible and works with async imported components.
https://v2.vuejs.org/v2/guide/components-registration#Local-Registration-in-a-Module-System
Templates within slot scopes do not have access to these local components, thus they do not work within blade either.

The only way i've found to get them to work is to import them lazily before vue is initialized.
I've added the imports to the listing and autocomplete so Vite can still help load them before Vue knows they're required on the page


Jade-GG:
With regards to scopes, do we still have access to variables of a parent component when inside of this `inline-template`? In other words, would this still work?

```blade
<my-component v-slot="{ myVariable }">
    [...]
    <listing inline-template>
        <div>
            @{{ myVariable }}
        </div>
    </listing>
</my-component>
```

indykoning:
> With regards to scopes, do we still have access to variables of a parent component when inside of this `inline-template`? In other words, would this still work?
> 
> ```
> <my-component v-slot="{ myVariable }">
>     [...]
>     <listing inline-template>
>         <div>
>             @{{ myVariable }}
>         </div>
>     </listing>
> </my-component>
> ```

That's a good one, it indeed does not work because the template is scoped to the `inline-template` component any slot scopes defined inside are not affected.
Slots defined outside of the component do not work though. You would need to do:

```vue
<my-component>
    [...]
    <listing inline-template>
        <div>
            @{{ $parent.myVariable }}
        </div>
    </listing>
</my-component>
```
```

Jade-GG (approved):

royduin:
> Unfortunately it seems impossible to register components globally after the Vue instance is created

So why don't we just register them globally before the Vue instance creation? This current solution feels a bit hacky.

Could you create a follow-up on this @indykoning?

indykoning:
Registering them globally will prevent Vite from bundling them properly. It will function, but we will be left with a file per component.

Good to note, this is exactly what we do: https://v3-migration.vuejs.org/breaking-changes/inline-template-attribute#:~:text=all%20templates%20are%20written%20directly%20inside%20the%20HTML%20page We need to keep this in mind when doing the vue 3 migration

I've created internal reference: RAP-1237

PR #766, opened 2025-03-06
[Instantsearch] Add possibilities for simplifying basefilters

Jade-GG:
Made a simple change that allows you do add simple filter stuff without having to write the whole query object. This only includes two:

- `filterQueryString` which takes a query string like `visibility: (3 OR 4)`
- `filterScoreScript` which lets you modify the scoring by using an ES script.

I've left it quite simple so that we can easily add more options in case we would like them in the future.

Side note: The baseFilters prop is a function on purpose, as we will want to be able to have dynamic things in there in the future. For example, it will be extremely useful in some project to change the filter based on the typed search query.

### DIFF
--- a/resources/js/components/Listing/Listing.vue
+++ b/resources/js/components/Listing/Listing.vue
@@ -60,6 +60,13 @@ export default {
         },
         baseFilters: {
             type: Function,
+            default: () => [],
+        },
+        filterQueryString: {
+            type: String,
+        },
+        filterScoreScript: {
+            type: String,
         },
     },
 
@@ -192,7 +199,7 @@ export default {
         // directly due the JS size
         initSearchClient() {
             return Client(this.searchkit, {
-                getBaseFilters: this.baseFilters,
+                getBaseFilters: this.getBaseFilters,
                 getQuery: this.query,
             })
         },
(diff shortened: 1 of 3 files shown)


resources/js/components/Listing/Listing.vue
```diff
@@ -60,6 +60,13 @@ export default {
         },
         baseFilters: {
             type: Function,
+            default: () => [],
+        },
+        filterQueryString: {
+            type: String,
+        },
+        filterScoreScript: {
+            type: String,
```
indykoning, line 69: Don't forget to add these to the list of props we need to document 😉

indykoning (approved):

royduin:
Why is this already merged? 😇 

I don't really like the `filterScoreScript` as it's likely only used for this one case. I think it's better to have something like:

```blade
<x-rapidez::listing
    :filter-query-string="'visibility:(2 OR 4) AND category_ids:'+config.category.entity_id"
    :filter-query-dsl="categoryPositions(config.category.entity_id)"
>
```

Where we're able to change the whole DSL instead of just the score piece with a callback. 

Or maybe we could add a preset for this with just:

```blade
<x-rapidez::listing :category="config.category.entity_id">
```

Which will add all these defaults; category, visibility and positions. Optionally you can change the query with those props.

Could you create a follow-up on this @Jade-GG?

Jade-GG:
> Could you create a follow-up on this @Jade-GG?

See RAP-1236

PR #767, opened 2025-03-06
[3.x] Add custom attributes to user store

Jade-GG:
#685 but then inside of the useUser store.

### DIFF
--- a/resources/js/stores/useUser.js
+++ b/resources/js/stores/useUser.js
@@ -163,6 +163,19 @@ export const user = computed({
         }
 
         userStorage.value.is_logged_in = Boolean(userStorage.value?.email)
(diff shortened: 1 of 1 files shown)


indykoning (dismissed):

royduin:
Do we need some info on this anywhere? Where is it for and how to use it. Maybe just a comment above it or maybe something in the docs?

Jade-GG:
> Do we need some info on this anywhere? Where is it for and how to use it. Maybe just a comment above it or maybe something in the docs?

I don't really see a nice place to put this in the docs, but I think a good comment here would suffice?

royduin:
Looking good! Thanks 🚀

PR #768, opened 2025-03-07
Move duplicate instantsearch code to mixin, improve autocomplete loading

indykoning:
Due to the guaranteed merge conflicts with https://github.com/rapidez/core/pull/765 i've based this one on it.

I've centralized the base instantsearch code to the InstantSearchMixin
Any places where the Autocomplete or Listing extend or change the configuration i've made it possible to do so by overriding the function.

This also results in improved bundling:
- The Autocomplete bundle contains the ais-index component as it's the only file using it.
- The Listing bundle contains the components for filtering, sorting and stats. It also contains the instantsearch code for pagination, sorting, stats and current refinements.
- The InstantSearchMixin contains the rest of the instantsearch library + ais-configure, ais-highlight, ais-search-box, ais-hits, ais-instantsearch. Since both the Autocomplete and Listing use them

You might also notice i do not include the `ais-search-box` in the `InstantSearchMixin` and yet it's still in the bundle.
This is because they both use the `InstantSearchMixin` and it's not used anywhere else, and they both use the `ais-search-box`. So it's more efficient to bundle it in the shared file.

I've not included it in the `InstantSearchMixin` myself because it's not necessary to get InstantSearch working and other implementations may not need to use the searchbox.

I've also stumbled upon https://v2.vuejs.org/v2/guide/components-dynamic-async.html#Handling-Loading-State which allows for the autocomplete to switch seamlessly between the facade and full component!

### DIFF
--- a/resources/js/components/Search/Autocomplete.vue
+++ b/resources/js/components/Search/Autocomplete.vue
@@ -1,21 +1,15 @@
 <script>
-import Client from '@searchkit/instantsearch-client'
-import Searchkit from 'searchkit'
-
-import AisInstantSearch from 'vue-instantsearch/vue2/es/src/components/InstantSearch'
 import AisSearchBox from 'vue-instantsearch/vue2/es/src/components/SearchBox.vue.js'
-import AisHits from 'vue-instantsearch/vue2/es/src/components/Hits.js'
 import AisIndex from 'vue-instantsearch/vue2/es/src/components/Index.js'
-import AisConfigure from 'vue-instantsearch/vue2/es/src/components/Configure.js'
 import AisHighlight from 'vue-instantsearch/vue2/es/src/components/Highlight.vue.js'
 
+import InstantSearchMixin from './InstantSearchMixin.vue'
+
 export default {
+    mixins: [InstantSearchMixin],
     components: {
-        'ais-instant-search': AisInstantSearch,
         'ais-search-box': AisSearchBox,
-        'ais-hits': AisHits,
         'ais-index': AisIndex,
-        'ais-configure': AisConfigure,
         'ais-highlight': AisHighlight,
     },
     data: () => ({
@@ -33,19 +27,9 @@ export default {
         })
     },
 
-    computed: {
-        // TODO: Not sure if this is the right place,
-        // the autocomplete also needs this but
-        // we don't want to load everything
-        // directly due the JS size
-        searchClient: function () {
-            let client = Client(this.searchkit, {
-                hooks: {
-                    beforeSearch: async (searchRequests) => {
-                        return searchRequests
-                    },
-                },
-            })
+    methods: {
+        async initSearchClient () {
+            const client = await InstantSearchMixin.methods.initSearchClient.bind(this).call();
 
             // Ensure no query is done if the search field is empty
             const oldSearch = client.search
@@ -70,32 +54,6 @@ export default {
             }
 
             return client
-        },
-
-        // TODO: Maybe extract this so we have this once?
-        searchkit: function () {
-            let url = new URL(config.es_url)
-
-            let searchkit = new Searchkit({
-                connection: {
-                    host: url.origin,
-                    auth: {
-                        username: url.username,
-                        password: url.password,
-                    },
-                },
-
-                // TODO: Should we split this as it could
-                // diff from the settings on listings.
-                search_settings: {
-                    highlight_attributes: config.searchkit.highlight_attributes,
-                    search_attributes: config.searchkit.search_attributes,
-                    result_attributes: config.searchkit.result_attributes,
-                    filter_attributes: config.searchkit.filter_attributes,
-                },
-            })
-
-            return searchkit
         },
     },
 }
--- a/resources/js/components/Search/Autocomplete.vue
+++ b/resources/js/components/Search/Autocomplete.vue
@@ -28,8 +28,8 @@ export default {
     },
 
     methods: {
-        async initSearchClient () {
-            const client = await InstantSearchMixin.methods.initSearchClient.bind(this).call();
+        async initSearchClient() {
+            const client = await InstantSearchMixin.methods.initSearchClient.bind(this).call()
 
             // Ensure no query is done if the search field is empty
             const oldSearch = client.search
--- a/resources/js/components/Search/InstantSearchMixin.vue
+++ b/resources/js/components/Search/InstantSearchMixin.vue
@@ -37,12 +37,12 @@ export default {
                     },
                 },
 
-                search_settings: await this.getSearchSettings()
+                search_settings: await this.getSearchSettings(),
             })
         },
 
         async getInstantSearchClientConfig() {
-            return {};
+            return {}
         },
 
         async getSearchSettings() {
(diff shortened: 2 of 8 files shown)


resources/js/components/Search/Autocomplete.vue
```diff
@@ -28,24 +22,14 @@ export default {
 
     mounted() {
         this.$nextTick(() => {
-            this.$emit('mounted')
+            window.setTimeout(() => this.$emit('mounted'))
```
Jade-GG, line 25: What's this for? Could this also be another `$nextTick`? What if you move the `this.loaded` to outside of the parent `$nextTick`?

resources/views/layouts/partials/header/autocomplete.blade.php
```diff
@@ -35,6 +25,11 @@ class="contents"
                                     spellcheck="false"
                                     name="q"
                                     v-bind:value="currentRefinement"
+                                    v-on:focus="() => {
+                                        refine($root.autocompleteFacadeQuery || currentRefinement);
+                                        $root.autocompleteFacadeQuery = null;
+                                    }
+                                    "
```
Jade-GG, line 32: Formatting, should the `"` be on the same line as the `}`?

resources/js/components/Search/Autocomplete.vue
indykoning: Unfortunately the 2nd nextTick is still too early, i do think i can use requestAnimationFrame though.

Out of curiosity i set some performance marks
image
You can see the small slivers at the top.

The first is the mounted method itself.
The second and third are $nextTicks nested in each other.
the fourth is the requestAnimationFrame, this is the least we need. We require the style recalculation to be done.
the fifth is the setTimeout, coming less than a milisecond later. But being in it's own task.

Jade-GG: Definitely prefer the requestAnimationFrame then, cool stuff 👍

Jade-GG (approved):

resources/views/layouts/partials/header/autocomplete.blade.php
```diff
@@ -50,5 +44,19 @@ class="contents"
...
             </div>
         </ais-instant-search>
+        <div v-else class="relative w-full">
+            {{-- TODO: Do we still need this double input? --}}
+            <form name="autocomplete-form" id="autocomplete-form" method="get" action="{{ route('search') }}" class="flex flex-row relative">
+                <x-rapidez::input
+                    type="search"
+                    name="q"
+                    :placeholder="__('What are you looking for?')"
+                    v-model="$root.autocompleteFacadeQuery"
```
royduin, line 54: Can you explain this one?

resources/views/layouts/partials/header/autocomplete.blade.php
```diff
@@ -1,22 +1,11 @@
...
 
-<autocomplete v-else inline-template>
+<autocomplete inline-template v-on:mounted="() => window.document.getElementById('autocomplete-input').focus()">
     <div class="relative w-full">
         <ais-instant-search
+            v-if="searchClient"
+            v-cloak
             class="contents"
-            :search-client="searchClient"
+            :search-client="searchClient "
```
royduin, line 8: I don't think we need the extra space here

resources/views/components/listing.blade.php
```diff
@@ -26,7 +26,7 @@
     >
         <div>
             <ais-instant-search
-                v-if="loaded"
+                v-if="loaded && searchClient"
```
royduin, line 29: Why isn't `loaded` enough? Multiple options will be forgotten easily.

resources/js/vue-components.js
```diff
@@ -34,7 +34,24 @@ Vue.component('images', images)
...
 Vue.component('quantity-select', quantitySelect)
 
-Vue.component('autocomplete', () => import('./components/Search/Autocomplete.vue'))
+Vue.component('autocomplete', () => ({
+    // https://v2.vuejs.org/v2/guide/components-dynamic-async.html#Async-Components
+    component: new Promise(function (resolve, reject) {
+        document.addEventListener('loadAutoComplete', () => import('./components/Search/Autocomplete.vue').then(resolve))
+    }),
+    // https://v2.vuejs.org/v2/guide/components-dynamic-async.html#Handling-Loading-State
+    loading: {
```
royduin, line 43: Nice option! But how do we benefit from this? Now we've 3 states?

1. Vue not loaded yet
2. Vue component loading
3. Component loaded

Does it load faster?

resources/js/components/Search/InstantSearchMixin.vue
```diff
@@ -0,0 +1,61 @@
...
+
+                search_settings: await this.getSearchSettings(),
+            })
+        },
+
+        async getInstantSearchClientConfig() {
+            return {}
+        },
+
+        async getSearchSettings() {
```
royduin, line 48: Do we want these settings globally for the listing and the autocomplete? I can imagine they could have different configs; for example the autocomplete doesn't need the filter attributes.

royduin (commented):
Could you create a follow-up on this @indykoning?

resources/js/vue-components.js
indykoning: We benefit in a much more stable experience on lower-end devices.
State 1 & 2 are the same.
Previously if you set the network speed to Slow 4G clicking the autocomplete made it disappear, until the real autocomplete is loaded.

What happens now is:
you start typing in the input, the autocomplete starts downloading in the background.
It maintains your current input in the autocompleteFacadeQuery (https://github.com/rapidez/core/pull/768#discussion_r2000730342)
If you start typing before Vue has even booted, it will simply catch up due to that v-model.

Once it has downloaded and prepared the real autocomplete it will replace this loading component, refine using autocompleteFacadeQuery thus setting the new input to the current search query, and focus on the input so the user can continue typing without interruption.

resources/views/components/listing.blade.php
indykoning: loaded can be set to true before searchClient is available, and it is absolutely required to start mounting `ais-instant-search`

resources/views/layouts/partials/header/autocomplete.blade.php
indykoning: https://github.com/rapidez/core/pull/768#discussion_r2001072653
> It maintains your current input in the autocompleteFacadeQuery 
If you start typing before Vue has even booted, it will simply catch up due to that v-model.
>
>Once it has downloaded and prepared the real autocomplete it will replace this loading component, refine using autocompleteFacadeQuery thus setting the new input to the current search query

resources/views/components/listing.blade.php
royduin: Can we couple those so only one condition is needed within a template?

resources/js/vue-components.js
royduin: Nice! 🚀

resources/js/package.js
```diff
@@ -100,7 +100,7 @@ function init() {
                 config: window.config,
                 loadingCount: 0,
                 loading: false,
-                loadAutocomplete: false,
+                autocompleteFacadeQuery: '',
```
royduin, line 103: Do we really need this prop to be here on the root Vue instance? Can't we have this on the (loading) autocomplete component?

PR #769, opened 2025-03-10
[Instantsearch] Config refactor

Jade-GG:
This PR refactors how the config files are used:

- Store-specific values now go under `/config/rapidez/stores/[store_code]/[config_file]`.
    - This means that the `themes` and `checkout_steps` variables no longer have their own store-specific 
- Move `window.config` data to a separate web route
- Move `useAttributes` data directly into the `window.config` data
- All searchkit configuration now gets mapped and merged directly in php so that we don't have to do any mapping on the frontend
    - Side effect of this is that sorting label translations are now managed in your json translations
- `indexer.php` config has been removed:
    - Adding custom models should be done by adding them to `models.php`
    - Changing the indexed visibility can be done by extending `product.php` and overwriting the `shouldBeSearchable()` function.

### DIFF
--- a/resources/views/layouts/app.blade.php
+++ b/resources/views/layouts/app.blade.php
@@ -21,7 +21,7 @@
     <link rel="canonical" href="@yield('canonical', url()->current())" />
     <link href="{{ route('config') }}" rel="preload" as="script">
 
-    <script defer src="{{ route('config') }}?v={{ Cache::get('cachekey') }}"></script>
+    <script defer src="{{ route('config') }}?v={{ Cache::rememberForever('cachekey', fn () => md5(Str::random())) }}"></script>
     @vite(['resources/css/app.css', 'resources/js/app.js'])
     @stack('head')
     @config('design/head/includes')


src/Models/Product.php
```diff
@@ -269,7 +269,11 @@ protected function makeAllSearchableUsing(Builder $query)
 
     public function shouldBeSearchable(): bool
     {
-        if (! in_array($this->visibility, config('rapidez.indexer.visibility'))) {
+        // VISIBILITY_NOT_VISIBLE    = 1
+        // VISIBILITY_IN_CATALOG     = 2
+        // VISIBILITY_IN_SEARCH      = 3
+        // VISIBILITY_BOTH           = 4
```
indykoning, line 275: It may be useful to turn these into constants to improve readability

src/Rapidez.php
```diff
@@ -165,6 +172,16 @@ public function setStore(Store|array|callable|int|string $store): void
...
+        $path = config('rapidez.frontend.theme', false);
+        if ($path) {
+            config([
+                'view.paths' => [
+                    $path,
+                    ...config('view.paths'),
+                ],
+            ]);
+        }
+
```
indykoning, line 184: We'll be too late if we set the paths here
https://github.com/laravel/framework/blob/e6753fc3836bb5fdddd1ee95835d8065bd7c5b16/src/Illuminate/View/ViewServiceProvider.php#L85
https://github.com/laravel/framework/blob/e6753fc3836bb5fdddd1ee95835d8065bd7c5b16/src/Illuminate/View/FileViewFinder.php#L53

it's set and kept after the register methods. Any changes on this config after that have no effect

src/Rapidez.php
```diff
@@ -130,6 +130,13 @@ public function setStore(Store|array|callable|int|string $store): void
         config()->set('rapidez.root_category_id', $store['root_category_id']);
         config()->set('frontend.base_url', url('/'));
 
+        // Set all store-specific values
+        foreach (array_keys(config('rapidez')) as $config) {
+            foreach (config('rapidez.stores.' . $store['code'] . '.' . $config, []) as $key => $value) {
+                config()->set('rapidez.' . $config . '.' . $key, $value);
+            }
+        }
+
```
indykoning, line 139: When changing to multiple stores (like we do during indexing) wouldn't we start setting config from store 2 on store 1, instead of setting the config from store 2 on the default config.

If that's the case we might not actually get the default config if we haven't overridden it in store 2, instead getting the config that store 1 overrode.

indykoning (changes requested):

src/Rapidez.php
Jade-GG: Good points---I've now added a thing that sets "default" values when they get changed, and then uses that to reset to when switching store again.

src/RapidezServiceProvider.php
```diff
@@ -285,6 +262,19 @@ protected function registerConfigs(): self
             $this->mergeConfigFrom(__DIR__ . '/../config/rapidez/' . $configFile . '.php', 'rapidez.' . $configFile);
         }
 
+        // Find all store-specific config files and merge them in
+        // TODO: Add how to use this in the Rapidez documentation
+        foreach (glob(__DIR__ . '../config/rapidez/stores/*/*.php') as $configFile) {
```
royduin, line 267: Isn't this a heavy task? Does this work with the config caching?

royduin (changes requested):
What are the pros and cons of this approach?

resources/views/layouts/config.blade.php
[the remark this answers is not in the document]
Jade-GG: Maybe, I wanted to keep things flexible for now so that you can always add anything else you might need here.

resources/js/components/Listing/Listing.vue
```diff
@@ -68,52 +66,14 @@ export default {
     },
 
     mounted() {
-        this.loaded = Object.keys(this.attributes).length > 0
+        this.loaded = true
```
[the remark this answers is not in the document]
Jade-GG: this is *probably* unnecessary now as we also check for the searchClient (`v-if="loaded && searchClient"`). I've left it here for now just in case it broke something by removing it.

lang/en.json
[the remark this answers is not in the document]
Jade-GG: The translations come directly from the configcomposer now, so this was my approach to keeping the translations. However, I did not realize we can still use the php translations (because I'm clearly stupid). I'll move them back.

src/Models/Product.php
```diff
@@ -269,7 +274,11 @@ protected function makeAllSearchableUsing(Builder $query)
 
     public function shouldBeSearchable(): bool
     {
-        if (! in_array($this->visibility, config('rapidez.indexer.visibility'))) {
+        if (! in_array($this->visibility, [
+            static::VISIBILITY_IN_CATALOG,
+            static::VISIBILITY_IN_SEARCH,
+            static::VISIBILITY_BOTH,
+        ])) {
```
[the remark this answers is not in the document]
Jade-GG: My take was that if you want to configure this, you can overwrite the function in your own product model. However with the searchable refactor that Indy did I'm not sure if this would still work. 🤔

src/Rapidez.php
```diff
@@ -130,6 +130,19 @@ public function setStore(Store|array|callable|int|string $store): void
...
+            foreach (config('rapidez.defaults.' . $config, []) as $key => $value) {
+                config()->set('rapidez.' . $config . '.' . $key, $value);
+            }
+
+            // Set all store-specific values and define the relevant defaults
+            foreach (config('rapidez.stores.' . $store['code'] . '.' . $config, []) as $key => $value) {
+                config()->set('rapidez.defaults.' . $config . '.' . $key, config('rapidez.' . $config . '.' . $key));
+                config()->set('rapidez.' . $config . '.' . $key, $value);
+            }
+        }
```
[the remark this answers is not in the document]
Jade-GG: I've added a comment in the code that is hopefully clear enough to explain it.

routes/web.php
```diff
@@ -20,4 +21,11 @@
     Route::get('checkout/{step?}', config('rapidez.routing.controllers.checkout'))->middleware('auth:magento-cart')->name('checkout');
     Route::get('search', config('rapidez.routing.controllers.search'))->name('search');
     Route::fallback(config('rapidez.routing.controllers.fallback'));
+
+    Route::middleware('cache.headers:public;max_age=3600;s_maxage=3600;stale_while_revalidate=3600;etag')->group(function () {
```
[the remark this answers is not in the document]
Jade-GG: @indykoning what do you think?

src/RapidezServiceProvider.php
Jade-GG: Hmm, we could cache the list of config files we get out of the glob and the few lines after? Keeps the code relatively simple and saves on the performance cost of that part.

royduin: - Doesn't Laravel provide something for this?
- Maybe simply override the config part from Laravel and add a store check there?
- Or reverse it; currently we're checking all files but we do know the stores from the database, maybe use that? Or keep a "custom store config list" somewhere?

resources/js/stores/useSwatches.js
```diff
@@ -1,4 +1,4 @@
-// TODO: In this file there is a lot of duplication compared to useAttributes. Can we improve that?
+// TODO: Should we move the swatches to the `config` data?
```
royduin, line 1: Pros/cons? If we're moving the attributes, let's also move the swatches.

resources/views/layouts/app.blade.php
```diff
@@ -19,7 +19,9 @@
     <meta name="description" content="@yield('description', Rapidez::config('design/head/default_description', ''))"/>
     <meta name="robots" content="@yield('robots', Rapidez::config('design/search_engine_robots/default_robots', 'INDEX,FOLLOW'))"/>
     <link rel="canonical" href="@yield('canonical', url()->current())" />
+    <link href="{{ route('config') }}" rel="preload" as="script">
 
+    <script defer src="{{ route('config') }}"></script>
```
royduin, line 24: `defer` is "soon enough"? We won't end up with situations where `window.config` isn't available yet?

routes/web.php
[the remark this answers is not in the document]
royduin: This approach vs caching in sessionstorage?

src/RapidezServiceProvider.php
```diff
@@ -154,33 +153,11 @@ protected function bootRoutes(): self
...
-
-        return $this;
-    }
-
     protected function bootViews(): self
     {
         $this->loadViewsFrom(__DIR__ . '/../resources/views', 'rapidez');
 
-        View::composer('rapidez::layouts.app', ConfigComposer::class);
+        View::composer('rapidez::layouts.config', ConfigComposer::class);
```
royduin, line 160: Does it make sense to still use a composer for this? Maybe a controller is a better idea?

resources/views/layouts/app.blade.php
indykoning: If we have it defined before all other scripts, and are not using it directly in the html we should be fine.

Previously we had it at the bottom of the page without defer, which should have the same effect as having it on the top of the page with defer.

routes/web.php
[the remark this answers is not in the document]
indykoning: I do know local/sessionstorage does have a size limit of ~5mb by default.
I don't know if we would ever reach that or if such a restriction exists for browser caching

resources/views/layouts/app.blade.php
Jade-GG: I tried my hardest to make this cause any errors and was unable to.

routes/web.php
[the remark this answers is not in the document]
royduin: Let's add a cache key as query param.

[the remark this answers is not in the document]
Jade-GG: I've added the cache key as a query param in the `app.blade.php` file for this

royduin:
The last 2 items will be fixed in another PR.

PR #770, opened 2025-03-10
[Instantsearch] Make basic highlight component that doesn't break when the attribute is not in the config

Jade-GG:
Adds the `ais-highlight` component to the mixin, allowing you to use it anywhere.

Also made a small component that makes sure you don't try to highlight something that's not in the `highlight_attributes` configuration. Instantsearch will, by default, just not show the value at all if it's not there.

Can be used like this:

```blade
<x-rapidez::highlight attribute="name"/>
```

If the JS variable is not named `item` you can specify it:

```blade
<x-rapidez::highlight attribute="description" item="product"/>
```

And you can add styling:

```blade
<x-rapidez::highlight attribute="name" class="text-red-500"/>
```

### DIFF
--- a/resources/css/app.css
+++ b/resources/css/app.css
@@ -29,3 +29,7 @@ input[type='number']::-webkit-inner-spin-button,
 input[type='number']::-webkit-outer-spin-button {
     @apply appearance-none m-0;
 }
+
+mark {
+    @apply bg-transparent font-bold
+}
--- a/resources/css/app.css
+++ b/resources/css/app.css
@@ -31,5 +31,5 @@ input[type='number']::-webkit-outer-spin-button {
 }
 
 mark {
-    @apply bg-transparent font-bold
+    @apply bg-transparent font-bold;
 }
--- a/resources/views/components/highlight.blade.php
+++ b/resources/views/components/highlight.blade.php
@@ -1,0 +1,12 @@
+@props(['attribute', 'item' => 'item'])
+
+@if (in_array($attribute, config('rapidez.searchkit.highlight_attributes')))
+    <ais-highlight
+        {{ $attributes }}
+        attribute="{{ $attribute }}"
+        :hit="{{ $item }}"
+        highlighted-tag-name="mark"
+    />
+@else
+    <span {{ $attributes }} v-text="{{ $item }}.{{ $attribute }}"></span>
+@endif
--- a/resources/views/components/highlight.blade.php
+++ b/resources/views/components/highlight.blade.php
@@ -1,12 +1,11 @@
-@props(['attribute', 'item' => 'item'])
+@props(['attribute', 'item' => 'item', 'highlightTag' => 'mark'])
 
 @if (in_array($attribute, config('rapidez.searchkit.highlight_attributes')))
-    <ais-highlight
-        {{ $attributes }}
-        attribute="{{ $attribute }}"
-        :hit="{{ $item }}"
-        highlighted-tag-name="mark"
-    />
+    <ais-highlight {{ $attributes->merge([
+        'v-bind:hit' => $item,
+        'attribute' => $attribute,
+        'highlighted-tag-name' => $highlightTag,
+    ]) }}/>
 @else
     <span {{ $attributes }} v-text="{{ $item }}.{{ $attribute }}"></span>
 @endif
--- a/resources/views/listing/partials/item.blade.php
+++ b/resources/views/listing/partials/item.blade.php
@@ -16,9 +16,7 @@
                 />
                 <x-rapidez::no-image v-else class="mb-3 h-48 rounded-t" />
                 <div class="px-2">
-                    <div class="text-base font-medium">
-                        <x-rapidez::highlight attribute="name"/>
-                    </div>
+                    <x-rapidez::highlight attribute="name" class="text-base font-medium"/>
                     @if (!Rapidez::config('catalog/frontend/show_swatches_in_product_list', 1))
                         <div class="flex items-center space-x-2">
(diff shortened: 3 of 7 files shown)


resources/views/components/highlight.blade.php
```diff
@@ -0,0 +1,12 @@
...
+@if (in_array($attribute, config('rapidez.searchkit.highlight_attributes')))
+    <ais-highlight
+        {{ $attributes }}
+        attribute="{{ $attribute }}"
+        :hit="{{ $item }}"
+        highlighted-tag-name="mark"
+    />
+@else
+    <span {{ $attributes }} v-text="{{ $item }}.{{ $attribute }}"></span>
+@endif
```
royduin, line 12: - Currently it's not possible to change the tag
- The `highlighted-tag-name` default is `mark`, no need to specify it
- The `attribute` is an actual attribute, maybe a bit weird to have it as prop? 
- `item` does have a default, so I would go for everything (+ `attribute` with `name`) or nothing

Maybe something like this is better:

```suggestion
@props(['item' => 'item'])

@if (in_array($attribute, config('rapidez.searchkit.highlight_attributes')))
    <ais-highlight {{ $attributes->merge([
        'attribute' => 'name',
        ':hit' => $item,
    ]) }}/>
@else
    <span {{ $attributes }} v-text="{{ $item }}.{{ $attributes->get('attribute', 'name') }}"></span>
@endif
```

resources/css/app.css
```diff
@@ -29,3 +29,7 @@ input[type='number']::-webkit-inner-spin-button,
 input[type='number']::-webkit-outer-spin-button {
     @apply appearance-none m-0;
 }
+
+mark {
+    @apply bg-transparent font-bold;
+}
```
royduin, line 35: Tailwind recommends against this, this vs having these classes within the Blade component?

resources/views/listing/partials/item.blade.php
```diff
@@ -16,7 +16,9 @@ class="mb-3 h-48 w-full rounded-t object-contain" :alt="item.name" :loading="con
                 />
                 <x-rapidez::no-image v-else class="mb-3 h-48 rounded-t" />
                 <div class="px-2">
-                    <div class="text-base font-medium">@{{ item.name }}</div>
+                    <div class="text-base font-medium">
+                        <x-rapidez::highlight attribute="name"/>
+                    </div>
```
royduin, line 21: I think we don't need the wrapper? And if we're going with the component suggestion; we don't need the atttribute as `name` is the default.

```suggestion
                    <x-rapidez::highlight class="text-base font-medium"/>
```

resources/css/app.css
Jade-GG: If we're putting it within the blade component you get `[&>mark]:bg-transparent [&>mark]:font-bold` which, at least to me, seems like a less preferable option than defining it right here 🤔

resources/views/components/highlight.blade.php
Jade-GG: What's the choice here between using `$attributes->get('attribute', 'name')` instead of just setting it as a prop?

We can do that option a bit cleaner with something like this (we can do without that bit of php at the start, but then you're using `$attributes->get('attribute', 'name')` 3 times if you want it to actually work):

```blade
@props(['item' => 'item'])

@php($attributes['attribute'] ??= 'name')
@if (in_array($attributes['attribute'], config('rapidez.searchkit.highlight_attributes')))
    <ais-highlight {{ $attributes->merge(['v-bind:hit' => $item]) }}/>
@else
    <span {{ $attributes }} v-text="{{ $item }}.{{ $attributes['attribute'] }}"></span>
@endif
```

But at that point I don't see why we shouldn't just keep it as a prop:

```blade
@props(['attribute' => 'name', 'item' => 'item'])

@if (in_array($attribute, config('rapidez.searchkit.highlight_attributes')))
    <ais-highlight {{ $attributes->merge([
        'v-bind:hit' => $item,
        'attribute' => $attribute,
    ]) }}/>
@else
    <span {{ $attributes }} v-text="{{ $item }}.{{ $attribute }}"></span>
@endif
```

This is exactly what Laravel props are for, allowing you to easily set default values and use them more liberally around the component. Especially because we don't need `attribute` directly set on the span anyway, I don't see why we should do it any other way.

Jade-GG: I also do find it a bit weird to have a default attribute here. Some resources may have a `title` instead of a name. It doesn't seem logical to have a default for that when it could be anything.

resources/css/app.css
Jade-GG: That does work, for sure. The question here would then be do we want to add all this into the highlight component, as it will get pretty messy (IMO), for example:

```blade
@props(['attribute', 'item' => 'item', 'highlightTag' => 'mark', 'highlightClasses' => 'bg-transparent font-bold'])

@if (in_array($attribute, config('rapidez.searchkit.highlight_attributes')))
    <ais-highlight
        {{ $attributes->merge([
            'v-bind:hit' => $item,
            'attribute' => $attribute,
            'highlighted-tag-name' => $highlightTag,
            'v-bind:class-names' => "{ 'ais-Highlight-highlighted': '$highlightClasses' }"
        ]) }}
    />
@else
    <span {{ $attributes }} v-text="{{ $item }}.{{ $attribute }}"></span>
@endif
```

It does give slightly more power to those using the component to do whatever they want with the styling. You do now have to override this component if you just want to change the default styling, however.

resources/views/components/highlight.blade.php
royduin: Thanks for checking! 🚀

PR #771, opened 2025-03-10
Refactor searchable models

indykoning:
This PR refactors the searchable models a little.
Removing code that is no longer needed, adressing some todos, reducing model size

### DIFF
--- a/src/Models/Category.php
+++ b/src/Models/Category.php
@@ -104,16 +104,11 @@ class Category extends Model
             ->get();
     }
 
+    /**
+     * @inheritdoc
+     */
     protected function makeAllSearchableUsing(Builder $query)
     {
-        // TODO: Is this filter still useful? You could override and
-        // extend the existing model which gives you full control.
-        // But... that will be applied always, this one is just
-        // for the index. Another option could be to have 2
-        // category models; a default and one specificly
-        // for the index: Models/IndexCategory.php
-        // directly handy to give these Scout
-        // methods their own place...
         return $query->withEventyGlobalScopes('index.categories.scopes')
             ->select((new (config('rapidez.models.category')))->qualifyColumns(['entity_id', 'name']))
             ->whereNotNull('url_key')
--- a/src/Models/Product.php
+++ b/src/Models/Product.php
@@ -8,8 +8,6 @@ use Illuminate\Database\Eloquent\Relations\BelongsToMany;
 use Illuminate\Database\Eloquent\Relations\HasMany;
 use Illuminate\Database\Eloquent\Relations\HasOne;
 use Illuminate\Database\Eloquent\Relations\HasOneThrough;
-use Illuminate\Support\Arr;
-use Illuminate\Support\Facades\Cache;
 use Rapidez\Core\Casts\Children;
 use Rapidez\Core\Casts\CommaSeparatedToArray;
 use Rapidez\Core\Casts\CommaSeparatedToIntegerArray;
@@ -26,7 +24,7 @@ use Rapidez\Core\Models\Traits\HasAlternatesThroughRewrites;
 use Rapidez\Core\Models\Traits\Product\CastMultiselectAttributes;
 use Rapidez\Core\Models\Traits\Product\CastSuperAttributes;
 use Rapidez\Core\Models\Traits\Product\SelectAttributeScopes;
-use Rapidez\Core\Models\Traits\Searchable;
+use Rapidez\Core\Models\Traits\Product\Searchable;
 use TorMorten\Eventy\Facades\Eventy;
 
 class Product extends Model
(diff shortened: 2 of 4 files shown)


src/Models/Product.php
```diff
@@ -259,105 +257,4 @@ public static function exist($productId): bool
...
-            ->mapWithKeys(fn ($position, $category_id) => [$category_id => $maxPositions[$category_id] - $position]);
-
-        return $data;
-    }
-
-    // TODO: With this and all methods above this model
-    // is getting pretty big... can we move this?
-    public function withCategories(array $data): array
-    {
-        // TODO: Do we still need to do this in this way? Can we refactor this to be nicer now that we're using Scout?
```
indykoning, line 317: We could solve this using a relationship on the product. But it would be less efficient

src/Models/Traits/Product/Searchable.php
```diff
@@ -0,0 +1,117 @@
...
+     * Add the category paths
+     */
+    public function withCategories(array $data): array
+    {
+        $categories = Cache::driver('array')->rememberForever('categories-' . config('rapidez.store'), function () {
+            return Category::withEventyGlobalScopes('index.category.scopes')
+                ->where('catalog_category_flat_store_' . config('rapidez.store') . '.entity_id', '<>', config('rapidez.root_category_id'))
+                ->pluck('name', 'entity_id');
+        });
+
```
indykoning, line 88: Remove the `>>>` structure as it has now been changed into the `category_lvlx` fields

src/Models/Traits/Searchable.php
```diff
@@ -2,42 +2,24 @@
...
     use ScoutSearchable;
 
+    /**
+     * {@inheritdoc}
+     */
     public function toSearchableArray(): array
     {
-        $data = $this->searchableData($this->toArray(), $this);
-
-        // TODO: Is this filter still useful? Overriding the model gives you full control
```
indykoning, line 20: Overriding the model should be used in projects. However packages do not have this luxury.
This is why i believe it's better to keep this.

src/Models/Traits/Searchable.php
```diff
@@ -2,42 +2,24 @@
...
-        return Eventy::filter('index.' . config('rapidez.index') . '.data', $data, $this);
-    }
-
-    // TODO: Not sure if this is the best idea, we already
-    // have the toSearchableArray for this.
```
indykoning, line 25: Agreed, i've removed this function. In favor of overriding `toSearchableArray`

src/Models/Traits/Searchable.php
```diff
@@ -51,14 +33,4 @@ public function searchableAs(): string
...
-
-    protected function makeAllSearchableUsing(Builder $query)
-    {
-        return $query;
-    }
-
-    public function makeSearchableUsing(Collection $models): Collection
-    {
-        return $models;
-    }
```
indykoning, line 63: These functions were not used other than in the actual model. They could be removed.

src/Models/Traits/Searchable.php
```diff
@@ -51,14 +34,4 @@ public function searchableAs(): string
...
```
indykoning, line 63: These functions were not used other than in the models themselves. They could be removed here.

src/Models/Category.php
```diff
@@ -104,16 +104,11 @@ public function getParentcategoriesAttribute()
             ->get();
     }
 
+    /**
+     * {@inheritdoc}
```
indykoning, line 108: I've used the inheritdoc to indicate these are overriding functions of the Searchable trait

src/Models/Traits/Searchable.php
```diff
@@ -2,42 +2,25 @@
...
-        // Return the array with data that you want to send to ES.
-
-        return $data;
-    }
-
-    public function shouldBeSearchable(): bool
-    {
-        return true;
+        // TODO: Maybe use the model name here?
+        return Eventy::filter('index.' . config('rapidez.index') . '.data', $this->toArray(), $this);
```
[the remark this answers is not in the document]
royduin: Shouldn't this be within `toSearchableArray()` as it's to manipulate data? And `rapidez.index` isn't / shouldn't be used right? See: https://github.com/rapidez/core/pull/764

Yeah I think we do have 3 options there:

1. Have a custom function name, like `searchableData()` previously which isn't a Scout thing: https://laravel.com/docs/12.x/scout#configuring-searchable-data
2. Stick with Scout and if you need to overwrite something you should use `parent::toSearchableArray()` to keep the filter
3. Handle it somewhere else; for example within `scoutMetadata()`? All drivers, including `matchish/laravel-scout-elasticsearch` do merge that with the data from `toSearchableArray()`, see: https://github.com/matchish/laravel-scout-elasticsearch/blob/e67bcdba248291cecb203d3e6d37569a75dc4bf9/src/ElasticSearch/Params/Bulk.php#L62

src/Models/Product.php
```diff
@@ -259,105 +257,4 @@ public static function exist($productId): bool
...
-        $data['positions'] = $product->categoryProducts
-            ->pluck('position', 'category_id')
-            // Turn all positions positive
-            ->mapWithKeys(fn ($position, $category_id) => [$category_id => $maxPositions[$category_id] - $position]);
-
-        return $data;
-    }
-
-    // TODO: With this and all methods above this model
-    // is getting pretty big... can we move this?
```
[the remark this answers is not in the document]
royduin: Shouldn't we do the same for the category?

src/Models/Traits/Searchable.php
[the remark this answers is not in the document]
indykoning: > Stick with Scout and if you need to overwrite something you should use parent::toSearchableArray() to keep the filter

That is what i've decided to go with here, it makes it easier to manage for someone who is not familiar with Rapidez' implementation since it does not hide any trickery.

I will go ahead and implement the index name

src/Models/Product.php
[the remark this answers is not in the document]
indykoning: We certainly can if need be, the category model currently overrides a single function of the Searchable trait. 
I didn't consider it to be causing enough bloat to necessitate it's own trait

PR #772, opened 2025-03-14
[3.x] Make sure JSON can't be empty

Jade-GG:
If the `additional_data` column is `EMPTY` (or invalid JSON) instead of `NULL`, this query throws an error. By adding in this check we can fix this.


indykoning (approved):

royduin:
Checking for an empty string is less heavy than validating json. Do we really need to validate it?

Jade-GG:
> Checking for an empty string is less heavy than validating json. Do we really need to validate it?

I chose to validate here only to make sure you can never get an error from invalid json. I could also check for an empty string, but I wanted to cover all bases :)

indykoning:
I would assume it does have some impact in performance over a simple string check, however i do not know how much the impact is.

PR #774, opened 2025-03-17
Added config for success fail redirect path

kevinmeijer97:
Added some config to change the redirect path in case of a failed checkout.

### DIFF
--- a/config/rapidez/frontend.php
+++ b/config/rapidez/frontend.php
@@ -92,4 +92,7 @@ return [
         'auto_select_configurable_options' => false,
         'auto_select_product_options'      => false,
     ],
+
+    // The path to redirect to after a failed checkout
+    'checkout_success_fail_redirect_path' => 'cart',
 ];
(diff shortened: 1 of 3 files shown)


royduin:
Maybe we should move all checkout configs to it's own file? @indykoning? Currently it's mixed:

- https://github.com/rapidez/core/blob/a554a07fe6ed9eb558cdf243aa7625d291c8e477/config/rapidez/frontend.php#L28
- https://github.com/rapidez/core/blob/a554a07fe6ed9eb558cdf243aa7625d291c8e477/config/rapidez/system.php#L31

And should we use named routes here?

indykoning:
That might make the config nicer, do we see any other configs being added for the checkout in the future? 
If it's just the 2 this is fine, if it becomes more maybe we should do the separate config.

kevinmeijer97:
Closed for https://github.com/rapidez/core/pull/780

PR #781, opened 2025-03-21
[Instantsearch] Select super attributes when filtering by them

Jade-GG:
Not sure if I'm fully happy with how this has been written. However, I was unable to find any other way to retrieve the current list of refinements nicely inside a component.

Injecting the `instantSearchInstance` this way is how the instantsearch package itself also does it. See:
- Provide: https://github.com/algolia/instantsearch/blob/d16fb1d64212db0cff7e855a87936c0bbf2c7ce3/packages/vue-instantsearch/src/util/createInstantSearchComponent.js#L16
- Inject example: https://github.com/algolia/instantsearch/blob/d16fb1d64212db0cff7e855a87936c0bbf2c7ce3/packages/vue-instantsearch/src/mixins/widget.js#L10

Would rather not use the `widget` mixin linked here as that would add a lot of things we don't need.

For now I've made it so that, if you select multiple options from the same attribute, it doesn't pre-select anything. This is the simplest solution that also makes sense UX-wise.


indykoning (approved):

royduin:
`SuperAttribute.vue` vs within the `AddToCart.vue` where we're already handling related stuff?

indykoning:
I have managed to get it working in #797, it will merge into this branch currently

PR #782, opened 2025-03-24
[3.x] Config defaults

Jade-GG:
I've gone through the magento `config.xml` files and grabbed all of the config values that are currently being used by Rapidez or Rapidez packages, but are not from a third-party magento extension. Subsequently, I removed all the hardcoded defaults from all the `Rapidez::config` calls.

I also removed the `show_customer_address_fields` part because it is not being used anymore.


royduin:
Looking good! But the tests are failing? And this solution vs using https://developer.adobe.com/commerce/webapi/graphql/schema/store/queries/store-config/, it looks like it can return a lot of configurations: https://developer.adobe.com/commerce/webapi/graphql-api/index.html#query-storeConfig

Jade-GG:
> Looking good! But the tests are failing?

Interesting, the error makes little sense for what I've changed. I'll look into it.

> And this solution vs using https://developer.adobe.com/commerce/webapi/graphql/schema/store/queries/store-config/, it looks like it can return a lot of configurations: https://developer.adobe.com/commerce/webapi/graphql-api/index.html#query-storeConfig

You can't get the actual defaults from what I can see there, which is what we're looking for here.

royduin:
+ we should double check this now https://github.com/rapidez/core/pull/738 is merged

royduin:
> You can't get the actual defaults from what I can see there, which is what we're looking for here.

The actual value is also fine, maybe even better. For the standalone checkout https://github.com/rapidez/core/pull/711 it would be nice to remove the database access requirement. Plus with this we don't have to keep a list of defaults. We can do the query from the PHP side and cache the result? But not sure if we can use it for all configs?

Jade-GG:
> But not sure if we can use it for all configs?

Just did a check, you cannot. Things you can't use it for include checking the flat table configuration, checking if backorders are enabled, seeing which customer address values we want to have on the frontend, seeing if guests can subscribe to the newsletter, and more---I didn't even check everything here. That's not even mentioning the fact that it's impossible to get a config value from extensions.

Basically, we would have to add functionality to Compadre (or a standalone Magento extension) which we don't really want to do here because then we would *require* that for basic Rapidez functionality.

royduin:
Alrighty, todo:

- Check the tests
- Fixed the conflict
- Link to the config file from the docs so we can remove the whole table of Magento configs: https://docs.rapidez.io/3.x/configuration.html#magento-configuration

src/Models/Scopes/Product/WithProductStockScope.php
```diff
@@ -24,7 +24,7 @@ public function apply(Builder $builder, Model $model)
         // - use_config_qty_increments
 
         $builder
-            ->selectRaw('ANY_VALUE(IF(cataloginventory_stock_item.use_config_backorders, ' . $configBackorder . ', cataloginventory_stock_item.backorders)) as backorder_type')
+            ->selectRaw('ANY_VALUE(IF(cataloginventory_stock_item.use_config_backorders, ' . ($configBackorder ?: '0') . ', cataloginventory_stock_item.backorders)) as backorder_type')
```
Jade-GG, line 27: This one is kind of stupid. The correct magento default is an empty string, however we can't put an empty string into the query. Hence this fallback.

royduin:
Let's wait for @indykoning his review. When this is going to be merged we also need to merge https://github.com/rapidez/docs/pull/85

indykoning (approved):

PR #792, opened 2025-04-01
Magento link widgets

Tjeerd-JW:
Added extra magento widgets.
Pagelink, productlink and categorylink.
Also added link.blade.php wich is used to show the links.


src/Widgets/CategoryLink.php
```diff
@@ -0,0 +1,29 @@
...
+{
+    protected $idPath;
+
+    protected $title;
+
+    protected $anchorText;
+
+    public function __construct($vars)
+    {
+        $this->idPath = str($vars->id_path)->after('/')->prepend('catalog/category/view/id/');
```
royduin, line 15: In Magento this seems to go through the rewrite table: https://github.com/magento/magento2/blob/2.4-develop/app/code/Magento/Catalog/Block/Widget/Link.php, advantage of that is we can get the actual url. No need for that right now as there will be a canonical but it also seems possible to have a `id_path` with a 3th parameter: `product/1/3`, will that work?

royduin (changes requested):

src/Widgets/CategoryLink.php
royduin: Seems like we currently don't support those product combined with category urls.

- Works: https://demo.rapidez.io/catalog/product/view/id/2
- Doesn't work: https://demo.rapidez.io/catalog/product/view/id/2/category/3

But within Magento it does work: https://magento-demo.rapidez.io/catalog/product/view/id/2/category/3, maybe something for a separated PR

src/Widgets/CategoryLink.php
```diff
@@ -0,0 +1,29 @@
...
+        $this->title = $vars->title;
+        $this->anchorText = $vars->anchor_text;
```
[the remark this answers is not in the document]
Tjeerd-JW: I added fallbacks to every widget so when their not set it won't give an error.

src/Widgets/ProductLink.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+namespace Rapidez\Core\Widgets;
+
+class ProductLink
```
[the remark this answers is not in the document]
Tjeerd-JW: I made a new widget ProductAndCategoryLink.php combining the 2 using match.

src/Widgets/CategoryLink.php
Tjeerd-JW: The extra /category/id doesn't seem to effect where it leads you, 
https://magento-demo.rapidez.io/catalog/product/view/id/2/category/3456789 
still leads to the same place as
https://magento-demo.rapidez.io/catalog/product/view/id/2/category/1 
Maybe a build in redirect that removes /category/id? But that seems like something for a separate PR. 
If id_path does have a 3th parameter it will add it to the back of the url but the url wont work.

src/Widgets/PageLink.php
```diff
@@ -0,0 +1,37 @@
...
+    {
+        if ($pageContent = Page::withoutGlobalScope(IsActiveScope::class)->find($vars->page_id)) {
+            $this->identifier = $pageContent['identifier'];
+            $this->anchorText = $vars->anchor_text ?? $pageContent['title'] ?? 'link';
+            $this->title = $vars->title ?? $pageContent['title'] ?? 'title';
+        } else {
+            $this->identifier = '';
+            $this->anchorText = $vars->anchor_text  ?? 'link';
+            $this->title = $vars->title ?? 'title';
+        }
```
royduin, line 26:
```suggestion
        $page = Page::withoutGlobalScope(IsActiveScope::class)->find($vars->page_id)

        $this->identifier = $page->identifier ?? '/';
        $this->anchorText = $vars->anchor_text ?? $page->title ?? 'link';
        $this->title = $vars->title ?? $page->title ?? '';
```

src/Widgets/ProductAndCategoryLink.php
```diff
@@ -0,0 +1,32 @@
...
+
+    public function __construct($vars)
+    {
+        $this->idPath = match (str($vars->id_path)->before('/')->value()) {
+            'category' => str($vars->id_path ?? '')->after('/')->prepend('catalog/category/view/id/'),
+            'product' => str($vars->id_path ?? '')->after('/')->prepend('catalog/product/view/id/'),
+            default => 'unknown'
+        };
+        $this->title = $vars->title ?? 'title';
+        $this->anchorText = $vars->anchor_text ?? 'link';
```
royduin, line 21:
```suggestion
        $type = str($vars->id_path)->before('/')->value();
        $id = str($vars->id_path)->after('/');

        $this->idPath = match ($type) {
            'category' => $id->prepend('catalog/category/view/id/'),
            'product' => $id->prepend('catalog/product/view/id/'),
        };

        $this->title = $vars->title ?? '';
        $this->anchorText = $vars->anchor_text ?? 'link';
```

src/Widgets/PageLink.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
+namespace Rapidez\Core\Widgets;
+
+use Rapidez\Core\Models\Page;
+use Rapidez\Core\Models\Scopes\IsActiveScope;
+
+class PageLink
+{
+    protected $identifier;
```
royduin, line 10: Maybe call this `$url` as that's what goes into it.

src/Widgets/ProductAndCategoryLink.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+namespace Rapidez\Core\Widgets;
+
+class ProductAndCategoryLink
+{
+    protected $idPath;
```
royduin, line 7: Maybe call this `$url` as that's what goes into it.

royduin (changes requested):

src/Widgets/CategoryLink.php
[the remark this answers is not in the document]
royduin: Nice, but as with the pages: if there isn't a title or anchor text set this should come from the product/category.

src/Widgets/CategoryLink.php
Tjeerd-JW: `        $vars->id_path = collect(explode('/', $vars->id_path))->take(2)->implode('/');
`
Now if there's anything after product/1 it removes it and I'll make another pr to support the redirects.

royduin: Thanks! 🚀

src/Widgets/CategoryLink.php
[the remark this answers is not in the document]
royduin: I think this is the last thing left to do.

routes/web.php
```diff
@@ -10,6 +10,9 @@
 
 Route::middleware('web')->group(function () {
     Route::get('catalog/product/view/id/{productId}', [config('rapidez.routing.controllers.product'), 'show'])->whereNumber('productId');
+    Route::get('catalog/product/view/id/{productId}/{any?}', function ($productId) {
+        return redirect('catalog/product/view/id/'.$productId);
```
royduin, line 14:
```suggestion
        return redirect('catalog/product/view/id/'.$productId, 301);
```

royduin:
Thanks! 🚀🚀

PR #794, opened 2025-04-02
Made the variables used within the graphql component reactive

indykoning:
When trying to add a v-model to the variables of the graphql component in the same way as done with the graphql-mutation component they actually get emptied when the query gets sent.

I've found out variables isn't actually reactive for the graphql component.
This makes it work reactive since the value is updated in the data.

I've purposefully left out the "watch" prop as that seems to cause a kind of two way binding.
Where suddenly changes to dataVariables are mirrored to variables, but when runQuery gets called both dataVariables and variables get reset to the initial value.

### DIFF
--- a/resources/js/components/Graphql.vue
+++ b/resources/js/components/Graphql.vue
@@ -50,7 +50,7 @@ export default {
             return null
         }
 
-        return this.$scopedSlots.default({...this, variables: this.dataVariables})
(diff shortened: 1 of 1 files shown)


royduin:
This won't case any issues with the current usage everywhere? The `watch="false"` from the GraphQL Mutation component is used frequently: https://github.com/search?q=org%3Arapidez%20watch%3D%22false%22&type=code

PR #795, opened 2025-04-04
Added additional data to orderV2 graphql

indykoning:
Note that Magento by default has it hardcoded empty, allowing for plugins to add data (Like purchase order number).
So this will go mostly unused, but i did have a project that wanted it.


royduin:
Do we also need this in the checkout theme and confira?

PR #796, opened 2025-04-04
Lazily load instantsearch globally

indykoning:
In this PR we've moved the components for instantsearch to be lazily imported, this makes it easier to create your own listings without having to register them again. And prepares us for a Vue 3 update

### DIFF
--- a/resources/js/components/Search/Autocomplete.vue
+++ b/resources/js/components/Search/Autocomplete.vue
@@ -1,14 +1,23 @@
 <script>
-import AisSearchBox from 'vue-instantsearch/vue2/es/src/components/SearchBox.vue.js'
-import AisIndex from 'vue-instantsearch/vue2/es/src/components/Index.js'
-
 import InstantSearchMixin from './InstantSearchMixin.vue'
+
+import InstantSearch from 'vue-instantsearch/vue2/es/src/components/InstantSearch'
+import Hits from 'vue-instantsearch/vue2/es/src/components/Hits.js'
+import Configure from 'vue-instantsearch/vue2/es/src/components/Configure.js'
+import highlight from 'vue-instantsearch/vue2/es/src/components/Highlight.vue.js'
+import SearchBox from 'vue-instantsearch/vue2/es/src/components/SearchBox.vue.js'
+import Index from 'vue-instantsearch/vue2/es/src/components/Index.js';
+
 
 export default {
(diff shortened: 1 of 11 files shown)


resources/js/components/Search/Autocomplete.vue
```diff
@@ -1,14 +1,22 @@
...
     mixins: [InstantSearchMixin],
     components: {
-        'ais-search-box': AisSearchBox,
-        'ais-index': AisIndex,
+        InstantSearch,
+        Hits,
+        Configure,
+        highlight,
+        SearchBox,
+        Index,
```
indykoning, line 19: we must still USE these components in order to tell the bundling to load these at the same time as the autocomplete.
If we don't we re-introduce the horrible flashing as can be seen by the downloads
image

In it's state where we still import _and use_ the components the downloads look like this:
image

src/RapidezServiceProvider.php
```diff
@@ -275,6 +275,25 @@ protected function bootMacros(): self
...
+        Vite::macro('getPathsByFilenames', function ($filenames) {
+            /** @var \Illuminate\Foundation\Vite $this */
+            $filenames = is_array($filenames) ? $filenames : func_get_args();
+            $manifest = $this->manifest($this->buildDirectory);
+
+            return array_filter(
+                array_map(
+                    function ($filename) use ($manifest) {
+                        foreach ($manifest as $path => $asset) {
+                            if (Str::endsWith($asset['name'] ?? '', $filename) || Str::endsWith($path, $filename)) {
```
indykoning, line 287: Instantsearch works with `.vue.js` files, checking the name ensures an import with `.vue` in the name still gets caught

resources/views/listing/products.blade.php
```diff
@@ -1,4 +1,18 @@
...
+@pushOnce('head', 'listing-products')
+    @vite(
+        vite_filename_paths([
+            'HitsPerPage.vue',
+            'SortBy.vue',
+            'Stats.vue',
+            'Hits.vue',
+            'Highlight.vue',
+            'Pagination.vue',
+            'Hits'
```
Jade-GG, line 11: Is this correct? `Hits.vue` and `Hits`?

Jade-GG (commented):
Can you take a look at the newly created phpstan errors?

PR #797, opened 2025-04-07
Move refinement attributes to addToCart.vue

indykoning:
This is another take on the automatically selecting attributes, moving it entirely to the add to cart.
It is also available in a computed object so it could be used to show the matching attributes first in the dropdowns in the future.

~~The changed files do not look correct from my end, the changer are here~~

~~[`906de50` (#797)](https://github.com/rapidez/core/pull/797/commits/906de502607c83b85b3c7285b54fd80ed4ab22f4)~~
Fixed


resources/js/components/Product/AddToCart.vue
```diff
@@ -3,6 +3,21 @@ import { GraphQLError } from '../../fetch'
...
 
 export default {
+    inject: {
+        instantSearchInstance: {
+            from: '$_ais_instantSearchInstance',
+            default: () => {
+                return {
+                    helper: {
+                        state: {
+                            disjunctiveFacetsRefinements: {},
```
indykoning, line 13: This is the default value for when the addToCart is not within context of an `instantSearchInstance`

resources/views/listing/partials/item/super_attributes.blade.php
indykoning, line 1: This file has been reset to the state before https://github.com/rapidez/core/pull/781/files#diff-bc56f0bb8882408b8979141aca087618a8de0b1fc576db77c0d30d23438f4991

Jade-GG (approved):

Jade-GG:
I assume this also then works when you go to the next page, for example?

PR #800, opened 2025-04-08
Move duplicate autocomplete parts to component

indykoning:
Unfortunately we do still need the additional input to prevent the AIS components from downloading prematurely. This way there's a consistent place to modify this.

### DIFF
--- a/resources/views/components/autocomplete/input.blade.php
+++ b/resources/views/components/autocomplete/input.blade.php
@@ -1,0 +1,27 @@
+@props([
+    'id' => 'autocomplete'
+])
+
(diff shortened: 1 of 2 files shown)


Jade-GG (approved):

resources/views/components/autocomplete/input.blade.php
```diff
@@ -0,0 +1,21 @@
+@props([
+    'id' => 'autocomplete'
+])
```
royduin, line 3:
```suggestion
@props(['id' => 'autocomplete'])
```

PR #801, opened 2025-04-08
[Instantsearch] Remove filterScoreScript

Jade-GG:
We could also remove the filterQueryString here and also replace that with a function we can call inside of the baseFilters, and maybe by that point we could also aggregate those functions into their own file? Not sure what would be nicer here.

Would look something like this:
```
v-bind:base-filters="() => [
    queryString('visibility:(2 OR 4) AND category_ids:{{ $category->entity_id }}')
    categoryPositions({{ $category->entity_id }}),
]"
```

### DIFF
--- a/resources/js/package.js
+++ b/resources/js/package.js
@@ -139,6 +139,18 @@ function init() {
 
                     return `/storage/${store}/resizes/${size}/magento${url}`
                 },
+
+                categoryPositions(categoryId) {
+                    return {
+                        function_score: {
+                            script_score: {
+                                script: {
+                                    source: `Integer.parseInt(doc['positions.${categoryId}'].empty ? '0' : doc['positions.${categoryId}'].value)`
+                                }
+                            }
+                        }
+                    }
+                },
             },
             computed: {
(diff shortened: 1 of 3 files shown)


resources/js/package.js
```diff
@@ -139,6 +139,18 @@ function init() {
 
                     return `/storage/${store}/resizes/${size}/magento${url}`
                 },
+
+                categoryPositions(categoryId) {
```
royduin, line 143: It's only used on listings, can't we have it in the `Listing.vue`?

Jade-GG:
> It would be nice to have a `categoryId` prop that handles this. If you don't want it; just don't specify it and if you'd like to have something custom you can still do that with a custom query.

Done!

resources/js/package.js
royduin: This one can be moved back to the `Listing.vue` after the recent changes?

royduin:
What do you think of https://github.com/rapidez/core/pull/801#discussion_r2048431991 @indykoning? Merging for now. We can change it later.

PR #802, opened 2025-04-08
[3.x] Use query params through javascript to better support static caching

BobWez98:
This grabs the query search parameter through javascript instead of the laravel helper to better support static caching. It also uses this logic to determine if we're on the search page to allow the title of the document to be set.

Once approved I will create a 2.x PR.

### DIFF
--- a/resources/js/package.js
+++ b/resources/js/package.js
@@ -157,6 +157,10 @@ function init() {
                 canOrder() {
                     return this.cart.items.every((item) => item.is_available)
                 },
+
+                queryParams() {
+                    return new URLSearchParams(window.location.search)
+                }
             },
             watch: {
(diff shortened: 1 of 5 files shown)


resources/js/package.js
```diff
@@ -157,6 +157,10 @@ function init() {
                 canOrder() {
                     return this.cart.items.every((item) => item.is_available)
                 },
+
+                queryParams() {
+                    return new URLSearchParams(window.location.search)
```
royduin, line 162: This is reactive? It changes as we type?

royduin: Nevermind, not needed. From InstantSearch it will be reactive, see: https://github.com/rapidez/core/pull/799

PR #804, opened 2025-04-10
[2.x] Add debounce to gql mutation

Jade-GG:
Allows you to add a debounce to graphql mutation queries. Useful, for example, when people are likely to spam a button (e.g. a `+` button in the cart).


royduin:
This vs disabling the button when mutating?

Jade-GG:
> This vs disabling the button when mutating?

Disabling the button means you can't click the `+` 10 times to increase your quantity quickly. You'd have to wait in between every click and that gets extremely tiring for customers.

PR #805, opened 2025-04-10
[3.x] Add debounce to graphql mutation

Jade-GG:
See #804

### DIFF
--- a/resources/js/components/GraphqlMutation.vue
+++ b/resources/js/components/GraphqlMutation.vue
@@ -59,6 +59,10 @@ export default {
             type: Boolean,
             default: false,
         },
+        debounce: {
+            type: Number,
+            default: 0,
+        },
     },
 
     data: () => ({
@@ -67,6 +71,7 @@ export default {
         running: false,
         initialVariables: {},
         data: {},
+        mutate: () => null,
     }),
 
     render() {
@@ -84,6 +89,12 @@ export default {
     created() {
         this.initialVariables = JSON.parse(JSON.stringify(this.variables))
         this.data = this.variables
+
+        if (this.debounce) {
+            this.mutate = useDebounceFn(this.mutateFn, this.debounce)
+        } else {
+            this.mutate = this.mutateFn
+        }
     },
 
     watch: {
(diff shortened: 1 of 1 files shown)


resources/js/components/GraphqlMutation.vue
```diff
@@ -105,7 +116,7 @@ export default {
     },
 
     methods: {
-        async mutate() {
+        async mutateFn() {
```
indykoning, line 119: This might cause problems with 
https://docs.rapidez.io/3.x/graphql-components.html#:~:text=partial%2Dsubmit%20must%20always%20be%20on%20the%20direct%20child%20of%20a%20Vue%20component%20and%20have%20its%20original%20function%20name.

Could you test to make sure this causes no problems with partials-submit?

resources/js/components/GraphqlMutation.vue
```diff
@@ -84,6 +89,12 @@ export default {
     created() {
         this.initialVariables = JSON.parse(JSON.stringify(this.variables))
         this.data = this.variables
+
+        if (this.debounce) {
+            this.mutate = useDebounceFn(this.mutateFn, this.debounce)
```
indykoning, line 94: Doe this method also return a promise that only resolves once mutateFn is resolved?

Otherwise this may cause issues with functions wanting to await or chain onto the mutate function e.g. https://github.com/rapidez/core/blob/97d2b62daf7953c313aa7d59e023e174ef3775fe/resources/views/checkout/steps/shipping_address.blade.php#L19

Jade-GG: Yes, `useDebounceFn` actually returns a promise by default (see docs https://vueuse.org/shared/useDebounceFn/) and it looks to also properly wait for the async function inside to finish running before resolving.

There does seem to be some weirdness with it also resolving continuously while you're spamming the button. This may be a vueUse bug because I don't really see why this would happen. I don't know if this will affect any current use cases though.

resources/js/components/GraphqlMutation.vue
Jade-GG: Yes, this works fine in this case because the actual underlying function name within `__vue__` stays the same. From what I can tell the issue that happens in the case shown in the docs is that the name within the render itself is different, where then the `submitPartials` function looks for that function name within the vue component and comes up blank.

indykoning (approved):

PR #806, opened 2025-04-10
[3.x] Maintain a list of fetches to ensure exact loading count

indykoning:
When navigating away while a request is in flight it may reduce the loadingcount by 1 when it finishes the fetch.

Using a map of fetches like this we can make the fetch calls cross themselves off the list.
If a fetch returns which is no longer on the list (by navigating away) it will not reduce the loadingCount anymore


Jade-GG:
Can we use a store here instead, so we don't have to worry about Vue being loaded in at all? Right now it's kind of weird working around Vue not yet being loaded in with two different hacks. It would probably be cleaner if we can do this outside of the Vue VM.

indykoning:
Suddenly we're running into 404s for the pages, did something happen to the core / actions?

Jade-GG:
Dusk tests look to be fixed now 👍

PR #807, opened 2025-04-10
[2.x] Maintain a list of fetches to ensure exact loading count

indykoning:
See: #806


indykoning:
The useFetches store seems to be doing it's work well!

Jade-GG (approved):
Nice and simple like this, lgtm!

PR #813, opened 2025-04-17
Display search suggestions if no results are available

indykoning:
This will add a list of search suggestions when there are no results to be shown.

So long as there are search queries in the database it will always have suggestions. This has been forced by using the function score. 

Instead of using a MUST on the results, we boost matching queries so suggestions matching your query will be sorted first.
An additional weight is added based on popularity (which will not outrank the search query)
and the existence of a redirect url on this query, indicating it has gotten special attention

### DIFF
--- a/resources/js/components/Listing/SearchSuggestions.vue
+++ b/resources/js/components/Listing/SearchSuggestions.vue
@@ -1,0 +1,89 @@
+<script>
+import InstantSearchMixin from '../Search/InstantSearchMixin.vue'
+
+export default {
+    mixins: [InstantSearchMixin],
+
+    render() {
+        return this.$scopedSlots.default(this)
+    },
+
+    mounted() {
+        this.$nextTick(() => {
+            requestAnimationFrame(() => this.$emit('mounted'))
+        })
+    },
+
+    methods: {
+        async getInstantSearchClientConfig() {
+            const config = await InstantSearchMixin.methods.getInstantSearchClientConfig.bind(this).call()
+
+            config.getQuery = (query, search_attributes) => {
+                return [
+                    {
+                        function_score: {
+                            functions: [
+                                {
+                                    filter: { match: { query_text: query } },
+                                    weight: 18
+                                },
+                                {
+                                    filter: { term: { redirect: query } },
+                                    weight: 24
+                                },
+                                {
+                                    field_value_factor: {
+                                        field: "popularity",
+                                        factor: 1,
+                                        modifier: "log1p",
+                                        missing: 0
+                                    },
+                                    weight: 5
+                                },
+                                {
+                                    filter: { exists: { field: "redirect" } },
+                                    weight: 10
+                                },
+                                {
+                                    filter: { term: { display_in_terms: 1 } },
+                                    weight: 2
+                                },
+                                {
+                                    filter: { term: { is_processed: 1 } },
+                                    weight: 2
+                                }
+                            ]
+                        },
+                    },
+                ]
+            }
+
+            return config
+        },
+
+        async getSearchSettings() {
+            return {
+                highlight_attributes: ['query_text'],
+                search_attributes: [
+                    {
+                        field: 'query_text',
+                        weight: 2.0,
+                    },
+                    {
+                        field: 'redirect',
+                        weight: 8.0,
+                    },
+                ],
+                result_attributes: [
+                    'query_text',
+                    'redirect',
+                    'popularity',
+                    'num_results',
+                    'display_in_terms',
+                    'is_processed',
+                ],
+            }
+        },
+    },
+}
+</script>
--- a/resources/js/components/Listing/SearchSuggestions.vue
+++ b/resources/js/components/Listing/SearchSuggestions.vue
@@ -25,34 +25,34 @@ export default {
                             functions: [
                                 {
                                     filter: { match: { query_text: query } },
-                                    weight: 18
+                                    weight: 18,
                                 },
                                 {
                                     filter: { term: { redirect: query } },
-                                    weight: 24
+                                    weight: 24,
                                 },
                                 {
                                     field_value_factor: {
-                                        field: "popularity",
+                                        field: 'popularity',
                                         factor: 1,
-                                        modifier: "log1p",
-                                        missing: 0
+                                        modifier: 'log1p',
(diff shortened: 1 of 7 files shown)


src/Http/Controllers/SearchController.php
```diff
@@ -4,12 +4,17 @@
...
+use Rapidez\Core\Models\Scopes\IsActiveScope;
 
 class SearchController
 {
     public function __invoke(Request $request)
     {
-        $searchQuery = config('rapidez.models.search_query')::firstOrNew([
+        if (! $request->q) {
+            return view('rapidez::search.overview');
+        }
```
indykoning, line 15: Make sure we don't push `""` as query text to the database.

src/Models/SearchQuery.php
```diff
@@ -2,13 +2,31 @@
...
     protected $table = 'search_query';
 
     protected $primaryKey = 'query_id';
 
     protected $guarded = [];
+
+    protected static function booting()
+    {
+        static::addGlobalScope('ForCurrentStore', fn (Builder $query) => $query->where('store_id', config('rapidez.store')));
+        static::addGlobalScope(new IsActiveScope);
```
indykoning, line 24: Only index active search queries in the current store

resources/js/components/Listing/SearchSuggestions.vue
```diff
@@ -0,0 +1,82 @@
...
+                                },
+                                {
+                                    filter: { term: { redirect: query } },
+                                    weight: 24,
+                                },
+                                {
+                                    field_value_factor: {
+                                        field: 'popularity',
+                                        factor: 1,
+                                        modifier: 'log1p',
```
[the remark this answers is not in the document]
Jade-GG: Out of curiosity, why `log1p` over just `log`? Can the value somehow be 0?

resources/js/stores/useInstantsearchMiddlewares.js
```diff
@@ -0,0 +1,9 @@
+export let instantsearchMiddlewares = []
+
+export async function addInstansearchMiddleware(middlewareFn) {
```
Jade-GG, line 3: Instansearch -> Instantsearch

Jade-GG (changes requested):

src/Models/Scopes/Product/WithProductAttributesScope.php
```diff
@@ -25,7 +25,8 @@ public function apply(Builder $builder, Model $model)
 
         $attributeModel = config('rapidez.models.attribute');
         $attributes = $attributeModel::getCachedWhere(function ($attribute) use ($model) {
-            return in_array($attribute['code'], $model->attributesToSelect);
+            return in_array($attribute['code'], $model->attributesToSelect)
+                && ! in_array($attribute['code'], ['visibility', 'sku', 'type_id', $model->getKeyName()]);
```
royduin, line 29: Why do we need this now and previously not? Can we move this to a separated PR @indykoning? cc @Jade-GG

resources/js/components/Listing/SearchSuggestions.vue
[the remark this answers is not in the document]
indykoning: Manually created search suggestions could result in popularity being 0 if they haven't actually been searched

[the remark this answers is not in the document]
Jade-GG: Those won't be indexed though right? Given this snippet:

```php
protected function makeAllSearchableUsing(Builder $query)
{
    return $query
        ->where('popularity', '>', 10);
}
```

Either way, doesn't make a difference 😄

src/Models/Scopes/Product/WithProductAttributesScope.php
indykoning: I've cherry-picked the commit and created a PR in #820

Jade-GG (approved):

resources/views/components/productlist.blade.php
```diff
@@ -4,16 +4,16 @@
...
             v-cloak
-            v-slot="{ loaded, index, searchClient }"
+            v-slot="{ loaded, index, searchClient, middlewares }"
         >
             <div>
                 <ais-instant-search
                     v-if="searchClient"
                     :search-client="searchClient"
                     :index-name="index"
+                    :middlewares="middlewares"
```
royduin, line 16: Can't we assign this somewhere globally? In this PR I see this 3 times including other duplication on those places. Can we extract this somewhere nicely?

resources/js/components/Listing/SearchSuggestions.vue
```diff
@@ -0,0 +1,82 @@
...
+        this.$nextTick(() => {
+            requestAnimationFrame(() => this.$emit('mounted'))
+        })
+    },
+
+    methods: {
+        async getInstantSearchClientConfig() {
+            const config = await InstantSearchMixin.methods.getInstantSearchClientConfig.bind(this).call()
+
+            config.getQuery = (query, search_attributes) => {
```
royduin, line 21: If someone wants to tweak this, how should that be done?

indykoning: Currently this could be done by overriding the SearchSuggestions and overriding the `getInstantSearchClientConfig`

We could move this to the config file but we would need to move the entire list of functions there since popularity wouldn't be possible otherwise. So that would result in
```
'search_suggestions' => [
    'score_functions' => [
        [
              'filter' => ['match'  => ['query_text' => '%query%']],
        ],
        [
        	'field_value_factor' => [
		        'field' => 'popularity',
		        'factor' => 1,
		        'modifier' => 'log1p',
		        'missing' => 0,
		    ],
		    'weight' => 5,
        ]
    ]
],
```

`%query%` since it would need to be replaced by the actual search query.
Is it worth it to set that up, or would it be used so little that it's better to simply override the `SearchSuggestions` component?

resources/views/components/productlist.blade.php
indykoning: We can, and we do. most middlewares exist in the `useInstantsearchMiddlewares` store.
However if we want to add a middleware specifically for the Listing (Or in the case of this PR, the Autocomplete) we can push that in the component itself.

PR #815, opened 2025-04-17
Only update changed variables when watch is true

indykoning:
Due to this change it is rarely needed to disable the watch functionality since only updated values are pushed through instead of the entire object.

This is especially useful for the billing address as it can continue to update the underlying fields when same_as_shipping is checked


resources/js/components/Graphql/Base.vue
royduin, line 1: Is this the right place? They're just 2 "helpers" I think?

indykoning: Correct, this file does lay the foundation to "Merge" some functionality of both graphql components.
The functions are just helpers though.
Do you suggest creating an e.g. helpers.js which exports these functions instead?

indykoning: I've created an object helper file, i do quite like it that way over the previous way

resources/js/helpers/object.js
```diff
@@ -0,0 +1,64 @@
...
+    return Object.fromEntries(
+        Object.entries(source)
+            .map(([key, val]) => {
+                if (!target || !(key in target)) {
+                    return [key, val]
+                }
+
+                if (target[key] === val || ((val === null || val === undefined) && (target[key] === null || target[key] === undefined))) {
+                    return null
+                }
```
Jade-GG, line 20: Should add an extra check here for `NaN` because `NaN === NaN` always returns false. The current setup would cause objects that have any value equal to NaN to always return that as a changed variable.

resources/js/helpers/object.js
```diff
@@ -0,0 +1,64 @@
+/**
+ * Get a nested difference between two objects
+ * The values of source will be returned
+ *
+ * @param {Object} target
+ * @param {Object} source
+ *
+ * @returns Object
+ */
+export function objectDiff(target, source) {
```
Jade-GG, line 10: Should we care about any keys that only exist within `target` and not `source`? If we're doing a pure diff, I can see that being relevant. For example, if you now delete a key from your variables in the GraphqlMutation component then it won't get recognized.

Jade-GG (changes requested):

resources/js/helpers/object.js
indykoning: For our current usecase it's not relevant but for a full implementation as this is now a helper it could be useful.
Currently if the key is no longer available in the source it will return the key as undefined.

royduin:
- Do the docs need an update?
- Dusk tests are failing

PR #819, opened 2025-04-18
[Instantsearch] Config refactor feedback

Jade-GG:
- useSwatches has been moved to the config
- Added the store as a query parameter to the config file as well
- ConfigComposer -> ConfigController, with a small refactor
- Fixed the `filter` attribute on the `Attribute` model


royduin (approved):

indykoning (approved):

PR #824, opened 2025-04-25
Keep track of returned results for search queries

indykoning:
This is an extension on #813

It would be more valuable to show suggested results that actually return results.
We had the idea to apply the insights middleware to get this information: https://www.algolia.com/doc/api-reference/widgets/insights/vue/

Unfortunately the insights middleware does not get provided with the data we want.
We do get the _current_ result count, but not the **total** result count.

The `connectStats` connector provides us with the information we'd like.
The search query and the number of **total** results.

The insights values could still be really valuable for other analytics, thus the events will get fired on the components implementing the instantsearchMixin, which means listeners can be added on the Listing and Autocomplete.

### DIFF
--- a/resources/js/components/Search/Autocomplete.vue
+++ b/resources/js/components/Search/Autocomplete.vue
@@ -29,6 +29,27 @@ export default {
         this.$nextTick(() => {
             requestAnimationFrame(() => this.$emit('mounted'))
         })
+
+        const stateChanged = useDebounceFn((event) => {
+            const query = event?.payload?.query
+
+            if (!query) {
+                return
+            }
+
+            rapidezAPI('post', '/search', {
+                q: query,
+                results: event?.payload?.nbHits,
+            })
+        }, 3000)
+
+        this.$on('insights-event:viewedObjectIDs', (event) => {
+            if (event?.eventType !== 'search') {
+                return;
+            }
+
+            stateChanged(event);
+        })
     },
 
     methods: {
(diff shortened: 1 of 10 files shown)


resources/views/layouts/partials/header/autocomplete.blade.php
```diff
@@ -9,7 +9,7 @@ class="contents"
             v-cloak
         >
             <div class="contents">
-                <ais-configure :hitsPerPage="3" />
+                <ais-configure :hits-per-page.camel="3" />
```
Jade-GG, line 12: How come this was changed? Did the other version not work?

resources/js/components/Search/Autocomplete.vue
```diff
@@ -29,6 +29,27 @@ export default {
...
+            const query = event?.payload?.query
+
+            if (!query) {
+                return
+            }
+
+            rapidezAPI('post', '/search', {
+                q: query,
+                results: event?.payload?.nbHits,
+            })
```
Jade-GG, line 43: My only worry here is that it makes abuse of this endpoint extremely easy. Should there be some sort of safeguarding on here?

Jade-GG (approved):

resources/js/components/Search/Autocomplete.vue
indykoning: There is not really a foolproof way to prevent abuse here.
But with the latest commit we can significantly reduce abuse.

We now check the CSRF token, so the requests must come from customers who have a currently active session on the site.
It can't completely prevent abuse but it will make it more difficult.

resources/views/layouts/partials/header/autocomplete.blade.php
indykoning: Correct, `hitsPerPage` with camelCase is only possible in `.vue` or `.js`, html does not allow camelCase in it's arguments.
`:hits-per-page` did not seem to work either. `hits-per-page.camel` restored functionality back.

PR #826, opened 2025-04-28
[Instantsearch] Add frontend values to config & fix productlist

Jade-GG:
Fixes the missing config values when on a product page & fixes product listings.

### DIFF
--- a/resources/views/listing/partials/item.blade.php
+++ b/resources/views/listing/partials/item.blade.php
@@ -10,7 +10,9 @@
                 <img
                     v-if="addToCart.currentThumbnail"
                     :src="'/storage/{{ config('rapidez.store') }}/resizes/200/magento/catalog/product' + addToCart.currentThumbnail + '.webp'"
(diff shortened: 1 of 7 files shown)


resources/views/listing/partials/item.blade.php
```diff
@@ -10,7 +10,9 @@
                 <img
                     v-if="addToCart.currentThumbnail"
                     :src="'/storage/{{ config('rapidez.store') }}/resizes/200/magento/catalog/product' + addToCart.currentThumbnail + '.webp'"
-                    class="mb-3 h-48 w-full rounded-t object-contain" :alt="item.name" :loading="config.category && count <= 4 ? 'eager' : 'lazy'"
+                    class="mb-3 h-48 w-full rounded-t object-contain"
+                    :alt="item.name"
+                    :loading="config.category && count <= 4 ? 'eager' : 'lazy'"
```
Jade-GG, line 15: If we can get rid of this `config.category` check, we don't need to push it to the frontend in `CategoryController` at all. However, because we're somewhere inside a component we can't directly check `$category`. I chose not to mess with this too much in this PR.

PR #827, opened 2025-04-28
[3.x] Also show self as alternate

Jade-GG:
This is (apparently---so I'm told) very slightly better for SEO purposes.

### DIFF
--- a/src/Models/Traits/HasAlternatesThroughRewrites.php
+++ b/src/Models/Traits/HasAlternatesThroughRewrites.php
@@ -10,13 +10,13 @@ trait HasAlternatesThroughRewrites
     protected function alternates(): Attribute
     {
         return Attribute::make(
-            get: function() {
+            get: function () {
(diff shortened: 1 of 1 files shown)


royduin:
Doesn't this result in alternate tags on single language stores?

Jade-GG:
> Doesn't this result in alternate tags on single language stores?

Good point, but I don't know if this is necessarily problematic?

Weirdly it would give me (as a human being) more information than if there were no alternate tags, because then I can rule out the possibility of there being other languages but they just forgot to add alternate tags (if that makes any sense). However I doubt the machines interpreting this data would work the same way.

royduin:
If it doesn't add any value, just don't show it?

Jade-GG:
> If it doesn't add any value, just don't show it?

I've changed it now to first check if there are any alternates of different stores before it returns anything.

indykoning (approved):

PR #828, opened 2025-04-30
Implement partial reindexing

indykoning:
This PR adds API endpoints to Rapidez to partially reindex searchable models

By default calling the index controller (`api/admin/index/product`) will result in it updating only for the current store.
Stores can be manually set using `{"stores": [1,2,3]}` or to reindex all stores in 1 api call `{"stores": "*"}`

You are required to add "ids" to the request, this will use the qualifiedKeyName so it does not matter wether the database has `entity_id`, `id`, `order_id`, etc. as primary key

There's also a new command `rapidez:index:update` which aims to update changed data on the current index.
It can only be called if the full indexer has been run since it will not create a new index.
And it will not remove from the index. That is what the API endpoint, or the full indexer is meant for.

docs: https://github.com/rapidez/docs/pull/89

### DIFF
--- a/src/Commands/UpdateIndexCommand.php
+++ b/src/Commands/UpdateIndexCommand.php
@@ -1,0 +1,77 @@
+<?php
+
+namespace Rapidez\Core\Commands;
+
+use Carbon\Carbon;
+use Illuminate\Console\Command;
+use Illuminate\Support\Facades\Storage;
+use Rapidez\Core\Events\IndexAfterEvent;
+use Rapidez\Core\Events\IndexBeforeEvent;
+use Rapidez\Core\Facades\Rapidez;
+use Rapidez\Core\Models\Traits\Searchable;
+
+class UpdateIndexCommand extends Command
+{
+    protected $signature = 'rapidez:index:update {--t|types= : To specify types of models to index, separated by commas} {--s|store= : To specify store IDs from Magento, separated by commas}';
+
+    protected $description = 'Attempt to update changed searchable models into ElasticSearch';
+
+    private ?Carbon $latestIndexDate = null;
+
+    public function handle()
+    {
+        if (!$this->getLatestIndexDate()) {
+            $this->error(__('No latest index date has been found yet, please run php artisan rapidez:index first.'));
+            return;
+        }
+
+        $baseSearchableModels = collect(config('rapidez.models'))
+            ->filter(fn ($class) => in_array(Searchable::class, class_uses_recursive($class)) && (new $class)->getQualifiedUpdatedAtColumn());
+
+        $types = $this->option('types')
+            ? $baseSearchableModels->filter(fn ($model) => in_array((new $model)->getIndexName(), explode(',', $this->option('types'))))
+            : $baseSearchableModels;
+
+        $stores = $this->option('store')
+            ? Rapidez::getStores(explode(',', $this->option('store')))
+            : Rapidez::getStores();
+
+        IndexBeforeEvent::dispatch($this);
+
+        foreach ($stores as $store) {
+            Rapidez::setStore($store);
+
+            $this->line('Store: ' . $store['name']);
+
+            foreach ($types as $type => $model) {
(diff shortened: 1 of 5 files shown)


src/Commands/UpdateIndexCommand.php
```diff
@@ -0,0 +1,78 @@
...
+        if ($this->latestIndexDate) {
+            return $this->latestIndexDate;
+        }
+
+        if (! Storage::disk('local')->exists('/.last-index')) {
+            return null;
+        }
+
+        return $this->latestIndexDate ??= Carbon::parse(Storage::disk('local')->get('/.last-index'));
+    }
```
Jade-GG, line 77: Do you think it would be useful to make this function accessible elsewhere? e.g. putting it into the Rapidez facade? Could potentially be useful, for example, as a debugging tool---to quickly pop into a `php artisan tinker` and run it.

src/Commands/UpdateIndexCommand.php
```diff
@@ -0,0 +1,78 @@
...
+
+use Carbon\Carbon;
+use Illuminate\Console\Command;
+use Illuminate\Support\Facades\Storage;
+use Rapidez\Core\Events\IndexAfterEvent;
+use Rapidez\Core\Events\IndexBeforeEvent;
+use Rapidez\Core\Facades\Rapidez;
+use Rapidez\Core\Models\Traits\Searchable;
+
+class UpdateIndexCommand extends Command
```
Jade-GG, line 13: I can imagine cases where this command takes longer than a minute to run, causing overlaps which can definitely be problematic.

We should make sure it can't overlap, either within the recommended scheduling usage (`->withoutOverlapping()`) or maybe within the command itself.

src/Listeners/UpdateLatestIndexDate.php
```diff
@@ -0,0 +1,26 @@
...
+
+class UpdateLatestIndexDate
+{
+    public function handle()
+    {
+        return Storage::disk('local')->put(
+            '/.last-index',
+            // With this we're just making sure the comparison
+            // is done within the same timezone in MySQL.
+            DB::selectOne('SELECT NOW() AS `current_time`')->current_time
```
Jade-GG, line 18: Because this is getting set specifically after indexing, if a product gets updated exactly during the command runtime, then it won't be caught. Would probably be an uncommon occurrence, but this seems fixable.

Jade-GG (changes requested):

src/Listeners/UpdateLatestIndexDate.php
Jade-GG: Also, quick optimization:

```suggestion
            DB::scalar('SELECT NOW()')
```

src/Commands/UpdateIndexCommand.php
indykoning: Doing that would probably support misuse of the value.
It's value must only be used in the indexer, the indexers update the value so if any other command depends on the datetime stored in that file it will cause unexpected bugs.

For debugging purposes the file will be available in `storage/app/private`

src/Commands/UpdateIndexCommand.php
indykoning: I've updated the doc 👍 
It shouldn't cause issues since it will just override the already changed product information, but it wouldn't make sense to let that happen indeed.

src/Listeners/UpdateLatestIndexDate.php
indykoning: We wouldn't want to set the time before we've actually finished indexing in case indexing fails which is why we listen to the `IndexAfterEvent` instead of the `IndexBeforeEvent`

Jade-GG: I guess ideal would be to get the time when the indexing starts, but only write it when the indexing succeeds. Not sure if that can be done cleanly though.

Jade-GG (approved):

royduin (approved):

PR #833, opened 2025-05-01
Listen to the autocomplete config for size

indykoning:


Jade-GG (approved):
Can you update the newly created 4.x docs to reflect this change? https://github.com/rapidez/docs/blob/master/src/4.x/extending.md#autocomplete

PR #835, opened 2025-05-02
[Instantsearch] Allow setting of mappings & settings per index for individual models

Jade-GG:
Removes the need to overwrite the `elasticsearch.php` config file just to change these.

### DIFF
--- a/src/Models/Traits/Product/Searchable.php
+++ b/src/Models/Traits/Product/Searchable.php
@@ -125,7 +125,7 @@ trait Searchable
      */
     public static function getIndexMappings(): ?array
     {
-        return Eventy::filter('index.' . (new static)->getIndexName() . '.mapping', [
+        return Eventy::filter('index.' . static::indexName() . '.mapping', [
             'properties' => [
                 'price' => [
                     'type' => 'double',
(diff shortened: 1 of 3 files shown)


src/Models/Traits/Product/Searchable.php
```diff
@@ -119,4 +119,27 @@ public function withCategories(array $data): array
 
         return $data;
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public static function getIndexMappings(): ?array
+    {
+        return [
```
royduin, line 128: Should we merge this with the mapping Eventy filter?
- https://docs.rapidez.io/3.x/package-development.html#eventy-filters
- https://github.com/search?q=org%3Arapidez+index.product.mapping+language%3APHP&type=code&l=PHP

Jade-GG: Yes, added (and also for the settings).

src/Models/Traits/Product/Searchable.php
```diff
@@ -125,7 +125,7 @@ public function withCategories(array $data): array
      */
     public static function getIndexMappings(): ?array
     {
-        return [
+        return Eventy::filter('index.' . (new static)->getIndexName() . '.mapping', [
```
royduin: Maybe doing `parent:: getIndexMappings()` is cleaner?

PR #836, opened 2025-05-02
[3.x] Merge the frontend config for all translations within a rapidez namespace

BobWez98:
This code will get every translation that is defined in a 'rapidez-*' namepsace within a frontend.php file. This will allow packages to add frontend translations with ease. 

Docs; incoming

### DIFF
--- a/src/Actions/CollectFrontendTranslations.php
+++ b/src/Actions/CollectFrontendTranslations.php
@@ -14,34 +14,32 @@ class CollectFrontendTranslations
         $translator = app(Translator::class);
         $loader = $translator->getLoader();
 
-        $allTranslations = [];;
-
         $allTranslations = collect($loader->namespaces())
             ->filter(function (string $path, string $namespace): bool {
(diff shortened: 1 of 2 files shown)


src/Actions/CollectFrontendTranslations.php
```diff
@@ -0,0 +1,46 @@
...
+{
+    public static function collect(): array
+    {
+        /** @var Translator $translator */
+        $translator = app(Translator::class);
+        $loader = $translator->getLoader();
+
+        $allTranslations = collect($loader->namespaces())
+            ->filter(function (string $path, string $namespace): bool {
+                return str($namespace)->startsWith('rapidez');
```
Jade-GG, line 19: Is this not too restrictive? This would only allow official Rapidez packages to add frontend translations, right? What if a different vendor wants to make a package for Rapidez that includes frontend translations?

royduin:
Closing, reference: RAP-729

PR #839, opened 2025-05-06
[Instantsearch] Fix price slider

Jade-GG:
Also adds the currency denominator to the price slider display values, through a slot that I've added to the range slider component (using a slightly hacky Vue trick to have one consistent `value` variable on both ends of the slider)

### DIFF
--- a/resources/js/filters.js
+++ b/resources/js/filters.js
@@ -8,10 +8,11 @@ window.truncate = function (value, limit) {
 
 Vue.filter('truncate', window.truncate)
 
-window.price = function (value) {
+window.price = function (value, extra = {}) {
     return new Intl.NumberFormat(config.locale.replace('_', '-'), {
         style: 'currency',
         currency: config.currency,
+        ...extra,
     }).format(value)
 }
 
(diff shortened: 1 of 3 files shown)


indykoning:
For _most_ cases it should be possible to use the number formatter to determine where the currency should be placed
`new Intl.NumberFormat('de-de', { style: 'currency', currency: 'EUR' }).formatToParts(1)`
image

In the edge case where the currency is not before of after the numbers we could ignore the currency

resources/js/components/Elements/CurrencyDisplay.vue
royduin, line 1: Why did you choose for a Vue component instead of a helper somewhere? Maybe just add those in the `filters.js` so we don't have to wrap a Vue component around it when we need those methods?

Jade-GG: Doesn't really make sense for me to have these as filters because they're not. I had originally put it in-line with `v-bind:set` but that was not great either.

Could put this as a computed object perhaps? `currencyDisplay` which then returns these 2 values?

royduin: Maybe an additional `helpers.js`?

PR #840, opened 2025-05-06
[Instantsearch] Fix reset filters always showing & category filter jumping

Jade-GG:
Category filter jumping was caused by it sorting in such a way that the selected category was shown first if there were multiple categories with the same amount of items. I've changed it to instead always sort alphabetically for categories with the same amount of items.


royduin (approved):

indykoning (approved):
Looks good to me!

PR #841, opened 2025-05-06
[2.x] Use queryParams function instead of as variable

Jade-GG:
Because this is a method, and not a computed value (I'd assume because `window.location.search` is not a reactive variable), it's a function to call like this.


indykoning:
See: https://github.com/rapidez/core/pull/811
It's a computed value since https://github.com/rapidez/core/releases/tag/2.23.0

PR #842, opened 2025-05-06
 Password strength tile

Tjeerd-JW:
Password strength tile with a progress bar and a list of requirements it gets from Magento.


resources/views/components/input/password/password-strength.blade.php
```diff
@@ -0,0 +1,42 @@
...
+                <div class="flex size-4 shrink-0 items-center justify-center rounded-full bg-red-500 text-white">
+                    <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16" fill="currentColor" class="size-4 text-red-500">
+                        <path fill-rule="evenodd" d="M8 15A7 7 0 1 0 8 1a7 7 0 0 0 0 14Zm2.78-4.22a.75.75 0 0 1-1.06 0L8 9.06l-1.72 1.72a.75.75 0 1 1-1.06-1.06L6.94 8 5.22 6.28a.75.75 0 0 1 1.06-1.06L8 6.94l1.72-1.72a.75.75 0 1 1 1.06 1.06L9.06 8l1.72 1.72a.75.75 0 0 1 0 1.06Z" clip-rule="evenodd"/>
+                    </svg>
+                </div>
+                <p class="text-sm text-red-500"> @{{ error }}</p>
+            </div>
+            <div v-for="strength in passwordStrength.strengths" class="my-1 flex items-center gap-x-2">
+                <div class="flex size-4 shrink-0 items-center justify-center rounded-full bg-green-700 text-white">
+                    <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16" fill="currentColor" class="size-4 text-green-700">
```
royduin, line 34: We do have `blade-ui-kit/blade-heroicons` for icons, can you get one from there?

resources/views/components/input/password/password-strength.blade.php
```diff
@@ -0,0 +1,42 @@
+@slots(['label'])
+
+<div {{ $attributes->only('class')->class('') }}>
```
royduin, line 3: Why do we need this? We can always just wrap our component within a div with the classes we need?

resources/js/components/PasswordStrength.vue
```diff
@@ -0,0 +1,68 @@
...
+            let satisfiedClasses = 4 - errors.length
+            if (satisfiedClasses >= this.minClasses) {
+                errors = []
+            }
+            if (this.password.length < this.minLength) {
+                errors.push(window.config.translations.password.characters.replace(':minLength', this.minLength))
+            }
+            return errors
+        },
+        strengths: function () {
```
royduin, line 47: It feels like we're doing every check twice now? Within errors, and within strengths.

royduin (changes requested):
Currently it's not actually implemented anywhere? All account parts are within https://github.com/rapidez/account, maybe move everything there with a simple prop on the Blade component to enable/disable this functionality?

resources/views/components/input/password/password-strength.blade.php
```diff
@@ -0,0 +1,42 @@
...
+                    class="h-2.5 rounded-full transition-all duration-300"
+                ></div>
+            </div>
+            <div v-for="error in passwordStrength.errors" class="my-1 flex items-center gap-x-2">
+                <div class="flex size-4 shrink-0 items-center justify-center rounded-full bg-red-500 text-white">
+                    <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16" fill="currentColor" class="size-4 text-red-500">
+                        <path fill-rule="evenodd" d="M8 15A7 7 0 1 0 8 1a7 7 0 0 0 0 14Zm2.78-4.22a.75.75 0 0 1-1.06 0L8 9.06l-1.72 1.72a.75.75 0 1 1-1.06-1.06L6.94 8 5.22 6.28a.75.75 0 0 1 1.06-1.06L8 6.94l1.72-1.72a.75.75 0 1 1 1.06 1.06L9.06 8l1.72 1.72a.75.75 0 0 1 0 1.06Z" clip-rule="evenodd"/>
+                    </svg>
+                </div>
+                <p class="text-sm text-red-500"> @{{ error }}</p>
```
indykoning, line 30: Right now if this was used within a form it would not prevent form submission.
A small trick we could use is adding a hidden input with custom validity:
```suggestion
                <p class="text-sm text-red-500 relative">@{{ error }}  <input type="checkbox" oninvalid="this.setCustomValidity(error)" required class="absolute inset-0 opacity-0 pointer-events-none"></p>
```

This would result in a browser message when attempting to submit 
image

instead of getting the popup message back from the server after it has gotten submitted.

PR #844, opened 2025-05-09
Input type number remove Firefox arrows

Roene-JustBetter:
Removed arrows in type number on Firefox
Scherm­afbeelding 2025-05-09 om 15 59 11

### DIFF
--- a/resources/css/app.css
+++ b/resources/css/app.css
@@ -29,12 +29,12 @@ listing {
 }
 
 /* Remove default browser arrows for input type="number" */
-input[type='number']::-webkit-inner-spin-button,
-input[type='number']::-webkit-outer-spin-button {
+.arrows-hidden[type='number']::-webkit-inner-spin-button,
+.arrows-hidden[type='number']::-webkit-outer-spin-button {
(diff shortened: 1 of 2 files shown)


resources/css/app.css
```diff
@@ -33,3 +33,8 @@ input[type='number']::-webkit-inner-spin-button,
 input[type='number']::-webkit-outer-spin-button {
     @apply appearance-none m-0;
 }
+
+/* Firefox */
+input[type='number'] {
+    -moz-appearance: textfield;
+}
```
indykoning, line 40: This matches the format above here so it's fine.
But wouldn't we want the up/down arrows to only get removed for the quantity component?
https://github.com/rapidez/core/blob/master/resources/views/components/quantity.blade.php

Where we replace these arrows with our own.

Maybe we could target a class with e.g. `.arrows-hidden` so we can explicitly request that behavior.

Roene-JustBetter: Good idea, I have added the class `arrows-hidden` so if you add a number input it will keep original state.

PR #849, opened 2025-05-14
Added search history in autocomplete

indykoning:
This PR adds the search history in your autocomplete.
This can be controlled and (re)moved using the frontend.php

It will listen to the global size unless overwritten in the config, it will filter the history using the current search


resources/js/components/Search/Autocomplete.vue
```diff
@@ -52,6 +53,13 @@ export default {
         })
     },
 
+    computed: {
+        searchHistory() {
+            return Object.entries(searchHistory.value).sort((a, b) => {
+                return Date.parse(b[1].lastSearched) - Date.parse(a[1].lastSearched)
+            })
```
Jade-GG, line 60: Should we sort this here? Or maybe in the useSearchHistory getter itself?

PR #850, opened 2025-05-14
Reorder turbolinks startup & event to ensure teardown on first page

indykoning:
Due to the turbolinks.js file loading later than Vue sometimes, the teardown wouldn't happen on the first page load.

This fixes that issue


royduin:
This should also go into 3.x

indykoning:
Closing for #851 which will be merged into master and then instantsearch

PR #853, opened 2025-05-15
Refactored filter accordions into detail summary components

jordythevulder:
I've added a new component: the "detail summary". This replaces the input peer checked component. Because the detail summary has nothing to do with the filters, i've seperated this within the /components/details, then made a seperate .filter variant of the detail summary.

Usage:
image


royduin (approved):
- The details/summary component is widely supported
- The idea is to move the details component later to https://github.com/rapidez/blade-components which will replace the accordion component, see: RAP-1303
- The custom CSS can we removed when we've updated to Tailwind 4
- Only the `interpolate-size` stays behind as it's required for the animation within the component, fallback for unsupported browser is no animation
- `<x-rapidez::details.filter>` is a variant of the `<x-rapidez::details>` for the filters

@JimmyHoenderdaal @Roene-JustBetter @indykoning please review

Roene-JustBetter (approved):

Roene-JustBetter:
> * The details/summary component is widely supported
> * The idea is to move the details component later to https://github.com/rapidez/blade-components which will replace the accordion component, see: RAP-1303
> * The custom CSS can we removed when we've updated to Tailwind 4
> * Only the `interpolate-size` stays behind as it's required for the animation within the component, fallback for unsupported browser is no animation
> * `<x-rapidez::details.filter>` is a variant of the `<x-rapidez::details>` for the filters
> 
> @JimmyHoenderdaal @Roene-JustBetter @indykoning please review

Looking good 🚀

JimmyHoenderdaal (approved):

indykoning (approved):

PR #855, opened 2025-05-16
Added basic view transition

indykoning:
This PR adds a simple view transition for a product image to the next page, more to serve as an example to create more.
See the result:

https://github.com/user-attachments/assets/dbb89f15-f40b-434f-bda2-16027177c192


royduin:
@Roene-JustBetter and @jordythevulder, what do you think? Do we need this on other elements?

PR #860, opened 2025-05-19
Implement search suggestions in autocomplete

indykoning:
This PR provides the search suggestions during autocomplete.
As opposed to the no products search suggestion it does get filtered.


resources/views/layouts/partials/header/autocomplete/results.blade.php
```diff
@@ -1,6 +1,6 @@
 @foreach (config('rapidez.frontend.autocomplete.additionals') as $key => $fields)
     @includeIf('rapidez::layouts.partials.header.autocomplete.' . $key)
 @endforeach
-@include('rapidez::layouts.partials.header.autocomplete.products')
+@includeWhen(config('rapidez.frontend.autocomplete.additionals.products') === null,'rapidez::layouts.partials.header.autocomplete.products')
```
indykoning, line 4: This way you can optionally override the positioning and config of the product listing with the autocomplete additionals

PR #862, opened 2025-05-20
Use Cache::memo instead of multicache

indykoning:
**breaking**
Support for Laravel 11 has been dropped as it doesn't have memo

This PR removes the rapidez/laravel-multi-cache package and instead uses the new [Cache::memo()](https://laravel.com/docs/12.x/cache#cache-memoization)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,10 +19,10 @@
         "php": "^8.2",
         "blade-ui-kit/blade-heroicons": "^2.6",
         "http-interop/http-factory-guzzle": "^1.2",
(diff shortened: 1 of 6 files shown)


royduin:
Do we want to drop Laravel 11 now? It's still supported: https://endoflife.date/laravel

indykoning:
> Do we want to drop Laravel 11 now? It's still supported: [endoflife.date/laravel](https://endoflife.date/laravel)

That's a good question, it might be a bit early to drop support already.
This PR does have it prepared, but we might want to change it to draft for now

PR #866, opened 2025-05-21
Improved the look and feel

jordythevulder:
Tweaked instantsearch and category overview page including the header, breadcrumbs and filters

image


resources/views/listing/partials/toolbar.blade.php
```diff
@@ -5,7 +5,7 @@
     <label class="flex items-center gap-x-1.5">
         @include('rapidez::listing.partials.toolbar.pages')
     </label>
-    <div class="max-md:hidden">
+    <div>
```
royduin, line 8: The pages does have a label wrapped around it, the sorting doesn't need that?

Roene-JustBetter: No `<label>` around here is not needed here for both items.

Roene-JustBetter (changes requested):

indykoning (approved):

PR #868, opened 2025-05-22
Frontend - Fixes / Finetuning InstantSearch

Roene-JustBetter:
Scherm­afbeelding 2025-05-22 om 11 24 48
Scherm­afbeelding 2025-05-22 om 11 24 57
Scherm­afbeelding 2025-05-22 om 11 25 11
Scherm­afbeelding 2025-05-22 om 11 25 28
Scherm­afbeelding 2025-05-22 om 11 27 02
Scherm­afbeelding 2025-05-22 om 11 27 12
Scherm­afbeelding 2025-05-22 om 11 46 10

### DIFF
--- a/resources/views/listing/partials/item.blade.php
+++ b/resources/views/listing/partials/item.blade.php
@@ -33,6 +33,14 @@
                     @if (App::providerIsLoaded('Rapidez\Reviews\ReviewsServiceProvider'))
                         <x-dynamic-component component="rapidez-reviews::stars" v-if="item.reviews_count" count="item.reviews_count" score="item.reviews_score" />
                     @endif
+                    <div class="flex items-center gap-x-2 mt-1">
+                        <div class="font-semibold text-lg">
+                            @{{ (addToCart.simpleProduct.special_price || addToCart.simpleProduct.price) | price }}
+                        </div>
+                        <div class="line-through text-sm" v-if="addToCart.simpleProduct.special_price">
+                            @{{ addToCart.simpleProduct.price | price }}
+                        </div>
+                    </div>
                 </div>
             </a>
             @include('rapidez::listing.partials.item.addtocart')
--- a/resources/views/listing/partials/item.blade.php
+++ b/resources/views/listing/partials/item.blade.php
@@ -20,16 +20,6 @@
                 <x-rapidez::no-image v-else class="mb-3 h-48 rounded-t" />
                 <div>
                     <x-rapidez::highlight attribute="name" class="text-base font-medium hover:underline decoration-2 "/>
-                    @if (!Rapidez::config('catalog/frontend/show_swatches_in_product_list'))
-                        <div class="flex items-center space-x-2">
-                            <div class="font-semibold">
-                                @{{ (item.special_price || item.price) | price }}
-                            </div>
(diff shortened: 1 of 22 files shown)


resources/views/listing/partials/item.blade.php
```diff
@@ -33,6 +33,14 @@ class="mb-3 h-48 w-full rounded-t object-contain"
                     @if (App::providerIsLoaded('Rapidez\Reviews\ReviewsServiceProvider'))
                         <x-dynamic-component component="rapidez-reviews::stars" v-if="item.reviews_count" count="item.reviews_count" score="item.reviews_score" />
                     @endif
+                    <div class="flex items-center gap-x-2 mt-1">
+                        <div class="font-semibold text-lg">
+                            @{{ (addToCart.simpleProduct.special_price || addToCart.simpleProduct.price) | price }}
```
royduin, line 28: In this file we currently do have the pricing twice, why? Please double check with `show_swatches_in_product_list` on and off. Maybe some legacy? Seems like it's related to this: https://github.com/rapidez/core/pull/698 from @Jade-GG, old situation:
```
@includeWhen(Rapidez::config('catalog/frontend/show_swatches_in_product_list', 1), 'rapidez::listing.partials.item.addtocart')
```
And current:
```
@include('rapidez::listing.partials.item.addtocart')
```
Which comes from a long time ago: https://github.com/rapidez/core/commit/33af18bcab288db0bb77323a59dbb8a817896277 as:

> When there are swatches an add to cart button won't work as you need to select swatches first

But the price was within the addtocart template so with that the price also disappeared, that's why the price was conditionally visible based on `show_swatches_in_product_list`.

resources/views/listing/partials/toolbar/stats.blade.php
```diff
@@ -1,11 +1,8 @@
...
     <template v-slot="{ nbPages, nbHits, page, processingTimeMS }">
         @{{ nbHits }} @lang('products')
-        <template v-if="nbPages > 1">
-            (@lang('page'): @{{ page + 1 }} / @{{ nbPages }})
-        </template>
-        <template v-if="window.debug">
-            - @{{ processingTimeMS }}ms
-        </template>
+    </template>
+    <template v-if="window.debug">
```
royduin: Does this work? As it's outside the `v-slot` where the `processingTimeMS` data is coming from.

royduin (changes requested):

resources/views/listing/partials/item.blade.php
```diff
@@ -20,19 +20,17 @@ class="mb-3 h-48 w-full rounded-t object-contain"
                 <x-rapidez::no-image v-else class="mb-3 h-48 rounded-t" />
                 <div>
                     <x-rapidez::highlight attribute="name" class="text-base font-medium hover:underline decoration-2 "/>
-                    @if (!Rapidez::config('catalog/frontend/show_swatches_in_product_list'))
-                        <div class="flex items-center space-x-2">
-                            <div class="font-semibold">
-                                @{{ (item.special_price || item.price) | price }}
```
royduin, line 26: See my other comment, just to make sure: the other price is the right one? We don't need this one anymore?

PR #871, opened 2025-05-22
[Instantsearch] Select options on product page from url parameters and set those parameters in the listing

Jade-GG:
This adds URL parameters to the items in the listing so that those parameters can be passed on to the product pages.

For this, I've done a slight refactor for the `superRefinements` functionality because it had a bunch of similarities. I also made sure that this only happens to an actual product, on product pages, by adding a `url-params` parameter.

### DIFF
--- a/resources/js/components/Product/AddToCart.vue
+++ b/resources/js/components/Product/AddToCart.vue
@@ -42,6 +42,10 @@ export default {
         callback: {
             type: Function,
         },
+        urlParams: {
+            type: Boolean,
+            default: false,
+        }
     },
 
     data: () => ({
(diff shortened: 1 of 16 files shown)


resources/js/components/Product/AddToCart.vue
```diff
@@ -350,6 +381,19 @@ export default {
...
+            let namedOptions = Object.fromEntries(
+                Object.entries(this.options).map(([key, value]) => [this.product.super_attributes[key]?.code, value]),
+            )
+
+            let params = new URLSearchParams(namedOptions)
+            if (params.size) {
+                return this.product.url + '?' + params
+            } else {
+                return this.product.url
+            }
```
royduin, line 394:
```suggestion
            if (params.size) {
                return this.product.url + '?' + params
            }

            return this.product.url
```

royduin:
Will https://github.com/rapidez/core/pull/855 play nicely? So if a color is set in the listing and you click the selected color is directly visible? No short flash of another color or something?

indykoning (approved):

PR #873, opened 2025-05-22
[Instantsearch] Refactor index settings & index mapping functions to allow for arbitrary functional traits

Jade-GG:
I've gone for a fairly wild approach here, would like to hear your thoughts on this.

With this PR, we can make any trait we want with a `indexSettingsXyz` and `indexMappingXyz` function that apply any settings and mapping we might want for said trait. Have a look at the `UsesSynonyms` trait here. Or, just add those functions to your model without needing a trait, as I've done for the mapping on the product model.

A limitation of this implementation is that we currently cannot guarantee the order that these functions might be executed in. This means you can't (consistently) overwrite mapping values or settings set by other functions. Sorting the methods list could work for this, but then people would just make functions named `indexMappingZZZ` to make sure it goes last.

If we want a more comprehensive solution to this that avoids this issue we could stick to something like the fallback routes (essentially just the Eventy filters), where you can then define all your extra indexing functionalities in the service provider. However I'm not sure if I necessarily like that approach. Using traits for this seems cleaner and more reusable to me.

### DIFF
--- a/src/Commands/IndexCommand.php
+++ b/src/Commands/IndexCommand.php
@@ -40,7 +40,7 @@ class IndexCommand extends Command
             foreach ($types as $model) {
                 $searchableAs = (new $model)->searchableAs();
 
-                $indexMappings = Eventy::filter('index.' . $model::getIndexName() . '.mapping', $model::getIndexMappings());
+                $indexMappings = Eventy::filter('index.' . $model::getIndexName() . '.mapping', $model::getIndexMapping());
                 if ($indexMappings) {
                     config()->set('elasticsearch.indices.mappings.' . $searchableAs, $indexMappings);
                 }
--- a/src/Models/Product.php
+++ b/src/Models/Product.php
@@ -25,6 +25,7 @@ use Rapidez\Core\Models\Traits\Product\CastMultiselectAttributes;
 use Rapidez\Core\Models\Traits\Product\CastSuperAttributes;
 use Rapidez\Core\Models\Traits\Product\Searchable;
 use Rapidez\Core\Models\Traits\Product\SelectAttributeScopes;
+use Rapidez\Core\Models\Traits\UsesSynonyms;
 use TorMorten\Eventy\Facades\Eventy;
 
 class Product extends Model
@@ -34,6 +35,7 @@ class Product extends Model
     use HasAlternatesThroughRewrites;
     use Searchable;
     use SelectAttributeScopes;
+    use UsesSynonyms;
 
     public const VISIBILITY_NOT_VISIBLE = 1;
     public const VISIBILITY_IN_CATALOG = 2;
@@ -261,5 +263,12 @@ class Product extends Model
     public static function exist($productId): bool
     {
         return self::withoutGlobalScopes()->where('entity_id', $productId)->exists();
+    }
+
+    public static function getSynonymFields(): array
+    {
+        return [
+            'name'
+        ];
     }
 }
--- a/src/Models/Product.php
+++ b/src/Models/Product.php
@@ -268,7 +268,7 @@ class Product extends Model
     public static function getSynonymFields(): array
     {
         return [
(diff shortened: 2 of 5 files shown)


royduin:
I think it's best to stick with the Eventy filters for backwards compatibility in packages and projects for now. Those can be automatically applied from a service provider ([see Amasty Label](https://github.com/rapidez/amasty-label/blob/8118315c00bb2b6a22add9ce06499f91ce27446f/src/AmastyLabelServiceProvider.php#L22)). With this approach you need to overwrite your model to apply the traits/methods right? Also the "free naming" could be confusing.

Jade-GG:
> I think it's best to stick with the Eventy filters for backwards compatibility in packages and projects for now. Those can be automatically applied from a service provider ([see Amasty Label](https://github.com/rapidez/amasty-label/blob/8118315c00bb2b6a22add9ce06499f91ce27446f/src/AmastyLabelServiceProvider.php#L22)). With this approach you need to overwrite your model to apply the traits/methods right? Also the "free naming" could be confusing.

I'm just not a huge fan of putting all functionality within the serviceprovider, where it really should just be booting things, i.e. I definitely wouldn't want to have to set the whole list of index settings and mappings from the serviceprovider.

However I could definitely mess around with something like this in the serviceprovider:

```php
Eventy::filter('index.product.settings', fn($settings) => array_merge($settings ?: [], [[WithSynonyms::class, ['name']]]);

// Maybe even make a helper function like this:
Rapidez::addFilter('index.product.settings', [WithSynonyms::class, ['name']])
```

It's tricky though because this would need to set the mapping and settings simultaneously.

royduin:
Agree, but maybe now isn't the moment to refactor this and break backwards compatibility. A service provider is to boot, but also register things and if that needs to be dynamic, you can always extract things elsewhere. A helper could be nice but with the Eventy filter you're also able to override values and change priority.

indykoning:
That could be fixed by using Laravel's events and merging the results, like done here https://github.com/rapidez/core/blob/master/src/Commands/ValidateCommand.php#L18

As listeners (so long as they're not queued) support classes, and inline functions while returning data.

I do agree that it might not be the right moment to roll that out though.

Jade-GG:
I've simplified it quite a bit now, keeping the previous working as it always had been, but allowing you to add classes into the mapping/settings that will get executed with any extra parameters, for example:

```php
// Simple without parameters
Eventy::addFilter('index.products.settings', fn($filters) => array_merge($filters ?: [], [WithSynonyms::class]));

// With parameters because those are needed for the mapping
Eventy::addFilter('index.products.mapping', fn($filters) => array_merge($filters ?: [], [
    [WithSynonyms::class, 'fields' => $productSynonymFields]
]));
```

src/RapidezServiceProvider.php
```diff
@@ -291,6 +294,19 @@ function ($filename) use ($manifest) {
...
+        $productSynonymFields = Eventy::filter('index.products.synonym-fields', ['name']);
+        Eventy::addFilter('index.products.settings', fn ($filters) => array_merge($filters ?: [], [WithSynonyms::class]));
+        Eventy::addFilter('index.products.mapping', fn ($filters) => array_merge($filters ?: [], [[WithSynonyms::class, 'fields' => $productSynonymFields]]));
+
+        $categorySynonymFields = Eventy::filter('index.categories.synonym-fields', ['name']);
+        Eventy::addFilter('index.categories.settings', fn ($filters) => array_merge($filters ?: [], [WithSynonyms::class]));
+        Eventy::addFilter('index.categories.mapping', fn ($filters) => array_merge($filters ?: [], [[WithSynonyms::class, 'fields' => $categorySynonymFields]]));
+
+        return $this;
+    }
```
royduin, line 308: Do we also want to use these events from the core? Maybe cleaner to just have them and able to use the filters to override/extend it?

PR #874, opened 2025-05-23
Add loading indicator in autocomplete slower connections

Roene-JustBetter:
In some cases when you have slow internet connection and you type something in the autocomplete the "no results" view show for a short time before it have results. 

By adding the `<ais-state-results v-slot="{ status }">` in this template we can check the status (https://www.algolia.com/doc/api-reference/widgets/state-results/vue/#status). When we have a slow connection we will see now the loader for a short time instead of the "no results".

### DIFF
--- a/resources/views/layouts/partials/header/autocomplete/no-results.blade.php
+++ b/resources/views/layouts/partials/header/autocomplete/no-results.blade.php
@@ -1,28 +1,33 @@
-<ais-hits>
-    <template v-slot="{ items }">
-        <div v-if="items.length === 0 && currentRefinement !== ''" class="p-5">
-            <div class="font-bold text text-lg break-all">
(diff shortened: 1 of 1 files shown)


indykoning (approved):

Roene-JustBetter:
> Could you make the loader visually a bit cleaner? Maybe a smaller loading icon with the text "searching..." next to it?

Added translatable string & made the icon little bit smaller from 20px to 12px.
See below: 
https://github.com/user-attachments/assets/1e167f9e-8885-41d4-a888-3ec7bcd67ff5

jordythevulder:
i feel like something is still missing. is there another type of icon instead of arrows running in circles? and i miss the loading text like roy mentioned

jordythevulder (approved):

PR #885, opened 2025-06-05
[Instantsearch] Add data-turbo to search suggestions & history, make route + replace usage consistent

Jade-GG:
The search broke for the search suggestions and history. We also didn't yet have a `data-turbo="false"` on the search suggestions, where we potentially redirect to an external URL.

The `data-turbo="false"` ended up fixing the search breaking bug, however I don't know if this will be the best solution. For that I've made an internal issue to look at this (RAP-1377).

Finally, I changed the `v-bind:href` in the history file just to be consistent with the one in the search suggestions.


royduin:
Merging in for now so we can release, maybe @indykoning knows another solution? External urls likely won't be used

PR #891, opened 2025-06-13
Playwright tests

royduin:


royduin:
Merging as `issue_comment`:
> This event will only trigger a workflow run if the workflow file exists on the default branch.

PR #892, opened 2025-06-13
Validate shipping address before sending same as shipping

indykoning:
This PR will prevent the 
image
error when unchecking & checking same as shipping when no valid shipping address was entered.

If same as shipping is selected we check wether a shipping address has been set before sending


royduin:
Can't we validate the whole form instead of just checking the postcode?

PR #901, opened 2025-06-19
Only select _value fields for int select fields

indykoning:
This change fixes an issue where it would try to get `_value` fields for varchar dropdowns, for which magento does not generate a `_value` key

### DIFF
--- a/src/Models/Category.php
+++ b/src/Models/Category.php
@@ -100,7 +100,7 @@ class Category extends Model
         $categoryIds = explode('/', $this->path);
         $categoryIds = array_slice($categoryIds, array_search(config('rapidez.root_category_id'), $categoryIds) + 1);
 
-        return ! $categoryIds ? [] : Category::whereIn($this->getQualifiedKeyName(), $categoryIds)
+        return ! $categoryIds ? collect([]) : Category::whereIn($this->getQualifiedKeyName(), $categoryIds)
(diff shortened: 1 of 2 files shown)


src/Models/Category.php
```diff
@@ -100,7 +100,7 @@ public function getParentcategoriesAttribute()
         $categoryIds = explode('/', $this->path);
         $categoryIds = array_slice($categoryIds, array_search(config('rapidez.root_category_id'), $categoryIds) + 1);
 
-        return ! $categoryIds ? [] : Category::whereIn($this->getQualifiedKeyName(), $categoryIds)
+        return ! $categoryIds ? collect([]) : Category::whereIn($this->getQualifiedKeyName(), $categoryIds)
```
royduin, line 103: `collect([])` vs `collect()`, output is the same right?

src/Models/Scopes/Product/WithProductAttributesScope.php
```diff
@@ -36,7 +36,7 @@ public function apply(Builder $builder, Model $model)
             $attribute = (object) $attribute;
 
             if ($attribute->flat) {
-                if ($attribute->input == 'select' && ! in_array($attribute->source_model, [
+                if ($attribute->input == 'select' && $attribute->type === 'int' && ! in_array($attribute->source_model, [
```
royduin, line 39: Let's do strict comparison (`===`) everywhere, or nowhere to keep code consistent.

PR #902, opened 2025-06-19
Automatically parse types to set as mappings

indykoning:
This PR basically takes all indexable attributes, tries to determine their type and pass it to elasticsearch.

The biggest usecase for this is custom range sliders which WILL break if the type is set to a textual type.
It's also some data optimisation for Elasticsearch.

### DIFF
--- a/src/Models/Traits/Product/Searchable.php
+++ b/src/Models/Traits/Product/Searchable.php
@@ -122,54 +122,49 @@ trait Searchable
     protected static function indexMapping(): array
     {
         $attributeModel = config('rapidez.models.attribute');
-        $attributeTypeMapping = Arr::map(
-            Arr::pluck(
-                $attributeModel::getCachedWhere(function ($attribute) {
-                    if (in_array($attribute['code'], ['msrp_display_actual_price_type', 'price_view', 'shipment_type', 'status'])) {
-                        return false;
-                    }
 
-                    if (in_array($attribute['type'], ['varchar', 'text', 'gallery', 'static'])) {
-                        // Types best left up to ES to interpret the type.
-                        return false;
-                    }
-
-                    if (! empty($attribute['source_model'])) {
-                        // Due to the source model value can be mapped to any type, best to let ES interpret them.
-                        return false;
-                    }
-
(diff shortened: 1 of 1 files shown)


Jade-GG:
You seem to filter out text fields but then define them anyway in the match below. As I don't have a perfect view of how this will run---does this keep the `keyword` mapping for text fields? If not, you will no longer be able to search text fields after this.

For example, this is currently the mapping given to the name field:

```json
"name": {
  "type": "text",
  "fields": {
    "keyword": {
      "type": "keyword",
      "ignore_above": 256
    }
  },
  "analyzer": "synonym"
},
```

src/Models/Traits/Product/Searchable.php
```diff
@@ -121,8 +121,51 @@ public function withCategories(array $data): array
...
+        $attributeModel = config('rapidez.models.attribute');
+        $attributeTypeMapping = Arr::map(
+            Arr::pluck(
+                $attributeModel::getCachedWhere(function ($attribute) {
+                    if (in_array($attribute['code'], ['msrp_display_actual_price_type', 'price_view', 'shipment_type', 'status'])) {
+                        return false;
+                    }
+
+                    if (in_array($attribute['type'], ['varchar', 'text', 'gallery', 'static'])) {
+                        // Types best left up to ES to interperet the type.
```
Jade-GG, line 133:
```suggestion
                        // Types best left up to ES to interpret the type.
```

src/Models/Traits/Product/Searchable.php
```diff
@@ -121,8 +121,51 @@ public function withCategories(array $data): array
...
+                        return false;
+                    }
+
+                    if (! empty($attribute['source_model'])) {
+                        // Due to the source model value can be mapped to any type, best to let ES interperet them.
```
Jade-GG, line 138:
```suggestion
                        // Due to the source model value can be mapped to any type, best to let ES interpret them.
```

Jade-GG (changes requested):

Jade-GG:
I'm still unsure about the default match being a `text` type which may still break with the keyword thing. Can't we make this a `null` so you can easily do a `Arr::whereNotNull(...)` to filter these out at the end?

Jade-GG (approved):

PR #906, opened 2025-06-20
Get non-flat table attributes using relations instead of join

indykoning:
This is inspired from https://github.com/rapidez/core/tree/replace-flat-with-eav and can be a start of the migration.

This PR will exclude non-flat attributes from being selected with selectForProductPage. preventing the joins from taking place.
Any attributes that are used and unavailable on the PDP will be pulled from relationships instead of the joins.


indykoning:
Closing in favor of https://github.com/rapidez/core/pull/943

PR #922, opened 2025-07-11
Prevent add to cart button bouncing

Roene-JustBetter:
When we click on the add to cart button the text inside the button changes this will make the button bouncing. We can't add a width because translations could be longer or shorter inside the button.

https://github.com/user-attachments/assets/813fe839-36f3-48ff-8257-3f59ed5cfe71

To prevent the bounce on the button I removed the text inside the button when we are in the loading and finished state. I added a loading indicator on the button and `cursur-progress` when we are in the loading sate. 

https://github.com/user-attachments/assets/5b2b8691-8536-4dae-939c-e368f449670b


JimmyHoenderdaal (dismissed):

resources/components/add-to-cart-animation.css
royduin, line 1: Directly within the `app.css` is fine, we already have some utilities there and this isn't to big.

Roene-JustBetter:
> Please fix the Dusk tests 😇

Should be fixed now.

resources/css/app.css
```diff
@@ -64,3 +64,12 @@ input[type='search']::-webkit-search-results-decoration {
 .turbo-progress-bar {
     @apply bg-primary;
 }
+
+/* This is the animation for the add to card */
```
jordythevulder, line 68: typo, "Add to cart"

PR #923, opened 2025-07-11
[4.x] Use Attribute casts instead of getXxxAttribute

Jade-GG:
We still used the old `getXxxAttribute()` functionality in a few places. This PR replaces all of those with the Attribute cast.

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -10,4 +10,5 @@ parameters:
         - '#^Result of static method TorMorten\\Eventy\\Events::#'
         - "#Access to an undefined property#"
         - "#::waitUntilIdle#"
+    checkModelAppends: false
     level: 1
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -10,5 +10,4 @@ parameters:
         - '#^Result of static method TorMorten\\Eventy\\Events::#'
         - "#Access to an undefined property#"
         - "#::waitUntilIdle#"
-    checkModelAppends: false
     level: 1
--- a/src/Models/Category.php
+++ b/src/Models/Category.php
@@ -109,7 +109,7 @@ class Category extends Model
                     ->orderByRaw('FIELD(' . $this->getQualifiedKeyName() . ',' . implode(',', $categoryIds) . ')')
                     ->get();
             }
-        );
+        )->shouldCache();
     }
 
     /**
--- a/src/Models/Category.php
+++ b/src/Models/Category.php
@@ -3,6 +3,7 @@
 namespace Rapidez\Core\Models;
 
 use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\Casts\Attribute;
 use Illuminate\Database\Eloquent\Relations\HasMany;
 use Illuminate\Database\Eloquent\Relations\HasManyThrough;
 use Rapidez\Core\Models\Scopes\IsActiveScope;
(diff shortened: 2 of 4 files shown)


Jade-GG:
Actually, I'm missing a few. Give me a moment.

phpstan.neon
```diff
@@ -10,4 +10,5 @@ parameters:
         - '#^Result of static method TorMorten\\Eventy\\Events::#'
         - "#Access to an undefined property#"
         - "#::waitUntilIdle#"
+    checkModelAppends: false
```
indykoning, line 13: Is there a reason this needs to be false?
By the looks of it this option solves real bugs 
https://github.com/larastan/larastan/blob/3.x/docs/rules.md#modelappendsrule

src/Models/Category.php
```diff
@@ -95,14 +98,18 @@ public function rewrites(): HasMany
...
-    public function getParentcategoriesAttribute()
+    public function parentcategories(): Attribute
     {
-        $categoryIds = explode('/', $this->path);
-        $categoryIds = array_slice($categoryIds, array_search(config('rapidez.root_category_id'), $categoryIds) + 1);
-
-        return ! $categoryIds ? collect() : Category::whereIn($this->getQualifiedKeyName(), $categoryIds)
-            ->orderByRaw('FIELD(' . $this->getQualifiedKeyName() . ',' . implode(',', $categoryIds) . ')')
-            ->get();
+        return Attribute::make(
```
indykoning, line 103: We'll probably want to add shouldCache here
https://laravel.com/docs/master/eloquent-mutators#accessor-caching

src/Models/Product.php
```diff
@@ -164,92 +164,114 @@ public function scopeByIds(Builder $query, array $productIds): Builder
...
+            get: fn (): string => '/' . ($this->url_key ? $this->url_key . Rapidez::config('catalog/seo/product_url_suffix') : 'catalog/product/view/id/' . $this->entity_id)
+        );
     }
 
-    public function getImagesAttribute(): array
+    public function images(): Attribute
     {
-        return $this->gallery->sortBy('productImageValue.position')->pluck('value')->toArray();
+        return Attribute::make(
+            get: fn (): array => $this->gallery->sortBy('productImageValue.position')->pluck('value')->toArray()
```
indykoning, line 246: https://github.com/rapidez/core/pull/923/files#diff-5f493903be8e6475da70fe592c682dd44751839dc2a58abc18ed1fa227946210R103
Same

src/Models/Product.php
```diff
@@ -164,92 +164,114 @@ public function scopeByIds(Builder $query, array $productIds): Builder
...
+            get: fn (?string $image): ?string => $image !== 'no_selection' ? $image : null
+        );
     }
 
-    public function getSmallImageAttribute($image): ?string
+    public function smallImage(): Attribute
     {
-        return $this->getImageAttribute($image);
+        return Attribute::make(
+            get: fn (): ?string => $this->image
```
indykoning, line 260: Wouldn't we be missing the ability to pass `$image` through to the image attribute?

phpstan.neon
Jade-GG: Yes, I tried to find an alternative but right now larastan will error out on the appends after changing the url function to an attribute cast. Where the functionality is still the same and stuff like indexing still does not work properly when the `appends` is left empty.

indykoning: Ah, you mean
https://github.com/larastan/larastan/issues/1917
https://github.com/larastan/larastan/blob/2.x/docs/features.md#laravel-9-attributes

royduin:
All good and ready to merge @indykoning?

PR #924, opened 2025-07-14
[2.x] Fix login & option to allow checkout without login

Jade-GG:
Contains a small fix for when a user refreshes making the password field not show up but the email field continues to be filled. This is (as far as I'm aware) only relevant for 2.x.

---

Magento does not actually have an option to require logging in when trying to check out with an existing email address. So, it is weird that Rapidez works the other way around right now.

In principle this config value should default to `true` but I don't want to break backwards compatibility here so I've kept it as `false`.

Note that this will also have to be done in checkout theme, confira, 3.x and 4.x.

### DIFF
--- a/resources/js/components/Checkout/Login.vue
+++ b/resources/js/components/Checkout/Login.vue
@@ -52,24 +52,24 @@ export default {
                 })
             } else if (this.email) {
                 await this.checkEmailAvailability()
+
+                if (this.emailAvailable) {
+                    this.$root.guestEmail = this.email
+                    this.$root.checkout.step = this.nextStep
+                } else {
+                    this.$nextTick(function () {
+                        this.$scopedSlots.default()[0].context.$refs.password.focus()
+                    })
+                }
             } else {
                 Notify(window.config.translations.account.email, 'error')
             }
         },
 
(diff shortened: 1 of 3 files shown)


Jade-GG:
For simplicity's sake we may want to only do this for 4.x. Given that we would have to update 3 packages for 2.x, 3.x, and 4.x otherwise.

resources/js/components/Checkout/Login.vue
```diff
@@ -50,24 +52,24 @@ export default {
...
+                    })
+                }
             } else {
                 Notify(window.config.translations.account.email, 'error')
             }
         },
 
         async checkEmailAvailability() {
-            let responseData = await window.magentoAPI('post', 'customers/isEmailAvailable', {
+            this.emailAvailable = await window.magentoAPI('post', 'customers/isEmailAvailable', {
```
indykoning, line 70: With these changes it might make sense to change the function name and return the value we get from Magento's API instead of setting a value.
As we no longer do any checking here, rather returning the result of `isEmailAvailable`

Jade-GG:
Closing in favor of only adding this in 4.x

PR #942, opened 2025-07-23
Implement eav structure short of options

indykoning:
Current status:
Everything of the eav structure is implemented, gets eager loaded.
Each eav attribute table does 2 queries (regardless of product count) it will re-use models already fetched across all products.

Sorting and filtering can be done by the added scopes.

~~Currently this PR is still lacking in product option values for multiselects saved in the varchar and text table.~~

The current query log for e.g. 10 products looks like
```
[
  [
    "query" => "select * from `catalog_product_entity` limit 10",
    "bindings" => [],
    "time" => 4.78,
  ],
  [
    "query" => "select * from `catalog_product_entity_varchar` where `catalog_product_entity_varchar`.`entity_id` in (1, 2, 3, 4, 5, 6, 7, 8, 9, 10)",
    "bindings" => [],
    "time" => 1.4,
  ],
  [
    "query" => "select * from `eav_attribute` where `eav_attribute`.`attribute_id` in (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)",
    "bindings" => [
      73,
      87,
      88,
      89,
      106,
      109,
      110,
      111,
      121,
      124,
      134,
    ],
    "time" => 0.9,
  ],
  [
    "query" => "select * from `catalog_product_entity_text` where `catalog_product_entity_text`.`entity_id` in (1, 2, 3, 4, 5, 6, 7, 8, 9, 10)",
    "bindings" => [],
    "time" => 0.8,
  ],
  [
    "query" => "select * from `eav_attribute` where `eav_attribute`.`attribute_id` in (?, ?, ?, ?, ?, ?)",
    "bindings" => [
      75,
      137,
      138,
      139,
      140,
      141,
    ],
    "time" => 0.37,
  ],
  [
    "query" => "select * from `catalog_product_entity_int` where `catalog_product_entity_int`.`entity_id` in (1, 2, 3, 4, 5, 6, 7, 8, 9, 10)",
    "bindings" => [],
    "time" => 0.57,
  ],
  [
    "query" => "select * from `eav_attribute` where `eav_attribute`.`attribute_id` in (?, ?, ?, ?, ?, ?, ?, ?, ?)",
    "bindings" => [
      93,
      97,
      99,
      136,
      145,
      146,
      147,
      148,
      149,
    ],
    "time" => 0.4,
  ],
  [
    "query" => "select * from `eav_attribute_option_value` where `eav_attribute_option_value`.`option_id` in (?, ?, ?, ?, ?)",
    "bindings" => [
[...]

PR #943, opened 2025-07-24
Use EAV structure instead of flat tables

Jade-GG:
~~See #942, this is my alternative version, which includes attribute options and uses joins for attribute data & option values (but a relation for the options themselves).~~

Based on better benchmarks and cleaner relations, we have chosen to keep this approach as opposed to #942. I've taken some of the extra functionality from there and have implemented into this PR.

---

With this PR we end up having 13 queries in total for products (regardless of how many products you query):
- 1 for the product itself
- 1 for the stock status
- 1 for the super attributes
- 2 for the category info (do we need this eager loaded?)
- 5 for each type of entity value
- 3 for the option relations, as 3 of the types (int, varchar, text) could have a custom option value, as far as I'm aware.

I've also made a small `backend_model` thing to parse data that may be in a weird format but I don't really know if this is a worthwhile thing to have abstracted like this. This is currently only used for comma separated values.

This also includes #944 currently, to add support for querying based on attributes. This is necessary to filter products based on the `status` attribute.

---

Current to do list:

- Optimizations where possible
- Functionality that came from the scopes
  - [x] Super attributes
  - [x] Stock information
  - [x] Upsells
  - [x] Grouped products
  - [x] Category information
  - [x] Selecting for specific use cases, e.g. `selectForProductPage`
- Attributes
  - [x] URL attribute
  - [x] Images (all types)
  - [x] Special price
  - [x] Price (including configurable and grouped products)
  - [x] Min sale qty
  - [x] Breadcrumbs
- Relations
  - [x] Product options
  - [x] categoryProducts (is this necessary?)
  - [x] reviewSummary (is this necessary?)
- Package updates
  - [ ] rapidez/reviews
  - [ ] ...finish this list
- [x] Check the data that enters the indexer
- [x] Optimize the indexer data because it is currently unacceptably slow
- [ ] Tier pricing
- [ ] Update and fix unit tests
- [ ] Compatibility with Runway
- Frontend
  - [x] Update product page
  - [ ] Super attributes on product listing page combined with swatch filters
  - [ ] ...finish this list
- [ ] Super attributes but actually working
- [x] Retrieving children (especially when needed in the indexer) is incredibly slow
- [ ] Anything else I'm missing? Test everything and finish this list


src/Models/EavAttribute.php
Jade-GG, line 1: Might have to rename this to something else, like `AbstractAttribute` or something. Right now its name suggests that it is the `eav_attribute` table when it is actually not.

src/Models/Traits/HasAttributeOptions.php
```diff
@@ -8,18 +8,29 @@
...
+    protected static $attributeCache = [];
 
     protected function options(): Attribute
     {
-        // Sort by store_id first to always get the higher store id if there are two.
-        return Attribute::get(fn () => $this->attributeOptions
-            ->sortBy('store_id')
-            ->keyBy('option_id')
-        );
+        return Attribute::get(function() {
```
royduin: Object caching is there by default? https://laravel.com/docs/master/eloquent-mutators#value-object-caching, but there is also a caching option: https://laravel.com/docs/master/eloquent-mutators#accessor-caching

PR #946, opened 2025-07-25
[4.x] Use slot props in listings to avoid instantsearch doing full re-renders

Jade-GG:
This is kind of a workaround, as a fix for this exact issue has eluded me.

---

When you have a Vue component around the `ais-instant-search` component, with a v-slot using the `{ a, b, ... }` notation, for example like this:

```blade
<toggler v-slot="{ isOpen, toggle }">
    <div>
        <ais-instant-search>
            [...]
        </ais-instant-search>
    </div>
</toggler>
```

If *any* of the variables that exist within that slot scope change(in a non-deep way)---even just a completely unused boolean going from true to false---the whole instantsearch tree gets refreshed. Adding a key does not resolve this issue.

Something of note here as well is that when this tree gets refreshed, the classname of the instantsearch element changes from:

```html
<div class="Ais-InstantSearch">
```

to:

```html
<div class="[object Object]">
```

I've not been able to figure out how that is even possible, as it appears the functions that control the classname here can only possibly return strings (or throw errors).

---

Because it's only non-deep variable changes, this means we can stick to objects. In this case, using SlotProps variables for the slots instead. This somewhat fixes this issue. We still end up with `[object Object]` as the classname for the ais-instant-search div on the product overview pages, which you can also see happens in the rapidez demo.

### DIFF
--- a/resources/views/components/listing.blade.php
+++ b/resources/views/components/listing.blade.php
@@ -8,17 +8,17 @@
 <div class="min-h-screen">
     <listing
         {{ $attributes }}
-        v-slot="{ loaded, index, searchClient, rangeAttributes, categoryAttributes, hitsPerPage, filters, sortOptions, withFilters, withSwatches, routing, middlewares }"
+        v-slot="listingSlotProps"
         v-cloak
         v-bind:root-path='@json($rootPath)'
     >
         <div>
             <ais-instant-search
-                v-if="searchClient"
-                :search-client="searchClient"
-                :middlewares="middlewares"
-                :index-name="index"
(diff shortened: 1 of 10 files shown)


royduin:
Previously we kept this off: https://github.com/rapidez/core/pull/601, this is the only way to fix this?

Jade-GG:
> Previously we kept this off: #601, this is the only way to fix this?

I spent quite a while debugging the code trying to figure out an alternative fix and I came up empty-handed. It's possible @indykoning could find something?

PR #950, opened 2025-07-28
[4.x] Make login required an optional thing

Jade-GG:
See #924 for details.

### DIFF
--- a/config/rapidez/frontend.php
+++ b/config/rapidez/frontend.php
@@ -28,7 +28,7 @@ return [
     ],
 
     // If set to false, you will be required to log in when attempting to check out with an existing email address
-    'allow_guest_on_existing_account' => true,
+    'allow_guest_on_existing_account' => false,
 
     'autocomplete' => [
         // Attach additional indexes to the autocomplete


indykoning (approved):
Do also note the 
https://github.com/rapidez/core/blob/master/resources/js/stores/useUser.js#L94

Which would sort-of fix checkout login by not offering login during the checkout

Jade-GG:
> Do also note the https://github.com/rapidez/core/blob/master/resources/js/stores/useUser.js#L94
> 
> Which would sort-of fix checkout login by not offering login during the checkout

While this is correct, this doesn't change the underlying issue that our current implementation isn't how magento handles it. If we really want to be correct about it, I should change the default value of this config value to `true` because that is the default magento behavior.

config/rapidez/frontend.php
```diff
@@ -27,6 +27,9 @@
         // 'onestep',
     ],
 
+    // If set to false, you will be required to log in when attempting to check out with an existing email address
+    'allow_guest_on_existing_account' => true,
```
royduin, line 31:
```suggestion
    'allow_guest_on_existing_account' => false,
```

PR #951, opened 2025-07-28
Vue 3 upgrade

indykoning:
This PR fixes the basic functionality, in Vue 3 like:
- Autocomplete
- Category
- Search
- Cart
- AddToCart
- Checkout
- Success page

Still broken:
- Account center

---
https://v3-migration.vuejs.org/ 

https://v3-migration.vuejs.org/breaking-changes/v-if-v-for.html 

[Dropped support for global $emit and $on](https://v3-migration.vuejs.org/breaking-changes/events-api#event-bus), move this over to e.g.
```
window.$emit('rapidez:notification-message', message, type, params, link)
window.$on('rapidez:notification-message', function (message, type, params, link) {})
```
These are functions that utilise Js' default event listeners

Rename all slot-scope to v-slot (and move them to the component in your template instead of the first slot element)

price, truncate and url filter no longer work, replace @{{ final_price | price }} with @{{ price(final_price) }}

async components should change from `() => import('...')` to `defineAsyncComponent(() => import('...'))`

.context has been replaced by .ctx

Replace window.app.<...> e.g. cart  with window.app.config.globalProperties.<...> or if you’re within vue templates directly do <...>

When inside js functions all computed (refs) must be retrieved with .value

`this.$el` now gives an empty text, this should be replaced by `this.$el.nextSibling`

All `Vue.set` calls should be removed

Partial submit has changed.
```diff
- <fieldset partial-submit="mutate" ...
+ <fieldset partial-submit v-on:partial-submit="async () => await mutate()" ...
```


resources/views/cart/sidebar.blade.php
```diff
@@ -1,46 +1,48 @@
...
+            <dd>-@{{ window.price(discount.amount.value) }}</dd>
         </div>
     </template>
 
     <div class="border-t pt-3 mt-3 font-bold">
         <dt>@lang('Total')</dt>
-        <dd v-if="showTax">@{{ cart.prices.grand_total.value | price }}</dd>
-        <dd v-else>@{{ cart.prices.grand_total.value - cart.taxTotal | price }}</dd>
+        <dd v-if="showTax">@{{ window.price(cart.value.prices.grand_total.value) }}</dd>
+        <dd v-else>@{{ window.price(cart.value.prices.grand_total.value - cart.value.taxTotal.value) }}</dd>
```
Roene-JustBetter:
```suggestion
        <dd v-else>@{{ window.price(cart.value.prices.grand_total.value - cart.value.taxTotal) }}</dd>
```

resources/views/layouts/partials/header/account.blade.php
```diff
@@ -1,6 +1,6 @@
 <div class="mr-3">
-    <toggler v-if="$root.loggedIn" v-cloak>
-        <div slot-scope="{ toggle, close, isOpen }" v-on-click-away="close">
+    <toggler v-if="window.app.config.globalProperties.loggedIn.value" v-cloak v-slot="{ toggle, close, isOpen }">
+        <div v-on-click-away="close">
             <button data-testid="account-menu" class="flex my-1" v-on:click="toggle">
                 <x-heroicon-o-user class="size-6"/>
                 @{{ $root.user.firstname }}
```
Roene-JustBetter:
```suggestion
                @{{ user.value?.firstname }}
```

PR #856, opened 2025-05-16
Cart / Checkout - Sidebar improvements. 

Roene-JustBetter:
This makes it easier to add more data inside the sidebar in the cart & checkout. You the itemsn inside here:
`<x-rapidez::summary.row>`. Use the `<x-rapidez::summary.title>` for a title and `<x-rapidez::summary.value>` for the value this will result in:
Scherm­afbeelding 2025-05-16 om 14 32 12
Add an `<x-rapidez::summary.description>` for smaller text under the title that will result in:
Scherm­afbeelding 2025-05-16 om 14 32 57

When you want a border between data or items you can use the `<x-rapidez::summary.divider />` this will create a border between the items with correct spacing: 
Scherm­afbeelding 2025-05-16 om 14 34 12

---

Also checked the WCAG 2.1AA with this tool: https://chromewebstore.google.com/detail/axe-devtools-web-accessib/lhdoppojpmngadmnindnejefpokejbdd for the HTML structure.

----

**Sidebar before:**
Scherm­afbeelding 2025-05-16 om 11 20 42
Scherm­afbeelding 2025-05-16 om 11 20 56
Scherm­afbeelding 2025-05-16 om 11 22 00

**Sidebar after**
Scherm­afbeelding 2025-05-16 om 12 02 38
Scherm­afbeelding 2025-05-16 om 12 02 53
Scherm­afbeelding 2025-05-16 om 14 27 01
Scherm­afbeelding 2025-05-16 om 14 27 11

PR #953, opened 2025-07-30
Cart and checkout visual summary improvements

jordythevulder:
Follow up of this [PR](https://github.com/rapidez/core/pull/856/files). That PR was over engineered with too many components. A good way to implement the summary but not within the core. My proposal just adds 1 component for the <dl.

image
image

### DIFF
--- a/resources/views/cart/item.blade.php
+++ b/resources/views/cart/item.blade.php
@@ -1,6 +1,6 @@
 <table class="w-full border-b">
     <tbody class="divide-y">
-        <tr v-for="(item, index) in cart.items" class="flex-wrap max-md:flex *:p-2 md:*:p-4">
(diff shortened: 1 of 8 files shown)


jordythevulder:
> I think you should move the remove button in the cart and give it a red color so it is clear for the user that this is a remove button.

If i make it red it would be visually more prominent, which i don't want. I think "Remove" indicates what it does.

Roene-JustBetter:
> If i make it red it would be visually more prominent, which i don't want. I think "Remove" indicates what it does.
Agree with it but remove still looks like a product option right now. 

You also have merge conflicts.

Roene-JustBetter (approved):

Roene-JustBetter (dismissed):

PR #955, opened 2025-07-31
Changed file naming from lower dashes to regular dashes

jordythevulder:
I've changed the lower dashes _ into > middle dashes -. I've done some research within other repositories such as Laravel, Statamic and Vue. And they also use middle regular dashes within file naming. To create consistency i've checked every functionality to make sure they still work.


Jade-GG (approved):

royduin:
Please also check other repos, for example checkout theme, confira and statamic

PR #959, opened 2025-08-04
[4.x] Allow passwordless on the form itself

Jade-GG:
Unfortunately I missed a spot with the "allow guest on existing account". This form will return a `false` right now when no password is provided, which means you can't continue in the checkout but it will not give you any error on the frontend.


Jade-GG:
Note that this will have to be fixed in checkout-theme

royduin:
No PR for checkout theme yet?

royduin:
Follow up on: https://github.com/rapidez/core/pull/950

PR #960, opened 2025-08-04
[3.x] Allow passwordless on the form itself

Jade-GG:
#959 but then for 3.x.


Jade-GG:
Config only exists on 4.x

PR #961, opened 2025-08-04
[4.x] Get value directly instead of computed to avoid cyclic value

Jade-GG:
When trying to stringify the cart value (to save into localstorage), these values caused an error because they were not serializable.

By directly setting the value here instead, we avoid this because it's no longer a computed object.

Note that this is currently a **breaking change** because `taxTotal.value` has to become `taxTotal` everywhere. We can work around this by setting `taxTotal` to `{ value: xxx }` instead if that's preferable.


Jade-GG:
> See the updated screenshots, the total is now "NaN" 😇
> 
> Or the report: https://rapidez.github.io/playwright-reports/rapidez/core/961/merge/16723699539/1/ from the action: https://github.com/rapidez/core/actions/runs/16723699539

Weird, I had this working on my end. Will look into it.

PR #965, opened 2025-08-06
[4.x] Add watch ignore and ignore 'same_as_shipping' to fix checkout bug

Jade-GG:
This fixes a bug where changing anything in your shipping address would make the billing address reappear in your checkout.


royduin:
Related to: https://github.com/rapidez/core/pull/815

PR #969, opened 2025-08-08
Added loader for buttons

Roene-JustBetter:
This will add a loader on a button when we are waiting for request for example inside the checkout. When you add `loader` on a button it will get a loader:
Scherm­afbeelding 2025-08-08 om 10 51 02

### DIFF
--- a/resources/views/components/button/tag.blade.php
+++ b/resources/views/components/button/tag.blade.php
@@ -1,0 +1,27 @@
+@props(['tag' => 'button', 'disableWhenLoading' => true, 'loader' => false])
+
+@php
+    $tag = $attributes->hasAny('href', ':href', 'v-bind:href') ? 'a' : $tag;
+    $tag = $attributes->has('for') ? 'label' : $tag;
+
+    if ($loader) {
+        $attributes['v-bind:disabled'] = 'loading';
+    }
+@endphp
+
+<x-rapidez::tag
+    is="{{ $tag }}"
+    {{ $attributes->merge([
+        ':disabled' => $attributes->has('href') || $attributes->has(':href') || !$disableWhenLoading ? null : '$root.loading',
+    ]) }}
+>
+    <span class="contents" @if ($loader) v-bind:class="{'invisible': loading}" @endif>
+        {{ $slot }}
+    </span>
+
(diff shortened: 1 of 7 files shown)


resources/views/components/button/tag.blade.php
```diff
@@ -0,0 +1,27 @@
...
+>
+    <span class="contents" @if ($loader) v-bind:class="{'invisible': loading}" @endif>
+        {{ $slot }}
+    </span>
+
+    @if ($loader)
+        <div v-if="loading" class="absolute right-1/2 bottom-1/2 translate-x-1/2 translate-y-1/2" v-cloak>
+            <x-heroicon-o-arrow-path class="animate-spin size-5"/>
+        </div>
+    @endif
```
jordythevulder, line 26: we still use the animate-spin? that one looks really bad, cant we give this an upgrade?

Roene-JustBetter: What would you like to see instead of this? The same animation as the add to cart button? Like is done here: 
https://test.rapidez.io/men/tops-men/jackets-men.html?

Jade-GG (approved):

resources/views/components/button/tag.blade.php
```diff
@@ -0,0 +1,16 @@
+@props(['tag' => 'button', 'disableWhenLoading' => true, 'loader' => false])
+
+@php
+    $tag = $attributes->hasAny('href', ':href', 'v-bind:href') ? 'a' : $tag;
+    $tag = $attributes->has('for') ? 'label' : $tag;
+@endphp
+
+<x-rapidez::tag
+    is="{{ $tag }}"
+    :v-bind:class="$loader ? '{ \'button-loading\': $root.loading }' : ''"
```
Roene-JustBetter, line 10: Checked this with @indykoning. 
```
    {{ $attributes->merge([
        ':disabled' => $attributes->has('href') || $attributes->has(':href') || !$disableWhenLoading ? null : '$root.loading',
        'v-bind:class' => $loader ? '{ "button-loading": $root.loading }' : null
    ]) }}
```
Is not working.

JimmyHoenderdaal (approved):

PR #972, opened 2025-08-11
[4.x] Listen to all changes on address data instead of only manual input

Jade-GG:
When something like the postcodeservice package sets the street and city data for you in the checkout, no event gets sent to the graphql mutation saying that it has to mutate. This means that if it's the last data you need to fill in the address data (as it usually is), the address data will never get sent to magento until you change something inside of it.

By listening directly to the data, we can fix this. However, this means we also should turn every `v-model` into a `v-model.lazy` to make sure we don't catch this on every single button press. There's a debounce on it to be safe, but this should then also be changed in checkout-theme.


Jade-GG:
> Previously everything had a `v-model.lazy`, that changed to `v-model` in: #503 that's not causing any new, or bring back old issues?
> 
> Can't we simply emit the change event?

I don't see any reason why the v-models shouldn't be lazy. Realtime updating of the addresses while you're typing---especially in the graphql case---will just cause constant graphql requests to the server. Having them only update the data when the field gets unfocused (or submitted) seems very reasonable to me.

As for sending the change event, I'm not sure what you mean. If you're talking about inside of the postcode package: We don't have access to any part of the DOM in there, the only thing passed to [that function](https://github.com/rapidez/postcodeservice/blob/master/resources/js/package.js#L20-L41) is the (reactive) object with the address data, which then updates the field through the reactivity of the v-model.

royduin:
The lazy was previously removed in: https://github.com/rapidez/core/commit/187d7a734f9922809733c08d03bfe70e05b3757e by @indykoning

Jade-GG:
> This should be solved by triggering the input and change event on the element that should now trigger a mutation.

Again, we can't do this without refactoring the way that the postcode packages work. We only send through the reactive address object, and not the element that gets changed.

royduin:
So what's the plan? Is this ready to merge and approved? + there are some conflicts now

Jade-GG:
> So what's the plan? Is this ready to merge and approved? + there are some conflicts now

I've fixed the conflicts at least :)

Jade-GG:
No longer needed

PR #976, opened 2025-08-14
POC fullscreen search

jordythevulder:
I've got an idea. What if we made more use of the instant search functionalities by letting the user dive into a search experience which is fullscreen. This becomes mostly interesting on desktop screens (B2B audience). Note this is a quick draft that needs allot more work and refinement. Some functionalities that i'd like to add:

- Default slider or tiles which can be picked by hand to boost sales
- Default populair categories, products, filters

Because there are allot of products, i want the user to be able to scroll down within the fullscreen search, so it's not just limited to the screen height.

It's like filtering on a category overview page where everything is instant. Maybe even removing the search result page, or integrating the fullscreen search together.

Current:
image
image
image

something that i'd like it to be: (quick designs)
image
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,43 @@
 # Changelog 
 
-[Unreleased changes](https://github.com/rapidez/core/compare/5.0.0-alpha.4...5.0.0-alpha.4)
+[Unreleased changes](https://github.com/rapidez/core/compare/5.0.0-alpha.5...5.0.0-alpha.5)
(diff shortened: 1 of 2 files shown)


Roene-JustBetter:
Also a good thing to add is the add to cart button on the product tiles in the full screen autocomplete.

resources/views/components/autocomplete/input.blade.php
```diff
@@ -2,7 +2,7 @@
     id="autocomplete-form"
     method="get"
     action="{{ route('search') }}"
-    class="flex relative z-header-autocomplete"
```
Roene-JustBetter: Input should have the z-index?

PR #980, opened 2025-08-18
[4.x] Catch when login is required to view the status of one's order

Jade-GG:
If `allow_guest_on_existing_account` is set to true, it's possible to create an order on an existing account, which means that once it's been turned into a `sales_order` you can no longer query to it without having logged in. Currently this breaks the checkout success page entirely.

After this patch, the checkout success page stays working by keeping the old order data alive. However, you can't fetch the status of your order anymore, so the text for that has been changed to ask for a login.


Jade-GG:
> Do other functions not suffer similar consequences with a guest order in an existing email address?
> 
> Such as placing an order using a PSP such as Paynl and cancelling the payment?
> 
> (I'm asking since i know Paynl reactivates the existing quote, if the quote is now also linked to the user you can no longer check out)

I just tested this with Multisafepay (specifically because that's the payment provider used in the project I've got this patch already applied to) and cancelling the order works fine, I get sent back to the cart with a working quote I can update without having to log into my account.

PR #983, opened 2025-08-21
Added metadata for reviews

Roene-JustBetter:
Added metadata for: https://github.com/rapidez/reviews/pull/31

Added check around div where product sliders could render but when there are no product sliders the div will still be shown.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -70,7 +70,7 @@
         }
     },
     "conflict": {
-        "rapidez/reviews": "<3.0"
+        "rapidez/reviews": "<4.1"
     },
     "scripts": {
         "analyse": "phpstan --memory-limit=256M"


jordythevulder (approved):

tailwind.config.js
```diff
@@ -61,6 +61,7 @@ export default {
                     muted: color('--background-muted', colors.slate[50]),
                 },
                 backdrop: color('--backdrop', 'rgba(0, 0, 0, 0.4)'),
+                success: color('--success', '#00B67A'),
```
royduin, line 64: Maybe just use a green color from Tailwind somewhere? If we're going this way you may need a background, border, etc later, see: https://github.com/rapidez/core/blob/0470c917a82036e871f1547916acb787493d512a/config/rapidez/frontend.php#L82

resources/views/product/overview.blade.php
```diff
@@ -53,30 +56,23 @@
...
-            field="entity_id"
-            :value="$product->relation_ids"
-        />
-        <x-rapidez::productlist
-            title="We found other products you might like!"
-            field="entity_id"
-            :value="$product->upsell_ids"
-        />
-    </div>
+    @if (($product->relation_ids ?? false) || ($product->upsell_ids ?? false))
```
royduin, line 64: In the productlist there is a value check; you're doing this for the container?

indykoning (approved):

PR #1005, opened 2025-09-15
[4.x] Delay watch functions to avoid loading race condition

Jade-GG:
Depending on what order the code gets loaded in, these watch functions may be called before the variable exists. For example, on a vanilla Rapidez installation this can happen with the `useCart` watch if you don't install `rapidez/account`. Possibly something to do with how `useMask` imports it---it's not entirely clear how the imports end up getting resolved under different circumstances.

Because we don't have exact control over when and how things get loaded in, the best solution I could find is to make the watch get executed later. This still works exactly as it should, but makes sure everything is loaded in before.

The one in `useUser` is probably unnecessary, but I've added it anyway to be safe.

### DIFF
--- a/resources/js/stores/useCart.js
+++ b/resources/js/stores/useCart.js
@@ -271,7 +271,7 @@ export const taxTotal = (cart) => {
     return cart.prices.applied_taxes.reduce((sum, tax) => sum + tax.amount.value, 0)
 }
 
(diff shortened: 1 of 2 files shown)


Jade-GG:
This seems to unfortunately not fix anything after all :(

PR #1006, opened 2025-09-15
[4.x] Use setTimeout around mask section in useCart

Jade-GG:
#1005 didn't fix it, but this seems to do the trick.


Jade-GG:
> We're both talking about CPU throttling right?

I had only tried network throttling, but same story for CPU throttling (which doesn't exist on firefox, it seems). I wasn't able to reproduce the error with this patch in place, no matter what I throttled and how much.

PR #1007, opened 2025-09-15
Remove render blocking 404 download

indykoning:
This should raise the speed score as it no longer tries a render blocking 404 css download which is in the Magento demo by default with `<link rel="stylesheet" type="text/css"  media="all" href="{{MEDIA_URL}}styles.css" />`


royduin:
Seems like we can raise the best practices score to 100% now 😎 maybe good to keep 90% for the speed as it may vary between runs. First run 95% + 96% 🚀

PR #1013, opened 2025-09-16
Added checkbox and radio options

indykoning:
This PR adds the missing radio and checkbox product options.
Also the dropdown has been updated to allow "No selection" when it's not a required option:
image


royduin (approved):
Looking good to me, but a good review from others would be nice. I've created a follow-up for this to add some Playwright tests: RAP-1541. I'll merge when all approved.

Roene-JustBetter (approved):

PR #1014, opened 2025-09-17
Recently viewed widget sorting

royduin:
So the products in the widget are sorted in order of visits instead of by the default; score.

- We could sort directly in the DSL query with a custom script but this feels cleaner and it is better readable
- As `transform-items` is the only prop on the [ais-hits component ](https://www.algolia.com/doc/api-reference/widgets/hits/vue/) we could use I added it on the listing component to passthrough. This way we can just set in on the productlist component (as all attributes will be set on the listing Vue component) without the need of a Blade component prop with all possible variants (with and without v-bind, etc)

### DIFF
--- a/resources/js/components/Listing/Listing.vue
+++ b/resources/js/components/Listing/Listing.vue
@@ -33,6 +33,10 @@ export default {
             type: Boolean,
             default: false,
         },
+        transformItems: {
+            type: Function,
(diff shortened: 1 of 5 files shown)


royduin:
Follow up for the randomly failing Playwright test: RAP-1540

PR #1016, opened 2025-09-18
[4.x] Fix visual swatches that aren't defined as super attributes not being shown

Jade-GG:
Usually, color attributes would become a super attribute (e.g. in the Rapidez demo) which means this check passed without issues.

However, if you add custom data or otherwise only visual swatch data that's not a super attribute, this didn't work because it tried to fetch the data as if it were a string with a keyword field. This fixes that.


royduin:
Do you have a shorter commit message I can use to merge?

indykoning (approved):

PR #1019, opened 2025-09-24
Fire login event when logged in customer changes

indykoning:
This method moves some functionality closer to their connected store.
Due to listening to events like this the login and register buttons also experience a performance boost as they no longer need to wait for the cart to be fetched before sending you to the account page. (Due to turbo page navigations do not break background fetches)
Also supporting emitting the login event when the token has been added due to outside sources


royduin:
Do you have a short squash commit message I can use?

Jade-GG (approved):

PR #1029, opened 2025-09-29
[4.x] Workaround for query function not being called when there's no query string #2

Jade-GG:
Workaround until this PR is merged in searchkit: https://github.com/searchkit/searchkit/pull/1407


royduin:
So with this the category positions are working again? That's why the product order has changed in the screenshots?

Jade-GG:
> So with this the category positions are working again? That's why the product order has changed in the screenshots?

This is correct.

royduin:
Let's wait for @indykoning his review before merging.

indykoning (approved):

PR #1031, opened 2025-09-29
Prevent layout shift for product names

Roene-JustBetter:
The product name will show later so it will have a layout shift. With this the wrapping `<div>` will fill the space before the product names is showed, so it will prevent a layout shift.


royduin:
Maybe this is caused by the lazy loading of the highlight component here: https://github.com/rapidez/core/blob/d6d73f9b81089c26489207fc328eab320440bf8b/resources/js/instantsearch.js#L7

royduin:
Let's wait for @indykoning his review, maybe we need to do this with more components?

PR #1034, opened 2025-10-02
[4.x] Use checkoutLogin to actually access go function

Jade-GG:
I'm surprised this didn't cause any issues before? Possible I'm missing something.


indykoning:
See: https://docs.rapidez.io/4.x/graphql-components.html#grouping:~:text=%3C/form%3E-,Keep%20in%20mind,-partial%2Dsubmit%20must
The old situation _should_ be the functional one

PR #1037, opened 2025-10-06
[4.x] Make labels not clip outside of the frame & allow them to show together with inputs

Jade-GG:
Previously, these labels would clip outside of the container of the filters. Because they're wrapped in a `details` element, this means they will clip as if it were an `overflow-hidden` element. I've fixed this by making the labels move around so that they'll be hanging left of the thumb when on the right side, and hanging right of the thumb on the left side. See screenshots:
image
image

I also added a `labels` prop so that we can actually display the labels together with the inputs. As it turns out, we never noticed this bug because we always use the inputs and never show the labels.


tests/playwright/category.spec.js-snapshots/category-filter-1-chromium-linux.png
Roene-JustBetter, line 1: This doesn't look good. It should be the inputs or the labels not both.

Roene-JustBetter (changes requested):

PR #1039, opened 2025-10-07
Disable inputs on $root.loading

Roene-JustBetter:
Inputs don't get a visual state when they are disabled. They should be disabled when `$root.loading` is true when we send a request to Magento.

### DIFF
--- a/resources/views/checkout/partials/address.blade.php
+++ b/resources/views/checkout/partials/address.blade.php
@@ -42,6 +42,7 @@
                                 <x-rapidez::input
                                     name="{{ $type }}_company"
                                     v-model="variables.company"
+                                    v-bind:disabled="$root.loading"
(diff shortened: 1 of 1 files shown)


Jade-GG:
Playwright tests are failing, can you check? https://rapidez.github.io/playwright-reports/rapidez/core/1039/merge/18310775521/1/#?testId=24e4b3ac21f7f473b7e5-628e57c0403e85c27ce5

Roene-JustBetter:
> This is now overriding the input from the Blade Components? Can't we just add the loading instead of overwriting everything so we don't have to specify the classes again?

Yes this is overriding the input from the Blade Components. 
Where would you add the loading? We can't add it in the Blade Components and we don't want to place it on every single input.

PR #1072, opened 2025-11-13
[4.x] Fixed email is not valid spam

indykoning:
Due to the v-model not being lazy `debouncePromise` is triggered on every character typed.
Which is not a problem for `isEmailAvailable` with the debounce. However `handleGuest` attempts to set the email on the cart, where validation is done wether the e-mail address is valid.

This means for every time an email is checked for availability it is also getting blocked by the `setGuestEmailOnCart` graphql call.

We do not want the v-model being lazy since the CheckoutLogin won't know the entered email until the field gets unfocussed.

see: https://github.com/rapidez/core/pull/1258


Jade-GG (dismissed):

royduin:
Tests are failing? And master is v5 now, so we also need this on 4.x?

PR #1223, opened 2026-03-06
Listen to is_html_allowed_on_front for attribute values

indykoning:
In this PR we move the html escaping and listening to is_html_allowed_on_front to earlier in the process, this way it will get indexed with the escaping already done if it needs to be escaped.

htmlSpecialChars will always be allowed so entities like `&trade;` are always rendered as `™`
If html is not allowed, it will not turn any html tags into the html entities so they're rendered as plain text


src/Models/AbstractAttribute.php
```diff
@@ -62,8 +62,15 @@ protected function value(): Attribute
     {
         return Attribute::get(function ($value) {
             $value = $this->option_values ?? $this->rawValue ?? $value;
+            $value = array_map(function ($val) {
+                if ($this->is_html_allowed_on_front) {
+                    return $val;
+                }
 
-            if (is_iterable($value) && count($value) === 1) {
+                return e(htmlspecialchars_decode($val), false);
```
indykoning: We want this `htmlspecialchars_decode` first since magento sample data has the htmlentities already encoded

`Cocona&reg; performance fabric`
if we didn't decode this first it would turn into
`Cocona&amp;reg; performance fabric`

royduin:
We do have an escaping issue in the menu now, see the screenshots

PR #1236, opened 2026-03-18
[4.x] Auto retry requests where the cart has expired

Jade-GG:
ref: RAP-1854, BORDEX-1358

Currently, when your cart expires you will sometimes get an ugly and scary error telling you your cart ID is not valid.

With this PR, I check for this error and underwater we retry the request once, adding a more friendly notification telling you that your cart has expired in the meantime. This means that, for example, if you add a product to your expired cart, it will make a new cart and then re-add the product to this new cart.

This makes the user experience for expired carts a bit nicer.


indykoning (dismissed):
Code wise it's looking good.
But perhaps we can simplify it a lot by doing what we are doing in the useUser store: https://github.com/rapidez/core/blob/08c16ce6f25db1ead55561f3975ddb51f1f39502/resources/js/stores/useUser.js#L214

I reckon refreshing the cart after loading the page will tackle most of the problems users experience.

indykoning (dismissed):

indykoning (commented):
Your tests are failing, other than that LGTM

indykoning:
image

Jade-GG:
Error fixed by reverting the change, but now the screenshots are still slightly off with the font rendering.

indykoning (dismissed):

PR #1246, opened 2026-04-01
[5.x] Retain UTM fields in listing pages

Jade-GG:
ref: HDB-1060

Right now we end up stripping away every param from the route that's not a valid filter. However, we may want to retain the utm fields for tracking reasons.

Right now I've specifically focused on those fields, as it's harder to figure out exactly which fields are not part of the filters set. Also note that this will cause any attributes starting with `utm_` to no longer work with the routing.


resources/js/components/Listing/Listing.vue
indykoning, line 210: Do these always return the filters, but as undefined if they're empty?
If so we could load all url params first and let the other data override it.

If that doesn't work this workaround will have to do

Jade-GG: Nope, they just get deleted from the object entirely.

PR #1250, opened 2026-04-03
[5.x] Add roundCurrency helper to round correctly for any currency

Jade-GG:
Ref: none

The `sumPrices` function relied on the fact that most currencies have 2 decimals. This will work, until you run into a currency like Yen (0 decimals) or Dinar (3 decimals)[^1]

By using `Intl.NumberFormat` and getting the `maximumFractionDigits` from that, we can round properly. I've made a helper function for this and applied it to the `sumPrices` function.

[^1]: https://en.wikipedia.org/wiki/ISO_4217#List_of_ISO_4217_currency_codes

### DIFF
--- a/resources/js/helpers.js
+++ b/resources/js/helpers.js
@@ -31,13 +31,13 @@ window.productSpecialPrice = function (product) {
     return window.productPrice(product) > specialPrice ? specialPrice : null
 }
 
-window.roundCurrency = function (price) {
-    const formatter = new Intl.NumberFormat([(config.locale ?? 'default').replace('_', '-'), 'default'], {
-        style: 'currency',
-        currency: config.currency ?? 'eur',
-    })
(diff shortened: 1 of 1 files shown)


resources/js/helpers.js
```diff
@@ -31,8 +31,18 @@ window.productSpecialPrice = function (product) {
     return window.productPrice(product) > specialPrice ? specialPrice : null
 }
 
+window.roundCurrency = function (price) {
+    const formatter = new Intl.NumberFormat([(config.locale ?? 'default').replace('_', '-'), 'default'], {
+        style: 'currency',
+        currency: config.currency ?? 'eur',
+    })
+    const digits = formatter.resolvedOptions().maximumFractionDigits
```
indykoning: Perhaps we could save this in a variable to reduce the amount of numberformatter objects instantiated for each roundCurrency call

indykoning (dismissed):

PR #1253, opened 2026-04-07
[5.x] - Change button tags to anchor tags

Roene-JustBetter:
Changed the `<button>` tags to `<a>` tags. 
This is better for the Google bot see also: https://developers.google.com/search/docs/specialty/ecommerce/pagination-and-incremental-page-loading

### DIFF
--- a/tests/playwright/category.spec.js
+++ b/tests/playwright/category.spec.js
@@ -15,7 +15,7 @@ test('category pagination', BasePage.tags, async ({ page }) => {
     await page.goto(process.env.CATEGORY_URL_SIMPLE)
     await expect(page.getByTestId('listing-item')).toHaveCount(12)
     const firstProductPage1 = await page.getByTestId('listing-item').first().textContent()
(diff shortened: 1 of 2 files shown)


tests/playwright/category.spec.js
```diff
@@ -15,7 +15,7 @@ test('category pagination', BasePage.tags, async ({ page }) => {
     await page.goto(process.env.CATEGORY_URL_SIMPLE)
     await expect(page.getByTestId('listing-item')).toHaveCount(12)
     const firstProductPage1 = await page.getByTestId('listing-item').first().textContent()
-    await page.getByTestId('pagination').getByRole('button', { name: '2' }).click()
+    await page.getByTestId('pagination').getByText('2').click()
```
indykoning: Couldn't we use
```suggestion
    await page.getByTestId('pagination').getByRole('link', { name: '2' }).click()
```
Which seems more explicit than the content text

PR #1257, opened 2026-04-13
[5.x] Prevent logout flow being triggered when incorrect credentials are used

indykoning:
See: https://github.com/rapidez/core/pull/1054


Jade-GG (approved):

royduin:
Follow-up: RAP-1907

indykoning:
The screenshots are correct, they have just moved up a number

PR #1261, opened 2026-04-17
[5.x] Laravel 13 support

Jade-GG:
ref: RAP-1870


composer.json
```diff
@@ -20,23 +20,24 @@
...
         "lcobucci/clock": "^2.0|^3.2",
         "lcobucci/jwt": "^4.0|^5.3",
-        "rapidez/blade-components": "^2.0",
-        "rapidez/blade-directives": "^1.1",
-        "rapidez/laravel-multi-cache": "^2.1",
-        "rapidez/laravel-scout-elasticsearch": "^1.0",
+        "rapidez/blade-components": "^3.0",
+        "rapidez/blade-directives": "^2.0",
+        "rapidez/laravel-multi-cache": "^3.0",
+        "rapidez/laravel-scout-elasticsearch": "^2.0",
```
Jade-GG: These are right now assumed release tags:
- rapidez/blade-components#39
- rapidez/blade-directives#22
- rapidez/laravel-multi-cache#4
- rapidez/laravel-scout-elasticsearch#2

They're probably not correct (e.g. blade-directives will probably be 1.2.0)

Jade-GG:
Tests are of course failing because the tags don't exist yet.

src/Auth/MagentoCustomerTokenGuard.php
```diff
@@ -16,8 +15,10 @@ public static function register()
             $middleware->disableFor('token');
         });
 
-        auth()->extend('magento-customer', function (Application $app, string $name, array $config) {
-            return new self(auth()->createUserProvider($config['provider']), request(), 'token', 'token');
+        $authGuardCreator = fn ($config) => new self(auth()->createUserProvider($config['provider']), request(), 'token', 'token');
+
+        auth()->extend('magento-customer', function (Application $app, string $name, array $config) use ($authGuardCreator) {
+            return $authGuardCreator($config);
         });
```
indykoning, line 21: We could use `version_compare(app()->version(), '13.0.0', '>')` to determine if we're running laravel 13

indykoning (changes requested):
I would prefer not to drop support for Laravel 12 entirely

indykoning:
& don't forget to update the versions in the actions

PR #1263, opened 2026-04-20
[5.x] Add error handling for config.js

Jade-GG:
see: #1262 

5.x variant of the above PR, using `$emit` and `$on` in lieu of a clean way to watch the variable and still avoid the racing condition as much as possible.

NOTE: The reactivity seems to be broken for `v-if` in the blade template. This is also true in general for the `loading` variable, which gets used in the same way.


indykoning (dismissed):

royduin:
Follow-up: RAP-1906

PR #1264, opened 2026-04-22
[5.x] Use stock->backorders instead

Jade-GG:
ref: RAP-1876


royduin:
Maybe we should do something with enums as what's 0, 1 and 2 doing?

Jade-GG:
> Maybe we should do something with enums as what's 0, 1 and 2 doing?

Good idea, I've made RAP-1900 for this :)

PR #1267, opened 2026-04-23
[5.x] Always trim category name

Jade-GG:
As it turns out, other places also need this trimmed. By fixing that at the root we don't have to trim each location individually.

### DIFF
--- a/src/Models/Category.php
+++ b/src/Models/Category.php
@@ -55,6 +55,11 @@ class Category extends Model
             : ('catalog/category/view/id/' . $this->entity_id)));
     }
 
+    public function name(): Attribute
+    {
+        return Attribute::get(fn($name) => trim($name));
+    }
+
     public function products(): HasManyThrough
     {
         return $this->hasManyThrough(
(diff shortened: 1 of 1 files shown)


Jade-GG:
check toArray() before merge

Jade-GG:
As it turned out, `toArray` did in fact not work yet with this. However, I feel like the current solution is not ideal. We can probably fix this at the root, on `customAttributesToArrayData()`, right?

indykoning:
It is indeed unfortunately the current limitation with https://github.com/laravel/framework/blob/f374d48bc97144bf2ed6e76b1fa2743e615ad446/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L249
Perhaps they would be open to passing the value along to the attribute getter (as is the case if you were to use `->name`)

For now i reckon the simples solution is something like done with the price attribute.
https://github.com/rapidez/core/blob/4a11d846f52f866163c5fd3e66df5dd4dd721c2b/src/Models/Product.php#L268

Getting it from the passed value if it is passed, if it is null try to get it from the attributes.

PR #1273, opened 2026-04-30
Include simpleProduct in cart-add event

indykoning:
For analytics it will be useful to see the simple product added to the cart in case you're adding a configurable product.


royduin:
Is this somewhere mentioned in the docs?

PR #1274, opened 2026-05-04
[5.x] Add heartbeat to keep CSRF tokens alive

Jade-GG:
ref: RAP-1888

This PR adds an empty route under the `web` middleware to keep the CSRF token alive, which it then calls every 15 minutes. This should significantly reduce the issue where people are doing requests with invalid CSRF tokens.

### DIFF
--- a/resources/js/package.js
+++ b/resources/js/package.js
@@ -94,9 +94,12 @@ async function init() {
 
     // Heartbeat every 15 minutes to keep CSRF alive
     if (!window.heartbeat) {
-        window.heartbeat = setInterval(() => {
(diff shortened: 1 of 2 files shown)


royduin:
This vs https://github.com/mikebronner/laravel-caffeine?

Jade-GG:
> This vs https://github.com/mikebronner/laravel-caffeine?

Looks like that package does a similar thing but just has some methods to make it not run on certain pages. It's also using classic ajax requests. Seems both unnecessary and overkill in our case. It'll end up doing the same thing as this, except as an unnecessary extra dependency with more complexity.

PR #1275, opened 2026-05-04
[5.x] Add robots.txt from blade

Jade-GG:
ref: RAP-1889

This fixes the sitemap issue where it uses a relative URL (where the absolute URL is vastly preferred).

Overrides [the `robots.txt` file from rapidez/rapidez](https://github.com/rapidez/rapidez/blob/master/public/robots.txt). We should probably remove it from that repo in the next major release.


royduin:
Follow-up: RAP-1905 as we have the route twice

PR #1276, opened 2026-05-04
Ensure values are refs

indykoning:
This pr fixes the reactivity by wrapping primitive types in a ref.


resources/js/package.js
```diff
@@ -188,7 +188,7 @@ async function init() {
         window.app.config.globalProperties.config = window.config
 
         watch(fetchCount, function (count) {
-            app.config.globalProperties.loading = count > 0
+            app.config.globalProperties.loading.value = count > 0
```
Jade-GG, line 191: Is this `.value` needed anywhere else? Or is this the only one?

PR #1282, opened 2026-05-12
[4.x] Add has_options to elastic data

Jade-GG:
We use this column to determine whether or not to redirect when you click on the add to cart button in the listing. However, this was not selected nor added to the searchkit result attributes. As a result, this functionality didn't work.

### DIFF
--- a/config/rapidez/searchkit.php
+++ b/config/rapidez/searchkit.php
@@ -28,6 +28,7 @@ return [
         'images',
         'url',
         'thumbnail',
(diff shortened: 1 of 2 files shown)


royduin:
Do we also need this in v5?

Jade-GG:
> Do we also need this in v5?

I was going to say no, but it seems like we still don't select `has_options` in the searchable trait. I'll double check this and make a PR later today when I have time.

PR #1299, opened 2026-06-01
[5.x] Mage-os compatibility

Jade-GG:
Draft PR for now that contains whatever changes are needed to make mage-os work. For now that is:

- Removing the `errors` field from `PlaceOrderOutput`.


indykoning:
Closing, this seems to already have been fixed in the latest versions of Mage-OS

PR #1300, opened 2026-06-01
[5.x] Use enums for constant values

Jade-GG:
ref: RAP-1900

This PR replaces a bunch of `const`s and a few hardcoded values with backed enums. This is a breaking change, as it removes a bunch of consts that may be used in projects using Rapidez. These are:

- `Category::STATUS_*` -> `CategoryStatus::*`
- `EavAttribute::ENTITY_TYPE_*` -> `EntityType::*`
- `Product::STATUS_*` -> `ProductStatus::*`
- `Product::VISIBILITY_*` -> `Visibility::*`
- `Review::REVIEW_ENTITY_*` -> `ReviewEntity::*`

It's very possible that I've missed some use cases. We can always update those when we come across them.

### DIFF
--- a/resources/views/product/partials/addtocart.blade.php
+++ b/resources/views/product/partials/addtocart.blade.php
@@ -5,7 +5,7 @@
 >
     <form v-on:submit.prevent="addToCart.add" class="flex flex-col gap-5">
         <h1 class="text-3xl font-bold" itemprop="name">{{ $product->name }}</h1>
-        @if (!$product->stock->is_in_stock && $product->stock->backorders === 0)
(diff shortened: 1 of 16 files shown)


royduin:
Was this ready to merge or was there a backwards compatibility issue?

Jade-GG:
> Was this ready to merge or was there a backwards compatibility issue?

As far as I know we considered this ready to merge after fixing the failing tests. The only issue is that we have to be careful with updating projects to this version if we use any of these constants there. I wasn't able to find any compatibility issues with packages under the Rapidez namespace.

PR #1301, opened 2026-06-02
[5.x] - Change canonical links for paginated pages

JimmyHoenderdaal:
**Some explanation here:**

On product overview pages we have pagination to navigate to pages. In the past we used a `<button>` tag here instead of an `<a>` tag. Google prefers `<a>` with `href`. _This can help Googlebot (the Google web crawler) find subsequent pages._

We already fixed that [here](https://github.com/rapidez/core/pull/1253/changes).

Next thing Google prefers:

> _In addition, consider linking from all individual pages in a collection back to the first page of the collection to emphasize the start of the collection to Google. This can give Google a hint that the first page of a collection might be a better landing page than other pages in the collection._

We already do this, so we don't need to change anything here — we can always go back to page 1 from every page. See the screenshot below.

Screenshot 2026-07-22 at 10 32 36

Then the last part for pagination: **Use URLs correctly**

- **Give each page a unique URL.** For example, include a `?page=n` query parameter, as URLs in a paginated sequence are treated as separate pages by Google.
  We already do this — when we navigate to the next page, you'll see `?page=2` in the URL.

- **Don't use the first page of a paginated sequence as the canonical page. Instead, give each page its own canonical URL.**
  This is what we haven't done yet. When we visit the Rapidez demo and use the pagination, the canonical stays the same: `<link rel="canonical" href="https://demo.rapidez.io/men.html">`. This should instead be different for every page. So what we've added now is that every page has its own canonical, like this: `<link rel="canonical" href="http://rapidez.test/men/tops-men.html?page=3" />`.

---

Also added in this pull request: `<link rel="next" href="{{ $pageUrl($currentPageIndex + 1) }}" />`. In the past Google used this, but it isn't needed anymore so I removed it in this [commit](https://github.com/rapidez/core/pull/1301/changes/a860cb8660ce042d62a8f4b0d75dd15b0b94fa99)

See also for more info: https://developers.google.com/search/docs/specialty/ecommerce/pagination-and-incremental-page-loading


resources/views/layouts/partials/head/pagination.blade.php
```diff
@@ -0,0 +1,17 @@
+@php
+    $currentPage = max((int) request('page', 0), 0);
```
Jade-GG: Shouldn't the minimum value for page be `1` and not `0`?
```suggestion
    $currentPage = max((int) request('page', 1), 1);
```

resources/views/layouts/partials/head/pagination.blade.php
```diff
@@ -0,0 +1,17 @@
...
+    $lastPage = max((int) ceil($total / max((int) $perPage, 1)) - 1, 0);
```
Jade-GG: Given that we always define what $perPage is when using this component, and ceil will always return a float, we don't need to do so many casts and extra checks.

Doing the -1 also doesn't make sense if pages are 1-indexed. We now have an off-by-one error. (you can check this yourself by assuming `$perPage` is 2: if you have 5 products, you should have 3 pages, and `ceil(5/2) = 3` already.)

```suggestion
    $lastPage = max(ceil($total / $perPage), 1);
```

resources/views/layouts/partials/head/pagination.blade.php
```diff
@@ -0,0 +1,17 @@
...
+    $pageUrl = fn (int $page) => $page > 0 ? $url . '?' . http_build_query(['page' => $page]) : $url;
```
Jade-GG, line 4: This seems unnecessary. Aside from the fact that, again, pages are 1-indexed so this will never return false in the first place, you can just have `?page=1` without any issues. I'd remove this line and keep it simple with the lower two lines using something like

```
href={{ $url . '?page=' . $currentPage - 1 }}
```

Jade-GG (changes requested):

src/Models/Category.php
```diff
@@ -75,6 +75,14 @@ public function products(): HasManyThrough
         );
     }
 
+    public function listingProducts(): HasManyThrough
+    {
+        return $this->products()->whereInAttribute('visibility', [
+            Product::VISIBILITY_IN_CATALOG,
+            Product::VISIBILITY_BOTH,
```
Jade-GG: https://github.com/rapidez/core/pull/1300 🙂 sorry

Jade-GG (dismissed):

resources/views/category/overview.blade.php
```diff
@@ -2,7 +2,15 @@
 
 @section('title', $category->meta_title ?: $category->name)
 @section('description', $category->meta_description)
-@section('canonical', url($category->url))
+@if ($category->is_anchor)
+    @include('rapidez::layouts.partials.head.pagination', [
+        'url' => url($category->url),
+        'total' => $category->products()->count(),
```
royduin: Don't we already do a query to get the total count of products? Do we now get this twice?

resources/views/layouts/partials/head/pagination.blade.php
```diff
@@ -0,0 +1,7 @@
+@php
+    $perPage = max((int) $perPage, 1);
+    $currentPageIndex = max((int) request('page', 0), 0);
+    $pageUrl = fn (int $page) => $page > 0 ? $url . '?' . http_build_query(['page' => $page]) : $url;
+@endphp
+
+@section('canonical', $pageUrl($currentPageIndex))
```
royduin, line 7: I thought the idea with this was to have the next/prev meta links? With this we only get the current url with the page query param.

royduin:
We don't need this, see: https://chatgpt.com/share/6a6876f5-be0c-83eb-a4ac-05817b8cf96f

PR #1302, opened 2026-06-02
[5.x] -  Fix product structured data on PDP

JimmyHoenderdaal:
This adds some extra microdata inside the **schema Offer**:
`<meta itemprop="itemCondition" content="https://schema.org/NewCondition" />`
`priceValidUntil` 

Both are **recommended** to add inside the schema. 

_Internal ref: RAP-1899_

### DIFF
--- a/resources/views/components/breadcrumb.blade.php
+++ b/resources/views/components/breadcrumb.blade.php
@@ -2,11 +2,10 @@
 
 <li class="flex items-center shrink-0 whitespace-nowrap" itemprop="itemListElement" itemscope itemtype="https://schema.org/ListItem">
     @if (!$active)
-        <a {{ $attributes->merge(['class' => 'text-sm hover:underline', 'href' => url($url)]) }}>
+        <a {{ $attributes->merge(['class' => 'text-sm hover:underline', 'href' => url($url), 'itemprop' => 'item']) }}>
             <span itemprop="name">{{ $slot }}</span>
+            <meta itemprop="position" content="{{ $position }}" />
         </a>
-        <meta itemprop="item" content="{{ url($url) }}" />
-        <meta itemprop="position" content="{{ $position }}" />
         <x-heroicon-o-chevron-right class="size-3 text-muted translate-y-px mx-1" stroke-width="2" />
     @else
         <span class="text-sm" itemprop="name">{{ $slot }}</span>
--- a/resources/views/components/breadcrumb.blade.php
+++ b/resources/views/components/breadcrumb.blade.php
@@ -2,10 +2,11 @@
 
 <li class="flex items-center shrink-0 whitespace-nowrap" itemprop="itemListElement" itemscope itemtype="https://schema.org/ListItem">
     @if (!$active)
-        <a {{ $attributes->merge(['class' => 'text-sm hover:underline', 'href' => url($url), 'itemprop' => 'item']) }}>
+        <a {{ $attributes->merge(['class' => 'text-sm hover:underline', 'href' => url($url)]) }}>
             <span itemprop="name">{{ $slot }}</span>
-            <meta itemprop="position" content="{{ $position }}" />
         </a>
+        <meta itemprop="item" content="{{ url($url) }}" />
+        <meta itemprop="position" content="{{ $position }}" />
         <x-heroicon-o-chevron-right class="size-3 text-muted translate-y-px mx-1" stroke-width="2" />
     @else
         <span class="text-sm" itemprop="name">{{ $slot }}</span>
--- a/resources/views/product/partials/microdata.blade.php
+++ b/resources/views/product/partials/microdata.blade.php
@@ -1,5 +1,15 @@
+@php
+    $description = $product->description
+        ? str(html_entity_decode(strip_tags($product->description), ENT_QUOTES | ENT_HTML5))->squish()
+        : null;
+@endphp
+
+<meta itemprop="name" content="{{ $product->name }}" />
 <meta itemprop="mpn" content="{{ $product->entity_id }}" />
 <meta itemprop="sku" content="{{ $product->sku }}" />
+@if ($description)
+    <meta itemprop="description" content="{{ $description }}" />
+@endif
 
 @foreach ($product->images as $image)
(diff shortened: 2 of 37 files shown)


resources/views/product/partials/microdata.blade.php
```diff
@@ -1,16 +1,27 @@
+@php
+    $description = $product->description
+        ? str(html_entity_decode(strip_tags($product->description), ENT_QUOTES | ENT_HTML5))->squish()
+        : null;
+@endphp
```
Jade-GG: Does this work correctly with magento content in the description? Is this all necessary? 🤔

resources/views/product/partials/microdata.blade.php
```diff
@@ -1,16 +1,27 @@
...
 
 <div itemprop="offers" itemtype="https://schema.org/Offer" itemscope>
     <meta itemprop="availability" content="https://schema.org/{{ $product->stock->is_in_stock ? 'InStock' : 'OutOfStock' }}" />
+    <meta itemprop="itemCondition" content="https://schema.org/NewCondition" />
     <meta itemprop="priceCurrency" content="@config('currency/options/default')" />
     <meta itemprop="price" content="{{ round($product->special_price ?: $product->price, 2) }}" />
     <meta itemprop="url" content="{{ url($product->url) }}" />
     @if ($product->special_to_date && $product->special_to_date > now()->toDateTimeString())
-        <meta itemprop="priceValidUntil" content="{{ $product->special_to_date }}" />
+        <meta itemprop="priceValidUntil" content="{{ \Illuminate\Support\Carbon::parse($product->special_to_date)->toDateString() }}" />
```
Jade-GG: Can you double check whether this causes any issues if there is a timezone set?

Jade-GG (changes requested):

Jade-GG (dismissed):

resources/views/product/partials/microdata.blade.php
```diff
@@ -1,16 +1,21 @@
+<meta itemprop="name" content="{{ $product->name }}" />
 <meta itemprop="mpn" content="{{ $product->entity_id }}" />
 <meta itemprop="sku" content="{{ $product->sku }}" />
+@if ($product->description)
+    <meta itemprop="description" content="{{ str($product->description)->stripTags()->squish() }}" />
+@endif
```
royduin: With this we do get the product description twice in the source

resources/views/product/partials/microdata.blade.php
```diff
@@ -1,16 +1,21 @@
...
 
 <div itemprop="offers" itemtype="https://schema.org/Offer" itemscope>
     <meta itemprop="availability" content="https://schema.org/{{ $product->stock->is_in_stock ? 'InStock' : 'OutOfStock' }}" />
+    <meta itemprop="itemCondition" content="https://schema.org/NewCondition" />
     <meta itemprop="priceCurrency" content="@config('currency/options/default')" />
     <meta itemprop="price" content="{{ round($product->special_price ?: $product->price, 2) }}" />
     <meta itemprop="url" content="{{ url($product->url) }}" />
     @if ($product->special_to_date && $product->special_to_date > now()->toDateTimeString())
-        <meta itemprop="priceValidUntil" content="{{ $product->special_to_date }}" />
+        <meta itemprop="priceValidUntil" content="{{ str($product->special_to_date)->before(' ')->before('T') }}" />
```
royduin: Can you explain this one?

resources/views/components/breadcrumb.blade.php
```diff
@@ -2,10 +2,11 @@
 
 <li class="flex items-center shrink-0 whitespace-nowrap" itemprop="itemListElement" itemscope itemtype="https://schema.org/ListItem">
     @if (!$active)
-        <a {{ $attributes->merge(['class' => 'text-sm hover:underline', 'href' => url($url), 'itemprop' => 'item']) }}>
+        <a {{ $attributes->merge(['class' => 'text-sm hover:underline', 'href' => url($url)]) }}>
             <span itemprop="name">{{ $slot }}</span>
-            <meta itemprop="position" content="{{ $position }}" />
         </a>
+        <meta itemprop="item" content="{{ url($url) }}" />
```
royduin: With this we do get the url twice? I don't see how this is an improvement, can you explain?

resources/views/product/partials/microdata.blade.php
JimmyHoenderdaal: Google expects a date without a time. This removes the time from Magento’s value

royduin: It's cleaner to use the some formatting; `->format(...)` and cast the attribute: https://laravel.com/docs/13.x/eloquent-mutators#attribute-casting

resources/views/product/partials/microdata.blade.php
```diff
@@ -7,10 +7,12 @@
 
 <div itemprop="offers" itemtype="https://schema.org/Offer" itemscope>
     <meta itemprop="availability" content="https://schema.org/{{ $product->stock->is_in_stock ? 'InStock' : 'OutOfStock' }}" />
+    <meta itemprop="itemCondition" content="https://schema.org/NewCondition" />
     <meta itemprop="priceCurrency" content="@config('currency/options/default')" />
     <meta itemprop="price" content="{{ round($product->special_price ?: $product->price, 2) }}" />
     <meta itemprop="url" content="{{ url($product->url) }}" />
-    @if ($product->special_to_date && $product->special_to_date > now()->toDateTimeString())
-        <meta itemprop="priceValidUntil" content="{{ $product->special_to_date }}" />
+    @php $specialToDate = $product->raw('special_to_date'); @endphp
```
royduin, line 14: Why do we need the `raw()`?

Roene-JustBetter: Normal `$product->special_to_date` access runs the value through HTML-escaping meant for display text, which turns the underlying Carbon date into a plain string. `raw()` skips that and returns the attribute's cast value directly — a real Carbon instance, since AttributeDatetime casts it to 'datetime'. The template needs that Carbon object to call `->isFuture()` and `->format('Y-m-d')`.

RAP-2010 created to remove html escaping from EAV dates

PR #1309, opened 2026-06-10
[5.x] Add success step

Jade-GG:
This step seems to be necessary, as we ran into this issue on certain projects where the checkout success did not work. This may specifically only go wrong when using a payment provider, I'm not certain on the details.


royduin:
A bit more details would be nice, will it play nice with the progress bar? Why didn't we need this earlier?

royduin:
Let's wait for RAP-1908 so we can see possible visual differences

royduin:
Could you update this PR and provide some more info?

Jade-GG:
PR updated; the visual difference here is that it adds the success step to the list of steps at the top of the page during the checkout.

We currently have the `success` step added in the overwritten config in all v5 projects. While I don't remember exactly what went wrong, it had to do with not being able to find a success step after finishing payment and that caused it to throw you back into the cart instead of into the checkout success.

royduin:
@Jade-GG is going to check this in the projects as 2 of them do have this but the latest one doesn't have it. Maybe it's related to payment providers? If it's still needed a new PR will be opened.

PR #1311, opened 2026-06-18
[5.x] Limit search query count when there are a lot of them

Jade-GG:
ref: AN-425

We ran into an issue on a site that had 300k+ search queries in the search_query table. While this table should probably be cleaned up, this also presented an issue where the indexer would break due to too many items being indexed, causing the indexing to take too long (and wasting resources).

Originally, I had wanted to do this with a simple `orderByDesc('popularity')->limit(10000)` directly on the query itself. However, that didn't work as scout will modify the query stripping off the limit in the process.

Instead, this PR sets a minimum popularity value based on what the 10000th highest popularity value is.

### DIFF
--- a/src/Models/SearchQuery.php
+++ b/src/Models/SearchQuery.php
@@ -3,6 +3,8 @@
 namespace Rapidez\Core\Models;
 
 use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Support\Facades\Cache;
+use Illuminate\Support\Facades\DB;
 use Rapidez\Core\Models\Scopes\IsActiveScope;
 use Rapidez\Core\Models\Traits\Searchable;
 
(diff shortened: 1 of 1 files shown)


src/Models/SearchQuery.php
```diff
@@ -24,10 +26,23 @@ protected static function booting()
...
+        // Decide on a minimum popularity to make sure we don't index too much.
+        // Simply using an orderBy with a limit on the query doesn't work as the limit gets stripped off.
+        $minPopularity = Cache::driver('array')->rememberForever(
+            'minPopularity',
+            fn () => DB::table('search_query')
+                ->orderByDesc('popularity')
+                ->limit(1)
+                ->offset(10000)
+                ->first()
+                ->popularity ?? 0
```
indykoning, line 38: If i remember correctly Mysql behaves badly with offsets.
Loading the first 10 000 records and throwing away 9 999 of them which is heavy on the database
https://medium.com/@twly.tw/efficient-pagination-in-mysql-avoiding-the-pitfalls-of-limit-offset-aef28f321065

PR #1313, opened 2026-06-22
[4.x] Add useNotifications store for notifications

Jade-GG:
See: #1314

### DIFF
--- a/resources/js/stores/useNotifications.js
+++ b/resources/js/stores/useNotifications.js
@@ -1,0 +1,23 @@
+export let pendingNotifications = []
+
+export function pushNotification(notification) {
+    let hasPreviousNotifications = pendingNotifications.length > 0
+    pendingNotifications.push(notification)
+
+    // Immediately trigger notifications if possible
+    if (window.app && window.app.$emit) {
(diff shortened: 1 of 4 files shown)


resources/js/stores/useNotifications.js
indykoning, line 1: What we could try here, is by making `pendingNotifications` reactive and making the `Notifications.vue` watch the variable we no longer need to fire events.

If we then read the `pendingNotifications` on mount in `Notifications.vue` we no longer need to know wether Vue is loaded or emit is available. Since `Notifications.vue` asks for the data when it itself is ready. And subscribes to updates using a watcher.

Then the notificationsStore simply needs to push and pop notifications to it't store

Jade-GG: Would making a reactive variable be possible when this code gets loaded in before Vue does?

indykoning: Yes, we do that with useFetches https://github.com/rapidez/core/blob/master/resources/js/stores/useFetches.js
and all the other stores as well

indykoning (approved):

PR #1314, opened 2026-06-22
[5.x] Add useNotifications store for notifications

Jade-GG:
ref: RAP-1708, BORDEX-1395

This store allows for notifications to be pushed regardless of Vue being loaded in or not. If Vue is not loaded, these will then get triggered upon vue being loaded in.

### DIFF
--- a/resources/js/package.js
+++ b/resources/js/package.js
@@ -1,6 +1,15 @@
+import { pushNotification } from './stores/useNotifications'
+
 window.debug = import.meta.env.VITE_DEBUG == 'true'
-window.Notify = (message, type = 'info', params = [], link = null) =>
-    setTimeout(() => window.$emit('notification-message', message, type, params, link), window.app ? 0 : 500)
+window.Notify = (message, type = 'info', params = [], link = null) => {
+    pushNotification({
+        message: message,
+        type: type,
+        params: params,
+        link: link,
+        timestamp: +new Date(),
+    })
+}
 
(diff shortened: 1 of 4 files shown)


resources/js/stores/useNotifications.js
```diff
@@ -0,0 +1,29 @@
...
+            'notification-message',
+            notification.message ?? '',
+            notification.type ?? 'info',
+            notification.params ?? [],
+            notification.link ?? null,
+        )
+    })
+}
+
+window.document.addEventListener('vue:loaded', triggerNotifications)
```
Jade-GG: I originally had an extra `setTimeout` in here, but without it this doesn't seem to cause a race condition with the `Notifications.vue` component possibly loading in later than this. As such, I've left it out.

resources/js/package.js
```diff
@@ -1,6 +1,15 @@
...
 window.debug = import.meta.env.VITE_DEBUG == 'true'
-window.Notify = (message, type = 'info', params = [], link = null) =>
-    setTimeout(() => window.$emit('notification-message', message, type, params, link), window.app ? 0 : 500)
+window.Notify = (message, type = 'info', params = [], link = null) => {
+    pushNotification({
+        message: message,
+        type: type,
+        params: params,
+        link: link,
+        timestamp: +new Date(),
```
Jade-GG, line 10: This is not used, but I figured it could be useful at some point.

PR #1317, opened 2026-06-26
[5.x] Fix v-else

Jade-GG:
Without this, the generated html contains `v-else="v-else"` which then tries to parse `v-else` as if it were javascript code. This gives warnings/errors.


indykoning (dismissed):

royduin:
Do we have the same issue with the `v-cloak`?

Jade-GG:
> Do we have the same issue with the `v-cloak`?

Not as far as I can tell. It does generate the same html, but it doesn't appear to find it a problem.

royduin:
On https://demo.rapidez.io/login I do see: `<svg v-else="1" v-cloak="1" class="h-5" ...`, maybe nice to fix it for both attributes?

PR #1318, opened 2026-06-26
[5.x] Prioritize core for rapidez namespace

Jade-GG:
ref: RAP-1916

This PR makes sure that when using a component under the Rapidez namespace, the component from the core always gets prioritized over other packages. Note that I added an extra hint path to make sure any custom overwrites within the project do still come first.

We ran into this issue specifically because of the `blade-components` package having a `button/tag` component that differed too much from the core's `button/tag` component. I'm not sure why we decided this should be the case.

It may be useful to use this same trick for the `core-overwrites` in checkout-theme.

### DIFF
--- a/src/RapidezServiceProvider.php
+++ b/src/RapidezServiceProvider.php
@@ -179,8 +179,6 @@ class RapidezServiceProvider extends ServiceProvider
 
     protected function bootViews(): self
     {
-        $this->loadViewsFrom(__DIR__ . '/../resources/views', 'rapidez');
-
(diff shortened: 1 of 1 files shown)


indykoning (dismissed):
Do we still need the loadViewsFrom when calling prependNamespace?

Jade-GG:
> Do we still need the loadViewsFrom when calling prependNamespace?

Apparently not. I double checked that it works without, and it also turns out that `loadViewsFrom` is basically just a wrapper for `addNamespace(...)`.

PR #1319, opened 2026-06-29
[5.x] Add message for .env changes

Jade-GG:
Because the .env will not get reloaded and cannot get reloaded mid-run.


src/Commands/InstallCommand.php
```diff
@@ -94,6 +94,7 @@ protected function validate()
             if ($this->call('rapidez:validate')) {
                 $this->newLine();
                 $this->error('The validation did not pass, please fix the errors first');
+                $this->error('Note: If you need to change something in the .env, you will need to restart this command');
```
indykoning:
```suggestion
                $this->error('Note: If you need to change something in the .env, you should answer no to the following question');
```
Maybe something like this to make the expected action clearer

PR #1320, opened 2026-07-01
[5.x] Support multiple categories in indexer

Jade-GG:
This was written wrong, using `breadcrumbCategories` which only uses the longest path. I've added an attribute that can be used to get the same data in a collection from not just the longest path. Flattening this gives us the same data in the indexer, but then for all categories.

### DIFF
--- a/src/Models/Product.php
+++ b/src/Models/Product.php
@@ -339,6 +339,20 @@ class Product extends Model
         });
     }
 
+    protected function allCategories(): Attribute
+    {
+        return Attribute::get(function (): Collection {
+            $this->loadMissing('categoryProducts');
+
+            $categories = $this->categoryProducts
+                ->where('category_id', '!=', config('rapidez.root_category_id'))
+                ->pluck('category')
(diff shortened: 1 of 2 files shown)


src/Models/Product.php
```diff
@@ -339,6 +339,20 @@ protected function categoryIds(): Attribute
...
     }
 
+    protected function allCategories(): Attribute
+    {
+        return Attribute::get(function (): Collection {
+            $this->loadMissing('categoryProducts');
+
+            $categories = $this->categoryProducts
+                ->where('category_id', '!=', config('rapidez.root_category_id'))
+                ->pluck('category')
```
royduin, line 349: Do we really need the whole category? We do have all the category info here: https://github.com/rapidez/core/blob/11cc74938bcca67a4701265a4658e34842f63fae/src/Models/Traits/Product/Searchable.php#L135 so maybe an ID is enough?

royduin:
Closing in favor of: https://github.com/rapidez/core/pull/1343

PR #1324, opened 2026-07-08
[5.x] Implement weee_tax models

Jade-GG:
ref: BORDEX-1189

For this I also had to implement a `ForCurrentWebsiteWithoutLimitScope` which I just copied from the `store` variant and search/replaced the words.

### DIFF
--- a/src/Models/WeeeTax.php
+++ b/src/Models/WeeeTax.php
@@ -21,8 +21,7 @@ class WeeeTax extends Model
     protected static function booting()
     {
         static::addGlobalScope('website', new ForCurrentWebsiteWithoutLimitScope(['entity_id', 'attribute_id']));
-        static::addGlobalScope('withAttributeData', fn (Builder $query) =>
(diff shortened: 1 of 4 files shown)


src/Models/WeeeTax.php
```diff
@@ -0,0 +1,36 @@
...
+namespace Rapidez\Core\Models;
+
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\Relations\BelongsTo;
+use Rapidez\Core\Models\Scopes\ForCurrentWebsiteWithoutLimitScope;
+
+class WeeeTax extends Model
+{
+    const CREATED_AT = null;
+    const UPDATED_AT = null;
```
royduin: `public $timestamps = false;` is cleaner? And used on other models

Jade-GG: Ah, I copied [SearchSynonym.php](https://github.com/rapidez/core/blob/master/src/Models/SearchSynonym.php) without really looking at this. I can change both.

PR #1326, opened 2026-07-09
Install script

royduin:
An install script to make it easier to try out Rapidez. At the end we like to put this on the Rapidez.io homepage:
```
curl -fsSL https://rapidez.io/install | bash
```
Which should point to this install file.

Requirements / todos:

- [x] Just a oneliner to copy paste
- [x] It should run next to any other services like Valet, Herd, MySQL, etc
- [x] Browsing and editing templates should be easy and locally
- [ ] It can run completely independent fully in Docker
  - [x] Webserver
  - [x] MySQL
  - [x] Elasticsearch
  - [ ] Redis
- [ ] It should run on macOS and Linux
- [ ] Magento admin is accessible
- [ ] Products are visible in listings
- [ ] Placing an order is possible
- [ ] Tested on macOS
- [ ] Tested on Linux (on a fresh VPS)

Ref: RAP-1921 + RAP-1957

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,0 +1,105 @@
+#!/usr/bin/env bash
+set -e
+
+echo '====================================='
+echo '  ____             _     _          '
+echo ' |  _ \ __ _ _ __ (_) __| | ___ ____'
+echo ' | |_) / _` | '\''_ \| |/ _` |/ _ \_  /'
+echo ' |  _ < (_| | |_) | | (_| |  __// / '
+echo ' |_| \_\__,_| .__/|_|\__,_|\___/___|'
+echo '            |_|                     '
+echo '  Welcome to the Rapidez installer!'
+echo '====================================='
+
+# Vars
+DEFAULT_TRAEFIK_PORT=80
+DEFAULT_DB_PORT=3306
+INSTALL_DIR="${RAPIDEZ_HOME:-rapidez}"
+
+# Helpers
+command_exists() {
+  command -v "$1" >/dev/null 2>&1
+}
+
+find_free_port() {
+  local port=$1
+
+  while nc -z localhost "$port" >/dev/null 2>&1; do
+    port=$((port + 1))
+  done
+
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -0,0 +1,105 @@
...
+  exit 1
+fi
+
+echo "✓ Docker found"
+
+mkdir -p "$INSTALL_DIR"
+cd "$INSTALL_DIR"
+
+composer create-project rapidez/rapidez:^5.0 .
+composer require rapidez/reviews
```
Jade-GG, line 53: What makes this package special enough that it needs to be included in a base install of Rapidez? If it's that special it may as well go into the core IMO 🤔

install.sh
```diff
@@ -0,0 +1,105 @@
...
+echo "✓ .env configured"
+
+docker exec rapidez php artisan key:generate
+echo "✓ key generated"
+docker exec rapidez php artisan rapidez:install --frontendonly
+echo "✓ frontend dependencies copied"
+docker exec rapidez yarn
+echo "✓ frontend dependencies installed"
+docker exec rapidez yarn run prod
+echo "✓ frontend dependencies build"
```
Jade-GG, line 80:
```suggestion
echo "✓ frontend dependencies built"
```

Past tense 😉

install.sh
```diff
@@ -0,0 +1,105 @@
...
+docker exec rapidez php artisan storage:link
+echo "✓ storage linked"
+docker exec rapidez php artisan rapidez:index
+echo "✓ products indexed"
+docker exec rapidez php artisan vendor:publish --provider "Rapidez\Core\RapidezServiceProvider"
+echo "✓ configs published"
```
Jade-GG, line 86: This also publishes the views, right?

```suggestion
echo "✓ configs, views, and translation files published"
```

Jade-GG (changes requested):

PR #1327, opened 2026-07-13
Finish instantsearch fixes

indykoning:
These changes complete the todos left from https://github.com/rapidez/core/pull/1316
Keeping the amount of instantsearch queries at 1


resources/js/components/Listing/AisClearRefinements.vue
indykoning, line 1: Unfortunately the AisClearRefinements component had no proper way to ignore the category_attributes in another way without causing another request to be done.
Any v-bind on a valid prop would cause it to duplicate _msearch requests.

resources/js/instantsearch.js
```diff
@@ -31,10 +31,7 @@ document.addEventListener('vue:loaded', function (event) {
     vue.component('ais-stats', component('Stats'))
 
     // Custom component
-    vue.component(
-        'ais-stats-analytics',
-        defineAsyncComponent(() => import('./components/Search/AisStatsAnalytics.vue')),
-    )
+    vue.component('ais-stats-analytics', component('StatsAnalytics'))
```
indykoning, line 34: Since statsAnalytics also used instantsearch it being the only separate lazy loaded element caused it to render later and create another msearch call

resources/views/listing/partials/filter/category.blade.php
```diff
@@ -1,6 +1,5 @@
 <ais-hierarchical-menu
-    v-bind:attributes="listingSlotProps.categoryAttributes"
-    v-bind:sort-by="['count','name']"
```
indykoning, line 3: Oddly enough only the sort-by bind caused another _msearch call.
This is implicitly the default so we could remove it.

PR #1329, opened 2026-07-14
[5.x] Improve turbo frame caching

Jade-GG:
ref: RAP-1969

This PR adds some basic standardized functionality for turbo frames, which allows us to improve the caching, adding the cachekey into the url. This solves the issue of turbo frames staying in cache longer than necessary when you clear the cache, which then in turn also often ended up causing a cache loop.

For this purpose, I've added a directive that can be used to insert turbo frames. It should be used like this in templates:

```blade
@turboframe('turbo-frame')
```

This then places the intended view with the `$complete` variable set to `false` directly on the page, and creates a turbo frame that will overwrite this with the same view but with `$complete` set to `true`.

For example, say you have a view included somewhere that looks like this, with one very heavy query:

```blade
<div class="...">
    Lorem ipsum dolor sit amet...
    @foreach (config('rapidez.models.category')::all() as $category)
        <div>{{ $category->name }}</div>
    @endforeach
</div>
```

We could change the `@include` to a `@turboframe`, and then modify the view like so:

```blade
<div class="...">
    Lorem ipsum dolor sit amet...
    @if ($complete ?? false)
        @foreach (config('rapidez.models.category')::all() as $category)
            <div>{{ $category->name }}</div>
        @endforeach
    @else
        {{-- You can also add a loading state if necessary! --}}
        <div>Loading...</div>
    @endif
</div>
```


Jade-GG:
Depends on rapidez/blade-directives#24

Jade-GG:
Playwright errors caused by:

> Access to fetch at 'http://localhost:1234/graphql' from origin 'http://localhost:8000' has been blocked by CORS policy: The 'Access-Control-Allow-Origin' header contains multiple values '*, *', but only one is allowed. Have the server send the header with a valid value.

royduin:
- How is it going with the Playwright tests @indykoning?
- Is this documented in the docs?

Jade-GG:
> * Is this documented in the docs?

I've not done anything to the docs yet, as I'd rather wait for approval first so I don't end up having to rewrite it in case we choose a different approach after all. However I can add it to my backlog of things to do :)

PR #1334, opened 2026-07-20
Perform current slide calculation based on child elements

indykoning:
This PR does the following to fix the bugs with irregularly sized sliders:
- Fixes the currentSlide to use the child elements to determine the slide number.
This way slides with different sizes no longer interfere with the current slide number
- Clamp the navigate on a min and max, so when sliders are supposed to slide multiple items (say currensSlide + visibleSlides) it can no longer try to go out of bounds.
- Calculate slidesVisible using the average slide width/length instead of the first child

Ref: FW-1849

### DIFF
--- a/resources/js/components/Elements/Slider.vue
+++ b/resources/js/components/Elements/Slider.vue
@@ -151,10 +151,7 @@ export default {
             this.navigate(next)
         },
         navigate(index, behavior = 'smooth') {
-            index = Math.min(
-                Math.max(index, 0),
-                this.container?.children.length - 1
-            )
+            index = Math.min(Math.max(index, 0), this.container?.children.length - 1)
 
             this.vertical
                 ? this.slider.scrollTo({ top: this.container?.children[index]?.offsetTop, behavior: behavior })
--- a/resources/js/components/Elements/Slider.vue
+++ b/resources/js/components/Elements/Slider.vue
@@ -232,7 +232,8 @@ export default {
             }
 
             return Math.round(
-                (this.sliderSpan / (this.vertical ? this.container.scrollHeight : this.container.scrollWidth)) * this.container.children.length,
+                (this.sliderSpan / (this.vertical ? this.container.scrollHeight : this.container.scrollWidth)) *
+                    this.container.children.length,
             )
         },
         slidesTotal() {
--- a/resources/js/components/Elements/Slider.vue
+++ b/resources/js/components/Elements/Slider.vue
@@ -151,13 +151,7 @@ export default {
             this.navigate(next)
         },
         navigate(index, behavior = 'smooth') {
-            index = Math.min(
-                Math.max(
-                    this.loop ? index + this.slides.length : index,
-                    0
-                ),
-                this.container?.children.length - 1
-            )
+            index = Math.min(Math.max(this.loop ? index + this.slides.length : index, 0), this.container?.children.length - 1)
 
             this.vertical
                 ? this.slider.scrollTo({ top: this.container?.children[index]?.offsetTop, behavior: behavior })
@@ -213,25 +207,21 @@ export default {
             const getSlideByGuess = (slide) => {
                 const bestGuessChild = this.container.children[slide]
                 if (!bestGuessChild) {
-                    return slide;
+                    return slide
                 }
-                const bestGuessSpan = this.vertical
-                        ? bestGuessChild.offsetHeight
-                        : bestGuessChild.offsetWidth
-                const bestGuessOffset = this.vertical
-                        ? bestGuessChild.offsetTop
-                        : bestGuessChild.offsetLeft
+                const bestGuessSpan = this.vertical ? bestGuessChild.offsetHeight : bestGuessChild.offsetWidth
+                const bestGuessOffset = this.vertical ? bestGuessChild.offsetTop : bestGuessChild.offsetLeft
 
                 // Past halfway to the slide it will snap
-                if (this.position + (bestGuessSpan / 2) < bestGuessOffset) {
-                    return getSlideByGuess(slide - 1);
+                if (this.position + bestGuessSpan / 2 < bestGuessOffset) {
+                    return getSlideByGuess(slide - 1)
                 }
 
                 if (this.position >= bestGuessOffset + bestGuessSpan) {
-                    return getSlideByGuess(slide + 1);
+                    return getSlideByGuess(slide + 1)
                 }
 
-                return slide;
+                return slide
             }
 
             return getSlideByGuess(bestGuess)
@@ -241,7 +231,9 @@ export default {
                 return 0
             }
 
-            return Math.round(this.sliderSpan / (this.vertical ? this.container.scrollHeight : this.container.scrollWidth) * this.slides.length)
+            return Math.round(
+                (this.sliderSpan / (this.vertical ? this.container.scrollHeight : this.container.scrollWidth)) * this.slides.length,
+            )
         },
         slidesTotal() {
             if (!this.mounted) {
--- a/resources/js/components/Elements/Slider.vue
+++ b/resources/js/components/Elements/Slider.vue
@@ -151,7 +151,13 @@ export default {
             this.navigate(next)
         },
         navigate(index, behavior = 'smooth') {
-            index = this.loop ? index + this.slides.length : index
+            index = Math.min(
+                Math.max(
+                    this.loop ? index + this.slides.length : index,
+                    0
+                ),
+                this.container?.children.length - 1
+            )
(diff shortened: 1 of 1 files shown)


resources/js/components/Elements/Slider.vue
```diff
@@ -202,14 +202,38 @@ export default {
...
         },
         slidesVisible() {
             if (!this.mounted) {
                 return 0
             }
 
-            return Math.round(this.sliderSpan / this.childSpan)
+            return Math.round(
+                (this.sliderSpan / (this.vertical ? this.container.scrollHeight : this.container.scrollWidth)) * this.container.children.length,
+            )
```
indykoning, line 303: The reason we're not calculating from the currentSlide to the last visible slide is because this could cause the visibleSlides to jump up and down depending on your position in the slider.

While visibleSlides is also used in slidesTotal, causing the total to change

resources/js/components/Elements/Slider.vue
```diff
@@ -202,14 +202,39 @@ export default {
             if (!this.mounted) {
                 return 0
             }
-            return Math.round(this.position / this.childSpan) % this.slides.length
+            // First make a calculated guess, improving performance by not having to loop through all slides
+            const bestGuess = Math.round(this.position / this.childSpan) % this.slides.length
```
indykoning: This bestGuess is correct on regularly sized sliders.
Image
Depending on the size differences in other childs it may only have to check one additional element.
Image

resources/js/components/Elements/Slider.vue
```diff
@@ -202,14 +202,39 @@ export default {
...
+                const bestGuessOffset = this.vertical ? bestGuessChild.offsetTop : bestGuessChild.offsetLeft
+
+                // Past halfway to the slide it will snap
+                if (this.position + bestGuessSpan / 2 < bestGuessOffset) {
+                    return getSlideByGuess(slide - 1)
+                }
+
+                if (this.position >= bestGuessOffset + bestGuessSpan) {
+                    return getSlideByGuess(slide + 1)
+                }
```
Jade-GG, line 288: Are you sure this is right? The way I'm reading this right now is that it will prefer the best guess when the actual position is to the left of the best guess, with up to half of the span of the *current slide* of extra threshold. This to me suggests that if the previous slide is less than half the size of the best guess slide, it will never be properly matched.

It also appears unbalanced, with this extra threshold not applying to the right side.

Personally, I would suggest a different approach where you "index" the lefts and rights of all slides once (and whenever a `MutationObserver` notices a change), and doing calculations based on that. Also consider the fact that there may be gaps between slider items due to frontend shenanigans.

resources/js/components/Elements/Slider.vue
Jade-GG: I think this will still cause incorrect calculations for the `slidesTotal` now. A correct calculation here would take into account first how many items there are, and then only subtract how many items are visible when the slider is all the way at the end.

Is this value still relevant? Because on most shops we already hide the "progress bar" under sliders anyway.

Jade-GG (changes requested):

resources/js/components/Elements/Slider.vue
indykoning: This is right, as you can see in https://github.com/rapidez/core/pull/1334#discussion_r3615109024 it exactly matches original behavior when spans are equally sized.
When expanded to irregularly sized slides it continues to maintain proper behavior.

Mutation/intersection observers was my first choice, the reason i have not done so is because the slides are not guaranteed to stay. (these sliders are used within some filtering where depending on your filters these slides shrink or grow)
This method is much less expensive than continually attaching and removing observers, measuring this entire computed property taking 70μs without causing style recalculations.

Margins and gaps are no longer of concern since we use the position of the element within the scrollbar, this would automatically take them into account.

resources/js/components/Elements/Slider.vue
indykoning: Where do you see this calculation being incorrect?
Say we have a slider of 7 items where 4 are visible.
The slider is 1252px wide and elements are 313px wide. This would turn into:
1252 / 2191 * 7 = 4
We're taking the visible size, dividing it by the total scroll width (each element including margins, paddings and gaps), and multiplying it by the amount of items in the slide.
Another way you could write it is:

```js
let averageSlideWidth = 2191 / 7; // 313
let visibleSlides = 1252 / 313; // 4
```

The only change in this calculation is that instead of using the width of the first child, we're using the average width of all childs.
I see slidesVisible being used and shown: In slidesTotal, in the "progress bars", the numbers and next and previous buttons where we want the slider to slide pages instead of single slides

Jade-GG: I'm imagining a hypothetical situation where you have 100 wide slides (say, 100px wide each) followed by a bunch extremely tiny slides (say, 1px) at the end, where only two of the wide slides are visible (so the slider span itself is 200px).

slidesVisible would still be 2 even if you have up to 25 of the tiny slides at the end. With the current calculation of the slidesTotal, you'd have 24 extra useless "slides" in your progress bar (as all of the tiny slides would all be visible after slide number 100). Obviously this is an absurd example, but more realistically I could see there being 1 too few or 1 too many slides in real world situations.

resources/js/components/Elements/Slider.vue
```diff
@@ -64,11 +64,11 @@ export default {
         }
         this.$nextTick(() => {
             useResizeObserver(this.slider, useThrottleFn(this.updateSpan, 150, true, true))
+            setTimeout(() => this.slider.dispatchEvent(new CustomEvent('scroll')))
+            this.mounted = true
```
indykoning, line 72: This new order is necessary due to https://github.com/rapidez/core/commit/176b2cd85dc9b7b3e39cf9a814ab38fcab6cb710#diff-2d53f035dc6e0201915f81219d324c2479c3acd61fde0b67ffbc091205965494R222

resources/js/components/Elements/Slider.vue
```diff
@@ -151,11 +181,25 @@ export default {
...
-            index = this.loop ? index + this.slides.length : index
+            this.handleLoopBoundary()
+            index = Math.min(Math.max(index, 0), this.container?.children.length - 1)
+            if (this.loop) {
+                if (index < this.slides.length - 1) {
+                    index += this.slides.length
+                } else if (index > this.slides.length * 2 + 1) {
+                    index -= this.slides.length
+                }
+            }
```
indykoning, line 192: By calling handleLoopBoundary and recalculating the index depending on wether loop is enabled. we can ensure scrolling remains smooth while resetting the scroll position. 

e.g. user is on the final slide but presses the right button again while scrolling.
before it may use smooth scrolling to scroll back instead of instant, or it might not scroll back at all.
now we first handle the loop boundary, resetting instant to the first slide, then recalculate what the index should be and smoothly scroll to the next slide

resources/js/components/Elements/Slider.vue
```diff
@@ -193,23 +237,72 @@ export default {
...
+            set (position) {
+                return this.vertical ? this.scrollY = position : this.scrollX = position
+            },
+        },
+        showLeft() {
+            return !(this.vertical ? this.arrivedState.top : this.arrivedState.left)
+        },
+        showRight() {
+            return !(this.vertical ? this.arrivedState.bottom : this.arrivedState.right)
+        },
```
indykoning, line 263: By using useScroll we can heavily simplify show left and show right. Since it simply checks wether we've reached the end of the slider we no longer need to check wether we're looping

resources/js/components/Elements/Slider.vue
```diff
@@ -83,6 +87,32 @@ export default {
...
+                throttle: 150,
+                idle: 100,
+                eventListenerOptions: { passive: true },
+                behavior: this.scrollBehavior,
+            })
+
+            this.scrollX = x
+            this.scrollY = y
+            this.isScrolling = isScrolling
+            this.arrivedState = arrivedState
```
indykoning, line 101: All these values are now reactive, meaning we can use listeners and simple setters to set the position.

royduin:
- Any ref? 
- Do you approve @Jade-GG?
- Ready to merge?

Jade-GG:
I mostly approve aside from my `slidesTotal` edge case that I'm worried about.

indykoning:
> I mostly approve aside from my slidesTotal edge case that I'm worried about.

We have since discussed this worry, the slidesTotal use is so niche that we considered removing the property altogether.
It is still used for a simple number in some projects, we're keeping it for that. The edge case is niche enough that a TODO suffices in case we ever run into issues due to it.

PR #1335, opened 2026-07-21
Added collapsible filter configuration

indykoning:
Depends on (but does not require) https://github.com/rapidez/magento2-compadre/pull/16

If compadre is not installed, default functionality is restored meaning all filters are open by default
Ref: FW-2421

### DIFF
--- a/resources/views/components/accordion/filter.blade.php
+++ b/resources/views/components/accordion/filter.blade.php
@@ -1,7 +1,7 @@
 @slots(['content', 'label', 'icon'])
 @props(['canToggleShowMore' => false])
 
-<x-rapidez::accordion open class="border-t" :$attributes :$icon>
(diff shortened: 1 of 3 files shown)


resources/views/components/accordion/filter.blade.php
```diff
@@ -1,7 +1,7 @@
 @slots(['content', 'label', 'icon'])
 @props(['canToggleShowMore' => false])
 
-<x-rapidez::accordion open class="border-t" :$attributes :$icon>
+<x-rapidez::accordion v-bind:open="!window.config.collapsed_attributes?.includes(filter?.code?.replace('super_', '')) || items.some(item => item.isRefined)" class="border-t" :$attributes :$icon>
```
indykoning, line 4: If config says the attribute filter should be collapsed, but filters are selected on this attribute we keep it open.

PR #1346, opened 2026-08-05
[5.x] Use subtotal_incl/excl_tax

Jade-GG:
ref: HDB-1129

`subtotal` is deprecated in favor of `subtotal_incl_tax` and `subtotal_excl_tax`, see https://developer.adobe.com/commerce/webapi/reference/graphql/latest/types-k-p#ordertotal

### DIFF
--- a/resources/views/checkout/queries/fragments/orderV2.graphql
+++ b/resources/views/checkout/queries/fragments/orderV2.graphql
@@ -51,7 +51,10 @@ fragment orderV2 on CustomerOrder {
         ...orderItem
     }
     total {
-        subtotal {
+        subtotal_incl_tax {
+            value
+        }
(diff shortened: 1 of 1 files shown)


royduin:
This is not available in Magento 2.4.7, do we really need this? In that case we've to bump the Magento version requirement to 2.4.8+

Jade-GG:
> This is not available in Magento 2.4.7, do we really need this? In that case we've to bump the Magento version requirement to 2.4.8+

You can work around it by manually calculating the tax values, which we did in the past in most places (except for here, for some reason). We could make a backwards compatible version, but is that worth it at all? 2.4.8 has been out for a while now.

royduin:
We dropped it 🚀

PR #1353, opened 2026-08-07
Vite 8 support

indykoning:
In a new project i updated all dependencies, only running into an error with the combininggraphql See: https://rolldown.rs/in-depth/bundling-cjs#recommendations-for-library-authors
Ref: DONG-37


indykoning:
It seems that the updated Vite stack handles bundling differently
image 
vs 
image

As you can see, the app.js significantly reduces in size due to Vue being bundled separately

indykoning:
We might want to consider switching to nginx in the pipelines.
On my local machine this is the difference between nginx and `php artisan serve` is 100 performance score and 90 performance score

PR #1356, opened 2026-08-14
Info for AI

royduin:
The `CLAUDE.md` here is just in case you're working directly on the core. The guideline is for [Laravel Boost](https://laravel.com/docs/boost), currently there are no skills. I did create some but it was kind of copy of the docs which didn't felt very maintainable. We could sync the docs automatically to skills but lets see how this is going with a link to https://docs.rapidez.io/llms.txt which was added in: https://github.com/rapidez/docs/pull/127. Also added a new AI page in the docs: https://github.com/rapidez/docs/pull/128

### DIFF
--- a/resources/boost/guidelines/core.blade.php
+++ b/resources/boost/guidelines/core.blade.php
@@ -12,3 +12,4 @@ IMPORTANT!! For all Rapidez specific questions please check the https://docs.rap
 - Cart, checkout, customer mutations, etc everything that has authentication, is dynamic or need to write to Magento → Magento GraphQL.
 - Search and category filtering → IntantSearch with Elasticsearch or OpenSearch, kept in sync via the indexing commands this package provides.
 - Views are plain Blade — no `.phtml`, no Magento XML layout files.
+- Vue components are renderless; separating the logic of a component from its presentation
--- a/resources/boost/guidelines/core.blade.php
+++ b/resources/boost/guidelines/core.blade.php
@@ -1,0 +1,14 @@
+## Rapidez Core
+
+Rapidez is a headless frontend for Magento 2 built with Laravel, Vue and InstantSearch (with Elasticsearch/Opensearch). `rapidez/core` is the foundation package — it provides the Eloquent models that read the Magento database directly (for catalog, category, CMS data, etc), integrates the Magento GraphQL APIs (for cart, checkout, customer actions, etc), and supplies the base Blade views, routing, and indexing commands that a Rapidez storefront is built on.
+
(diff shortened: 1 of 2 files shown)


resources/boost/guidelines/core.blade.php
indykoning, line 1: Perhaps we should add a line that we use [renderless vue components](https://www.patterns.dev/vue/renderless-components/)

royduin: Added ✅

resources/boost/guidelines/core.blade.php
```diff
@@ -0,0 +1,15 @@
+## Rapidez Core
+
+Rapidez is a headless frontend for Magento 2 built with Laravel, Vue and InstantSearch (with Elasticsearch/Opensearch). `rapidez/core` is the foundation package — it provides the Eloquent models that read the Magento database directly (for catalog, category, CMS data, etc), integrates the Magento GraphQL APIs (for cart, checkout, customer actions, etc), and supplies the base Blade views, routing, and indexing commands that a Rapidez storefront is built on.
+
+Never treat this as a "normal" Laravel e-commerce package: catalog data is read straight from Magento's database schema via Eloquent, not through Magento's own ORM or REST API — GraphQL is only used for write/transactional actions like cart and checkout.
+
+IMPORTANT!! For all Rapidez specific questions please check the https://docs.rapidez.io/llms.txt and the relevant links first!
```
Jade-GG, line 7: Is this really the best way to make sure any LLM understands this? Is there a standardized way to give "important" context? Isn't everything in this file important enough anyway?

Given the sliding context windows issue, if it's really important it should also probably be all the way at the bottom of the file as a precaution. Especially if we expand this list below eventually.

resources/boost/guidelines/core.blade.php
```diff
@@ -0,0 +1,15 @@
...
+
+### Architecture
+
+- Catalog, category, CMS data, etc everything that's we need to read from the Magento → Eloquent models querying the Magento database directly (see `Rapidez\Core\Models\*`)
+- Cart, checkout, customer mutations, etc everything that has authentication, is dynamic or need to write to Magento → Magento GraphQL.
```
Jade-GG, line 12: These 2 instructions don't seem particularly clear to me and I could easily see an LLM tripping up and mixing them up. I'm not an expert on this, but I would personally expect a small table to be easier to interpret.

resources/boost/guidelines/core.blade.php
royduin: - Not sure, I've seen other packages do it and it seem to work fine.
- For the second part; I don't understand what you're saying here.

resources/boost/guidelines/core.blade.php
royduin: So far so good, suggestions are welcome

PR #1360, opened 2026-08-17
Run cleanup functions on before-cache

indykoning:
Since Vue 3 functionality regarding unmounting has changed with useEventListener.
This results in the cleanup function only being called for `useEventListener` calls done within a .vue component.

This caused e.g. addToCart listeners of the gtm package to start stacking the same event listener.
Luckily `useEventListener` also returns the cleanup function, we unmount Vue on `turbo:before-cache` so we can cleanup the listeners ourselves.

Ref: FW-2481

### DIFF
--- a/resources/js/polyfills/emit.js
+++ b/resources/js/polyfills/emit.js
@@ -1,5 +1,7 @@
 import { useEventListener } from '@vueuse/core'
 
+let autoRemoveCleanupFunctions = []
+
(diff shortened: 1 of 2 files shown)


Jade-GG (approved):

Jade-GG:
How much value is there still to using `useEventListener` here if we have to clean the event listeners up ourselves anyway?

PR #1361, opened 2026-08-18
Fixed productlist causing errors when value is false

indykoning:
https://github.com/rapidez/core/pull/1312 changed the checks. Adding an additional template check.

However due to the check only being for !is_iterable it activates on "false" as well.
This will render `<template v-if=".length"` causing a syntax error in Vue

We should instead check if it is a string


royduin (approved):
Looking good to me, let's wait for @Jade-GG his review.

Jade-GG (approved):

PR #1367, opened 2026-08-31
[5.x] Use useLocalStorage for popup

Jade-GG:
ref: HDB-1136

The old version caused some errors, presumably due to racing conditions. `useLocalStorage` provides more of a safety net.


royduin:
- Playwright tests are failing?
- The default / fallback is now defined 3 times, can we centralize that?

PR #1376, opened 2026-09-04
[5.x] Remove default loading state from input

Jade-GG:
This is an issue because the disabled state also disables validation on this input. As a result, when a request is running (e.g. postcode validation), the whole checkout form gets disabled and suddenly everything is considered valid. This then causes a ton of errors from magento.

see: RAP-1967


royduin:
But how do we prevent editing fields when a GraphQL request is running?

royduin:
Follow-up: RAP-1967

PR #1381, opened 2026-09-08
Use super labels instead of their int values

indykoning:
As a quick example. this is what gets indexed:
image
meaning the following shows up as filter:
image
After this change it will show up as:
image

Both the super_xxx and super_xxx_values are used in the AddToCart.vue
(though we could remove the need for `super_xxx` by using `Object.keys("super_xxx_values")` but that will be a breaking change)


indykoning:
This turned out to be a fault of the project, overwriting text swatches to work as a regular select without calling `withSwatches` to get the labels of these text swatches

PR #1382, opened 2026-09-09
[4.x] Get latest timestamp from indexer

Jade-GG:
ref: RAP-1998

This is not an issue in 5.x because we don't use the flat table there anymore.

We need to know when the flat table has been actually updated last, because otherwise changes my fall through the cracks. This is more relevant for the code that clears the cache in rapidez/statamic.

### DIFF
--- a/src/Actions/GetLatestIndexTimestamp.php
+++ b/src/Actions/GetLatestIndexTimestamp.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace Rapidez\Core\Actions;
+
+use Illuminate\Support\Carbon;
+use Illuminate\Support\Facades\DB;
+
+class GetLatestIndexTimestamp
+{
+    public function get(): Carbon
+    {
+        $reindexAllInvalidTime = new Carbon(DB::scalar('SELECT scheduled_at FROM cron_schedule WHERE `job_code` = \'indexer_reindex_all_invalid\' AND `status` = \'success\' ORDER BY `scheduled_at` DESC LIMIT 1') ?? 0);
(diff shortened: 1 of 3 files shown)


src/Actions/GetLatestIndexTimestamp.php
```diff
@@ -0,0 +1,22 @@
...
+
+use Illuminate\Support\Carbon;
+use Illuminate\Support\Facades\DB;
+
+class GetLatestIndexTimestamp
+{
+    public function get(): Carbon
+    {
+        $reindexAllInvalidTime = new Carbon(DB::scalar('SELECT scheduled_at FROM cron_schedule WHERE `job_code` = \'indexer_reindex_all_invalid\' AND `status` = \'success\' ORDER BY `scheduled_at` DESC LIMIT 1') ?? 0);
+        $latestFullIndexTime = new Carbon(DB::scalar('SELECT updated FROM indexer_state WHERE `indexer_id` = \'catalog_product_flat\' OR `indexer_id` = \'catalog_category_flat\' ORDER BY updated LIMIT 1') ?? 0);
```
royduin:
```suggestion
        $reindexAllInvalidTime = new Carbon(DB::scalar("SELECT scheduled_at FROM cron_schedule WHERE `job_code` = 'indexer_reindex_all_invalid' AND `status` = 'success' ORDER BY `scheduled_at` DESC LIMIT 1") ?? 0);
        $latestFullIndexTime = new Carbon(DB::scalar("SELECT updated FROM indexer_state WHERE `indexer_id` = 'catalog_product_flat' OR `indexer_id` = 'catalog_category_flat' ORDER BY updated LIMIT 1") ?? 0);
```

PR #1385, opened 2026-09-11
JS performance improvements

royduin:
Claude tried to improve the performance and came with these results. We've to double check and tests these changes, the measurements with 4× CPU throttle:

  | before | after |  
-- | -- | -- | --
warm cache | 453ms | 429ms | −24ms (~5%)
cold cache | 782ms | 695ms | −87ms (~11%)

Ref: FW-2599

### DIFF
--- a/resources/js/components/Listing/Listing.vue
+++ b/resources/js/components/Listing/Listing.vue
@@ -2,24 +2,10 @@
 import { user } from '../../stores/useUser'
 import { history } from 'instantsearch.js/es/lib/routers'
 import InstantSearchMixin from '../Search/InstantSearchMixin.vue'
-import Pagination from 'vue-instantsearch/vue3/es/src/components/Pagination.vue.js'
-import SearchBox from 'vue-instantsearch/vue3/es/src/components/SearchBox.vue.js'
-import RangeInput from 'vue-instantsearch/vue3/es/src/components/RangeInput.vue.js'
-import HierarchicalMenu from 'vue-instantsearch/vue3/es/src/components/HierarchicalMenu.vue.js'
-import RefinementList from 'vue-instantsearch/vue3/es/src/components/RefinementList.vue.js'
-import SortBy from 'vue-instantsearch/vue3/es/src/components/SortBy.vue.js'
 import { instantsearchMiddlewares } from '../../stores/useInstantsearchMiddlewares'
 
 export default {
     mixins: [InstantSearchMixin],
-    components: {
-        Pagination,
-        SearchBox,
-        RangeInput,
-        HierarchicalMenu,
-        RefinementList,
-        SortBy,
-    },
     props: {
         index: {
             type: String,
--- a/resources/js/instantsearch.js
+++ b/resources/js/instantsearch.js
@@ -1,8 +1,13 @@
 import { defineAsyncComponent } from 'vue'
 import { addQuery } from './stores/useSearchHistory'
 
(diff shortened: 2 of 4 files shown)


resources/js/components/Listing/Listing.vue
royduin, line 1: These components are globally registered and the `ais-` prefixed versions are used so it looks like we don't need them here.

resources/js/instantsearch.js
royduin, line 1: With this the components are loaded directly on import with a fallback on defineAsyncComponent

resources/js/package.js
```diff
@@ -56,6 +56,24 @@ if (import.meta.env.VITE_DEBUG === 'true') {
     )
 }
 
+// Booting is deferred to the next frame so the server rendered markup can paint
+// before Vue starts. requestAnimationFrame never fires while the tab is hidden,
+// which would leave the page unbooted until it's focused, so a timer runs against
+// it and whichever comes first wins.
+function nextFrame(callback) {
```
royduin, line 63: With this everything still renders in the background (not an active tab)

resources/js/package.js
```diff
@@ -149,11 +167,17 @@ async function init() {
                 },
 
                 categoryPositions(categoryId) {
```
royduin, line 169: With this die script could be cached and doesn't need to compile for each category.

resources/views/components/listing.blade.php
```diff
@@ -2,7 +2,7 @@
 
 @pushOnce('head', 'es_url-preconnect')
     <link rel="preconnect" href="{{ config('rapidez.es_url') }}">
-    @vite(vite_filename_paths(['Listing.vue', 'InstantSearch', 'instantsearch-components']))
+    @vite(vite_filename_paths(['Listing.vue', 'InstantSearch', 'instantsearch-components', 'RangeSlider.vue']))
```
royduin, line 5: On a listing with filters we need this component, so let's get it faster.

resources/js/instantsearch.js
```diff
@@ -1,8 +1,13 @@
...
 
-const instantsearchComponents = import('./instantsearch-components')
-const component = (name) => defineAsyncComponent(() => instantsearchComponents.then((m) => m[name]))
+let loaded = null
+export const instantsearchComponents = import('./instantsearch-components').then((m) => (loaded = m))
+
+// Once the chunk is in, registering the component itself instead of an async
+// wrapper saves a render pass per component; without it every one of them resolves
+// separately before InstantSearch can register its widgets and search.
+const component = (name) => (loaded ? loaded[name] : defineAsyncComponent(() => instantsearchComponents.then((m) => m[name])))
```
Jade-GG, line 10: Is this something we can manage in `instantsearch-components.js` instead? This feels like extreme magic when done this way.

resources/js/package.js
```diff
@@ -56,6 +56,24 @@ if (import.meta.env.VITE_DEBUG === 'true') {
...
+// it and whichever comes first wins.
+function nextFrame(callback) {
+    let called = false
+    let run = () => {
+        if (called) {
+            return
+        }
+        called = true
+        callback()
+    }
```
Jade-GG, line 71: Can we use a proper mutex for this (possibly with `Atomics.wait()` + `Atomics.notify()`)? I feel like this construction, while seemingly robust, still can invite racing conditions.

PR #1387, opened 2026-09-11
Reduce the data indexed for children in ElasticSearch

royduin:
A follow-up on https://github.com/rapidez/core/pull/1386, if we're not returning the data, we don't need it in the index. This is code from Claude en currently untested!

Ref: FW-2596

### DIFF
--- a/src/Models/Traits/Product/Searchable.php
+++ b/src/Models/Traits/Product/Searchable.php
@@ -6,6 +6,7 @@ use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Collection;
 use Illuminate\Support\Facades\Cache;
+use Illuminate\Support\Str;
 use Rapidez\Core\Enums\Visibility;
 use Rapidez\Core\Facades\Rapidez;
 use Rapidez\Core\Models\Category;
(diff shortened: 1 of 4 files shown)


src/Models/Traits/Product/Searchable.php
```diff
@@ -52,6 +53,28 @@ public function toSearchableArray(): array
...
+            ->filter(fn (string $code) => str_starts_with($code, 'children.*.'))
+            ->map(fn (string $code) => Str::after($code, 'children.*.'))
+            ->values()
+            ->toArray();
+
+        $childAttributeCodes = collect($childAttributePaths)
+            ->map(fn (string $path) => Str::before($path, '.'))
+            ->unique()
+            ->values()
+            ->toArray();
```
Jade-GG, line 71: These repeated `->values()->toArray()`s seem unnecessary :)

src/Models/Traits/Product/Searchable.php
```diff
@@ -93,6 +124,37 @@ public function toSearchableArray(): array
         return Eventy::filter('index.' . static::getModelName() . '.data', $data, $this);
     }
 
+    /**
+     * Recursively keep only the given dot-notated paths (e.g. `stock.is_in_stock`) of an array.
+     */
+    protected static function onlyPaths(array $data, array $paths): array
```
Jade-GG, line 130: Can we not just use `data_get()` in a foreach here?

Jade-GG (changes requested):
I'm not so sure about this solution. Now, if you want to add or remove just one attribute from the children data, you will then have to re-define all of the other ones including any super attribute codes which aren't even static.

I wouldn't expect this behavior from just adding a single `children.my_custom_value` to the list.

PR #1389, opened 2026-09-15
[5.x] Fix racing condition with login and checkout data

Jade-GG:
ref: BORDEX-1483

We saw a Sentry replay that was interesting. What happened was:

- Customer has a cart and enters the checkout
- Browser autofill is used to simultaneously fill the address data and login fields
- Customer logs in immediately while the graphql-mutation component is still within its debounce time
- Login completes, cart id changes to the logged in user, which means the pending graphql-mutation now has an invalid cart id attached
- This request then gives a "cart expired" error, but Rapidez considers the *current* cart expired

With this PR, I've implemented two different fixes:

- First, I fixed the `retryOnCartError` code which already almost fixed this issue by itself; it just checked the wrong cart id.
- Then, if this still somehow occurs (unlikely but theoretically still possible) then fix this issue by comparing the request's cart id with the current cart id. If it's different, we let graphql error out without consequence.


royduin:
Let's wait until @indykoning is back from his vacation

PR #1390, opened 2026-09-15
Indexer exit code based on sub command exit codes

royduin:
A nice little improvement so the index command returns the right exit code

Ref: RAP-2017


Jade-GG:
Does this have any use right now? I don't see the command actually returning anything right now by itself, and if it actually throws an error the whole command fails instantly anyway.

royduin:
The subcommand throws exceptions so this isn't needed.

PR #1392, opened 2026-09-15
Reduce the queries to track a product view

royduin:
We don't need all the eager loads here:

https://github.com/rapidez/core/blob/4393478c056031b794c92b4f6a14aef69f0a84d7/src/Models/Product.php#L37-L46

Ref: RAP-2017


src/Http/Controllers/ProductController.php
```diff
@@ -94,7 +94,7 @@ public function track(int $productId)
         $productModel = config('rapidez.models.product');
 
         /** @var \Rapidez\Core\Models\Product $product */
-        $product = $productModel::findOrFail($productId);
+        $product = $productModel::query()->withOnly([])->findOrFail($productId);
```
Jade-GG:
```suggestion
        $product = $productModel::withOnly([])->findOrFail($productId);
```

As far as I can tell this should work. Right?

PR #1400, opened 2026-09-15
Cleanup the image component listener

royduin:
Without this the listener seems to be left behind

Ref: RAP-2017


Jade-GG:
Strange, is this at all related to #1360? I'm not sure if this is the way we want to go here because if this one fails, other places may also go awry.

royduin:
Better safe than sorry

PR #1401, opened 2026-09-15
Various js improvements

royduin:
A bundle of various js improvements, best to review without whitespace! Mostly focussing on cancelling simultaneous requests

Ref: RAP-2017


resources/js/components/Checkout/CheckoutLogin.vue
royduin, line 1: Here we're just keeping track of the request with `submitting`, if it's still running don't start another one

resources/js/components/Listing/Listing.vue
```diff
@@ -88,7 +88,7 @@ export default {
 
         sortOptions() {
             let groupId = user?.value?.group_id || 0
-            let sorting = config.searchkit.sorting
+            let sorting = structuredClone(config.searchkit.sorting)
```
royduin, line 91: Previously we where changing the config data itself, now we're cloning it.

resources/js/components/Product/AddToCart.vue
royduin, line 1: Also a check on `adding` to stop when there is already a request running and an additional check in case the response isn't json.

resources/js/stores/useCart.js
royduin, line 1: This is a cool one, when 2 requests are running and the oldest one finishes after the latest one the response will not be used to update the cart

resources/js/callbacks.js
royduin, line 1: For the last one some Claude info, for the first change:
```
// Keep the global `loading` flag (and therefore every "Next"/submit button bound to
// it) true for the whole span of this call, not just while an individual request is
// in flight. In sequential mode each partial-submit step is awaited one at a time,
// and combiningGraphQL defers its actual fetch by a few ms to allow batching -- so
// there's a brief gap between steps where fetchCount would otherwise drop to 0 and
// the button re-enables mid-submit, both flickering the loading indicator and
// opening a window for a second click to slip through.
```
I think you're the best in reviewing this part @indykoning, it feels like a good addition but we do need to make sure this isn't causing any other issues + we should not merge this before making sure the Playwright tests pass.

And the second change:
```
// Treat a missing order as a failure even when `errors` isn't populated (e.g. a
// payment-gateway decline or a differently-shaped partial response): previously
// this fell through silently, set order.value to null, and let GraphqlMutation
// redirect straight to the success page -- only CheckoutSuccess.vue's own guard
// caught it, bouncing the customer back to /cart with no explanation why.
```

resources/js/components/Listing/Listing.vue
Jade-GG: First time we're using this function anywhere. Available since 2022 so I'm sure it's fine by now, but given that the JSON.parse trick is apparently generally faster on chrome, do we want to use it everywhere?

resources/js/stores/useCart.js
```diff
@@ -200,6 +201,8 @@ export const cart = computed({
         return cartStorage.value
     },
     set(value) {
+        const seq = ++cartWriteSeq
```
Jade-GG, line 204: These kinds of inline tricks are cool but I'd rather just see it written out fully.

```suggestion
        const seq = cartWriteSeq
        cartWriteSeq++
```

Maybe also rename the variable to something more sensible, like `cartWriteIndex`. Doesn't make much sense to shorten it here when it's already a long-ish variable name.

Jade-GG (approved):

royduin:
Let's wait until @indykoning is back from his vacation

